PR #17, opened 2014-06-21
Fixing LoadLibrary in the case the application is using UNICODE.

nicolasnoble:
The current code is assuming the library is compiled and used in ANSI, by always passing an ANSI string to LoadLibrary. It means that if the library is in fact compiled in UNICODE, the LoadLibrary call will assume you're passing a widechar string, and will ultimately fail.

This commit fixes it by forcing the usage of the UNICODE version of the API, and ignoring the compilation settings of the library.

This is mostly relevant if the library is compiled statically within another application.

### DIFF
--- a/ares_library_init.c
+++ b/ares_library_init.c
@@ -45,7 +45,7 @@ static int ares_win32_init(void)
 #ifdef USE_WINSOCK
 
   hnd_iphlpapi = 0;
-  hnd_iphlpapi = LoadLibrary("iphlpapi.dll");
+  hnd_iphlpapi = LoadLibraryW(L"iphlpapi.dll");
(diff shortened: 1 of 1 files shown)


bagder:
I would like this acked by someone else who knows Windows before I merge it. I do not.

saghul:
FWIW, this patch has been floated by Node for quite a while, so I guess one could say it's battle tested. https://github.com/nodejs/node/commit/cfafba68c64fed0626af8b22d6c4cdccb3261ced

Also, the person who reviewed the patch (Bert Belder) is a Windows expert.

bagder:
Good enough for me, thanks!

bagder:
thanks @nicolasnoble

PR #22, opened 2014-09-05
Add support for reading trusted dnssec servers from /etc/resolv.conf

nmav:
This patch uses /etc/resolv.conf and the 'trusted-nameserver' extension to determine the servers to use when dnssec is requested.

### DIFF
--- a/ares_init.c
+++ b/ares_init.c
@@ -1067,67 +1067,82 @@ static int init_by_resolv_conf(ares_channel channel)
   if (channel->nservers > -1)  /* don't override ARES_OPT_SERVER */
      return ARES_SUCCESS;
 
-  if (get_DNS_Windows(&line))
-  {
-    status = config_nameserver(&servers, &nservers, line);
-    free(line);
-  }
+  /* FIXME: don't know how to read trusted DNSSEC servers on this platform */
+  if ((channel->flags & ARES_FLAG_DNSSEC) != ARES_FLAG_DNSSEC) {
+    if (get_DNS_Windows(&line))
+    {
+      status = config_nameserver(&servers, &nservers, line);
+      free(line);
+    }
 
-  if (status == ARES_SUCCESS)
-    status = ARES_EOF;
-  else
-    /* Catch the case when all the above checks fail (which happens when there
-       is no network card or the cable is unplugged) */
-    status = ARES_EFILE;
+    if (status == ARES_SUCCESS)
+      status = ARES_EOF;
+    else
+      /* Catch the case when all the above checks fail (which happens when there
+         is no network card or the cable is unplugged) */
+      status = ARES_EFILE;
+  } else {
+      status = ARES_EOF;
+  }
 
 #elif defined(__riscos__)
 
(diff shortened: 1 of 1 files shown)


nmav:
I'm delaying this patch until the glibc interface for resolv.conf is fixed.

bradh352:
As far as I can tell, the trusted-nameserver extensions never made it into glibc and were never standardized anywhere and abandoned, so closing this.  Please re-open with new information if this is wrong.

jscott0:
> As far as I can tell, the trusted-nameserver extensions never made it into glibc and were never standardized anywhere and abandoned, so closing this. Please re-open with new information if this is wrong.

@bradh352 FYI and for the record, this was addressed a few years ago. I think the matter especially picked up steam with systemd-resolved gaining popularity, because the systemd-resolved stub resolver that runs on the loopback interface commonly performs DNSSEC validation (some distros change this default), but—on the other hand—sometimes resolv.conf is automatically populated with servers on a LAN that aren't trustworthy. This design problem has been solved by the glibc folks a little while ago, and I believe musl and at least one of the BSD's has picked it up.

Newer versions of glibc will, in the absence of any indication otherwise, not trust DNSSEC validation performed by nameservers listed in resolv.conf, and will effectively discard the AD bit, including when exposing validation results to applications with interfaces like res_query() and such. So this is a safe default for a world where people use DNS resolvers that aren't so trustworthy.

Nowadays it's increasingly common to run a validating resolver on the loopback interface such as systemd-resolved or Unbound. On installation usually 'nameserver 127.0.0.1' or the like will get inserted into /etc/resolv.conf. In this case you can obviously trust those results—the validation is being performed on your machine—so these daemons will arrange for 'options edns0 trust-ad' to be inserted into resolv.conf as well, when the stub resolver is "taking ownership" of resolv.conf and is the sole nameserver designated there. Now folks parsing resolv.conf manually or using the Name Service Switch APIs to learn what the nameservers of choice are, have an affirmative signal that "yes, if this nameserver sets the AD bit in replies, the sysadmin considers that trustworthy, so you may make decisions based on it". (In practice this almost always is a resolver on the loopback address, but it doesn't technically have to be that way.)

This is also useful so applications that wish to use the normal system-wide nameservers like usual can be informed "Yes, these TLSA resource records are authentic, and you can trust that I'm not a liar."

So anyway, this has come a long way, DNSSEC and DANE are much more widely used than they once were (especially for SMTP and XMPP), allied projects have been pioneers in this space regardless (Debian supports DANE, including SSHFP and OPENPGPKEY resource records), and the solution is settling in. Some libraries like libunbound and (via libunbound) GnuTLS effectively run their own in-process internal recursive resolvers because they didn't know if they could "trust" the normally-configured nameservers, but this obviously doesn't scale well... Now it's practical to use the system-wide resolver, especially for oppurtinistic features. Network Manager, resolvconf-type tools, and higher-level administrative tools also recognize or insert the trust-ad option as appropriate.

Although this isn't strictly necessary, this would make "DANE by default" with curl + c-ares + OpenSSL especially sound.

PR #25, opened 2015-01-08
Replace LoadLibrary by LoadLibraryW

zcbenz:
On Windows, when compiling with `UNICODE` defined, `LoadLibrary` becomes `LoadLibraryW`. When an ASCII string is passed to that function it crashes.

See https://github.com/iojs/io.js/pull/226 for more discussions on this.


nicolasnoble:
Fun, I made the exact same pull request a while ago: #17

zcbenz:
@nicolasnoble lol.

PR #27, opened 2015-10-21
Master

qwerty258:
1. Add VS 2013 project files.
2. Rename README to README.md so it would look good on Github web.


dimbleby:
This is surely unnecessary?  Just use 

    buildconf.bat
    nmake /f Makefile.msvc

IMO the `vc` directory should be removed altogether.  It's only for MSVC 6.0 which surely no-one is using these days; and anyway the `nmake` mechanism works there too.

(Or at least, `nmake` works no worse than the project files do.  I tried and, unsurprisingly, support for VC6 seems to have bit-rotted - confirming my suggestion that no-one is actually using it).

qwerty258:
IMO this won't hurt anyone.
This is good for beginner who doesn't know nmake or those favour GUI.
It's OK to close this request.

bagder:
It will "hurt" the second the files go stale and don't work anymore as then people will come asking or whining and needing help. I could see us merging files to allow building with project files but then I think we should *generate* the output files necessary so that when we update the project and change the source file set, it will still keep working.

qwerty258:
It will take sometime for me to figure out how to generate VS project files automatically like cURL.

PR #28, opened 2015-11-09
added another version case to Makefile.msvc

nyx:
nmake version 11.00.61030.0 resolves to CC_VERS_NUM = 110


nyx:
this was merged via the mailing list. closing.

PR #47, opened 2016-02-24
remove compilation warning in MSVC

kolomenkin:


bradh352:
There are 3 instances of GetVersionEx in c-ares, this appears to only address a single instance.

Also, Microsoft wants you to use VerifyVersionInfo() these days as they say GetVersionEx() may not be available or return the right thing in the future, so that would be the more proper fix, but would involve killing ancient Windows versions (which I'd be fine with, but can't guarantee others may not have a different opinion).

So we'd need to figure out how we want to move forward here.

kolomenkin:
The point about 3 instances of GetVersionEx is good.

I know about deprecation of GetVersionEx. We can think about rewriting the code, but you should keep in mind:

- [ ] While MS tells they can drop GetVersionEx, they will not drop it ever for compatibility. It may return not correct Win version like it does on latest Windows, but it will be available.
- [ ] In this particular case we are working only with dwPlatformId member. This member is correctly filled in all known Windows versions. The problem sometimes occurs when you try to detect Windows version (like 8.1, 6.3).
- [ ] GetVersionEx gives better compatibility for c-ares compared to VerifyVersionInfo. Moving to new function will loose compatibility with UWP apps (universal applications for winstore) and potentially with some extra old Windows/compilers.

bradh352:
If you want to update your PR that silences the warnings for the rest of the cases, I'll merge it.  I don't think its a big deal to keep using GetVersionEx.

kolomenkin:
Here is updated and rebased pull request

bradh352:
I do also see deprecation warnings for inet_addr(), then in the adig utility for gethostbyname() on windows, not sure if you wanted to address those too or let them ride.   They don't really bother me.

PR #57, opened 2016-06-29
add ares_process_timeouts

Garik-:
If you have two libraries c-ares and libev needed feature that extra ends the process.
The result was a situation that libev library comes ev_timer event (eg maximum time-out) and in the library of c-ares works ares_process function and there was no way to correctly close the connection.

example: https://github.com/Garik-/crawler/blob/libev/ev_ares.c

### DIFF
--- a/ares.h
+++ b/ares.h
@@ -420,6 +420,8 @@ CARES_EXTERN void ares_process_fd(ares_channel channel,
                                   ares_socket_t read_fd,
                                   ares_socket_t write_fd);
 
+CARES_EXTERN void ares_process_timeouts(ares_channel channel);
+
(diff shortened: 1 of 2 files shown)


Garik-:
OK I will try

dimbleby:
Isn't this just the same as calling `ares_process(channel, NULL, NULL)` (or `ares_process_fd(channel, ARES_SOCKET_BAD, ARES_SOCKET_BAD)`)?

bradh352:
@dimbleby I'd agree, I think the real issue is simply the manpage for ares_process() and ares_process_fd(), no need to add another API.

bradh352:
fixed via documentation update rather than adding a new API in ca118e9

PR #62, opened 2016-09-07
configure.ac: do not check for ar if specified manually

tkoecker:
The `configure` script of c-ares checks for the `ar` binary to use no matter if one was specified manually or not and ignores a manually specified one. With this change it only searches for it in case none was specified manually. This change is identical to what already is done in `configure.ac` of curl.


bagder:
Thanks!

PR #68, opened 2016-10-22
fix build on OpenBSD

qbit:
From the OpenBSD [getservent(3) man page](http://man.openbsd.org/OpenBSD-current/man3/getservent.3):

```
     getservbyport_r(int port, const char *proto, struct servent *servent,
         struct servent_data *servent_data);
```


bagder:
thanks!

PR #77, opened 2017-01-03
Updated CMake minimum version to 2.8.12.

stephen-sorley:
The proposed patch explicitly sets the minimum required CMake version to 2.8.12.  This version is new enough to allow the build system to be updated to follow standard practices for modern CMake.

Specifically, this patch changes the way usage requirements (include directories, compile definitions, dependent libraries) are specified. This involves using target-specific functions (target_include_directories, target_compile_definitions, etc.), along with the PUBLIC, PRIVATE and INTERFACE modifiers.

The patch also updates chain-building support to imitate modern find_package modules (import libraries), instead of old-style find_package modules (cache variables).  This new way of doing chain building is quite a bit simpler:

Old way to use bundled c-ares:
`if (CARES_FOUND)`
`include_directories(${CARES_INCLUDE_DIRS})`
`add_definitions(${CARES_DEFINITIONS})`
`target_link_libraries(myexe ${CARES_LIBRARIES})`
`endif ()`

New way (after patch) to use bundled c-ares:
`if (TARGET Cares::cares)`
`target_link_libraries(myexe PRIVATE Cares::cares)`
`endif ()`

@bradh352 - these are the changes we discussed last week


daviddrysdale:
@bradh352 or @gjasny, care to LGTM?  (I don't speak CMake)

CMakeLists.txt
```diff
@@ -580,28 +601,34 @@ IF (CARES_INSTALL)
...
+SET (CARES_FOUND 1 CACHE INTERNAL "CARES LIBRARY FOUND")
+SET (CARES_LIBRARIES Cares::cares CACHE INTERNAL "CARES LIBRARIES")
+
+
 # Build ahost
 ADD_EXECUTABLE (ahost ahost.c ${SAMPLESOURCES})
-SET_TARGET_PROPERTIES(ahost PROPERTIES COMPILE_FLAGS "-DHAVE_CONFIG_H=1 ${CARES_DEFINITIONS}")
-TARGET_LINK_LIBRARIES (ahost ${CARES_LIBRARIES} ${CARES_DEPENDENT_LIBS})
+TARGET_COMPILE_DEFINITIONS (ahost PRIVATE HAVE_CONFIG_H=1)
+TARGET_LINK_LIBRARIES (ahost PRIVATE Cares::cares)
```
gjasny: please don't link against the hackish alias targets here.

CMakeLists.txt
```diff
@@ -580,28 +601,34 @@ IF (CARES_INSTALL)
...
 	INSTALL (TARGETS ahost RUNTIME DESTINATION ${CARES_INSTALL_LOCATION_BIN})
 ENDIF ()
 
 
 # Build adig
 ADD_EXECUTABLE (adig adig.c ${SAMPLESOURCES})
-SET_TARGET_PROPERTIES(adig PROPERTIES COMPILE_FLAGS "-DHAVE_CONFIG_H=1 ${CARES_DEFINITIONS}")
-TARGET_LINK_LIBRARIES (adig ${CARES_LIBRARIES} ${CARES_DEPENDENT_LIBS})
+TARGET_COMPILE_DEFINITIONS (adig PRIVATE HAVE_CONFIG_H=1)
+TARGET_LINK_LIBRARIES (adig PRIVATE Cares::cares)
```
gjasny: and here

CMakeLists.txt
```diff
@@ -580,28 +601,34 @@ IF (CARES_INSTALL)
...
 	INSTALL (TARGETS adig RUNTIME DESTINATION ${CARES_INSTALL_LOCATION_BIN})
 ENDIF ()
 
 
 # Build acountry
 ADD_EXECUTABLE (acountry acountry.c ${SAMPLESOURCES})
-SET_TARGET_PROPERTIES(acountry PROPERTIES COMPILE_FLAGS "-DHAVE_CONFIG_H=1 ${CARES_DEFINITIONS}")
-TARGET_LINK_LIBRARIES (acountry ${CARES_LIBRARIES} ${CARES_DEPENDENT_LIBS})
+TARGET_COMPILE_DEFINITIONS (acountry PRIVATE HAVE_CONFIG_H=1)
+TARGET_LINK_LIBRARIES (acountry PRIVATE Cares::cares)
```
gjasny: and here

CMakeLists.txt
```diff
@@ -1,4 +1,11 @@
-CMAKE_MINIMUM_REQUIRED (VERSION 2.8)
+CMAKE_MINIMUM_REQUIRED (VERSION 2.8.12)
```
gjasny: Thank you for working on this. 2.8.12 is a lot more reasonable than 2.8.

stephen-sorley:
As requested, I've replaced uses of Cares::cares in the internal targets
with the PROJECT_NAME variable.  Anything else that needs changing?

Thanks,
Stephen S.

PR #80, opened 2017-01-08
Export CMake targets

gjasny:
With the recent raise of the minimum CMake version, exporting of targets became possible. Those exports could be easily imported within a CMake project by adding the following line:

```
find_package(c-ares)
```

The preparing commit utilises CMake's knowledge of library and include directories. So on RHEL lib64 would be used and on Debian lib/<architecture>.


gjasny:
CC: @bradh352

bradh352:
CC: @stephen-sorley

stephen-sorley:
This PR has two compile bugs on Windows.  I went ahead and made a patch against your PR that fixes them, please see attached:

[0001-Windows-build-fixes.patch.zip](https://github.com/c-ares/c-ares/files/696847/0001-Windows-build-fixes.patch.zip)

First, when installing shared libs, you should set RUNTIME, LIBRARY and ARCHIVE destinations.  This is because on windows, the DLL part of the shared library is installed to the RUNTIME destination (DLL's must be in same folder as exe's), and the import-lib part of the shared library is installed to the ARCHIVE destination.  Setting all three destinations ensures the correct behavior on all platforms.  In the patch, I just set all three destinations on every install(TARGETS...) command, using a helper variable called TARGETS_INST_DEST.

Second, if both static and shared libs are built, the output name of the static library needs to have the suffix appended, to avoid name-clashes with the import lib from the shared library.  This is because on Windows static libraries and import libraries have the same extension (.lib).  In the patch I have it adding the suffix for all platforms. Alternatively, you could choose to only add the suffix on Windows, it's up to you guys.

Other than that, LGTM.  The export stuff is pretty cool.

gjasny:
@stephen-sorley Thank you for reviewing and fixing. I will give the review two more days and merge by Friday.

PR #84, opened 2017-01-11
Always use check_symbol_exists instead of check_function_exists. 

stephen-sorley:
This PR replaces all uses of the check_function_exists() macro with check_symbol_exists(). It fixes a build error with MinGW 32-bit builds, but it applies to all platforms.

The check_function_exists() macro has two big issues:

(1) It can't detect functions that are specified in system headers as a #define macro.

(2) For 32-bit Windows builds, it can't detect functions declared with a different calling convention than __cdecl.  Since the Win32 API uses __stdcall, it can't detect anything from the Win32 API.  This is the true cause of the detection failure mentioned in the code comments describing CARES_FUNCTION_EXISTS.

The check_symbol_exists() macro fixes both of these problems. It solves (1) by checking to see if the function name is a defined macro.  It solves (2) by accepting a list of includes to add to the test source code. The linker is then involved in the check by getting the address of the function.


stephen-sorley:
@gjasny

stephen-sorley:
@bradh352

bradh352:
looks good to me

gjasny:
@bagder What's the policy with merging in c-ares? Should patches be cleaned up before merging?

bradh352:
@gjasny What do you mean by "cleaned up" in relation to this patch?

Unfortunately @stephen-sorley noticed I made a bad assumption with CheckSymbolExists, I thought it only validated the symbol (define or function) was in the headers and didn't do a link check, and for that CheckFunctionExists should be preferred.  However this turns out not to be the case, CheckSymbolExists does both and CheckFunctionExists does a link/library-only check in a flawed way.  Essentially, CheckFunctionExists looks like it really should be marked as deprecated in CMake as it does the wrong thing completely (though will appear to "work" in a large number of cases).

gjasny:
I mean you can see the work progressing in those commits. One commit adds the `SET (CMAKE_EXTRA_INCLUDE_FILES stdlib.h)` line, the other one removes it. In my company we ususally squash those patches together to reduce noise. But other teams have different views and actually like to keep the **whole** history without rewriting anything.

I was just asking what the c-ares project preferences are on that matter.

stephen-sorley:
I think github actually allows you to squash a pull request when you merge it, according to this:

[https://github.com/blog/2141-squash-your-commits](https://github.com/blog/2141-squash-your-commits)

If that doesn't work for you, I can squash it in my local repo and make another pull request.

bagder:
Yes, patches should be squashed properly before merging but may still be more than one commit if they make sense separately - and we prefer rebase instead of merge-commits.

stephen-sorley:
@gjasny 

I went ahead and squashed everything into 1 commit, then did a force push.  I think the PR is ready to merge, unless you noticed any other issues.

PR #90, opened 2017-03-11
Add support for Windows DNS Suffix Search List

ChristianAmmer:
This change should solve issue #53. Support for suffix lists in c-ares was already built-in for Linux. With this code change the function `get_SuffixList_Windows` was added which reads the search list from the registry key:
> HKLM\System\CurrentControlSet\Services\TCPIP\Parameters\SearchList

and latter updates the configuration with `set_search`.

The registry key `SearchList` should be [available since Windows 2000](https://support.microsoft.com/de-at/help/275553/how-to-configure-a-domain-suffix-search-list-on-the-domain-name-system-clients). If the `SearchList` could not be read or is empty, then the content from the registry key `Domain` is returned (the Primary DNS suffix).

### DIFF
--- a/ares_init.c
+++ b/ares_init.c
@@ -1093,6 +1093,52 @@ static int get_DNS_Windows(char **outptr)
   /* Fall-back to registry information */
   return get_DNS_Registry(outptr);
 }
+
+/*
+* get_SuffixList_Windows()
+*
+* Reads the "DNS Suffix Search List" from registry and writes the list items
+* whitespace separated to outptr. If the Search List is empty, the
+* "Primary Dns Suffix" is written to outptr.
+*
+* Returns 0 and nullifies *outptr upon inability to return the suffix list.
+*
+* Returns 1 and sets *outptr when returning a dynamically allocated string.
+*
+* Implementation supports Windows Server 2003 and newer
+*/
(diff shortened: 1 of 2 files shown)


bradh352:
It seems another patch serving the same purpose, but taking a different approach was submitted to the mailing list in January:
https://c-ares.haxx.se/mail/c-ares-archive-2017-01/0003.shtml

Care to comment on the differences?

ChristianAmmer:
The major difference I think is the source for the DNS Suffix Search List. I didn't pay attention for the Adapter Addresses (`IP_ADAPTER_ADDRESSES`) because it seems that nslookup seems to ignore them and therefore it's maybe good enough for c-ares?

If `Domain`, `DhcpDomain`, `SearchList` from `HKLM\...\Tcpip\Parameters` are set all to an empty string and instead set `Domain` and `SearchList` of the Ethernet Adapter, nslookup does not use any suffix. So I think it ignores the Adapter addresses at all. Maybe this is handled different in other Windows versions (I tried it on Windows 10)?

ChristianAmmer:
Closed because I will create a feature branch for this issue with a patch that now also supports Connection specific suffix lists.

PR #93, opened 2017-03-17
Added support for Windows DNS Suffix Search List

ChristianAmmer:
This change solves issue #53.

Support for suffix search lists was already built in for Linux. The search list could be set via `set_search`. With this change the suffix search list from Windows is read from the registry and then set into the ares configuration via `set_search`. There are two sources for the search list:

1. The global DNS suffix search list.
2. The primary and connection specific DNS suffixes if the global is not available.

### DIFF
--- a/ares_init.c
+++ b/ares_init.c
@@ -1323,10 +1323,10 @@ static void replace_comma_by_space(char* str)
 static bool contains_suffix(const char* const searchlist,
                             const char* const suffix, const size_t len)
 {
-  if (!*suffix)
-    return true;
   const char* beg = searchlist;
   const char* end;
+  if (!*suffix)
+    return true;
   for (;;)
   {
     while (*beg && (ISSPACE(*beg) || (*beg == ',')))
@@ -1430,7 +1430,7 @@ static int get_SuffixList_Windows(char **outptr)
       while (len = next_suffix(&pp, len))
       {
         if (!contains_suffix(*outptr, pp, len))
-          commajoin(outptr, pp, len);
+          commanjoin(outptr, pp, len);
       }
       ares_free(p);
       p = NULL;
--- a/ares_init.c
+++ b/ares_init.c
@@ -1302,6 +1302,102 @@ static int get_DNS_Windows(char **outptr)
   /* Fall-back to registry information */
   return get_DNS_Registry(outptr);
 }
+
+static void replaceColonBySpace(char* str)
+{
+  /* replace ',' by ' ' to coincide with resolv.conf search parameter */
+  char *p;
+  for (p = str; *p != '\0'; p++)
+  {
+    if (*p == ',')
+      *p = ' ';
+  }
+}
+
+/*
+* get_SuffixList_Windows()
+*
+* Reads the "DNS Suffix Search List" from registry and writes the list items
+* whitespace separated to outptr. If the Search List is empty, the
+* "Primary Dns Suffix" is written to outptr.
+*
+* Returns 0 and nullifies *outptr upon inability to return the suffix list.
+*
+* Returns 1 and sets *outptr when returning a dynamically allocated string.
+*
+* Implementation supports Windows Server 2003 and newer
+*/
+static int get_SuffixList_Windows(char **outptr)
+{
+  HKEY hKey, hKeyEnum;
+  char *p;
+  char  keyName[256];
+  DWORD keyNameBuffSize;
+  DWORD keyIdx = 0;
+  char *newbuf;
+  size_t newsize;
+
+  *outptr = NULL;
+
+  if (ares__getplatform() != WIN_NT)
+    return 0;
+
+  /* 1. Global DNS Suffix Search List */
+  if (RegOpenKeyEx(HKEY_LOCAL_MACHINE, WIN_NS_NT_KEY, 0,
+      KEY_READ, &hKey) == ERROR_SUCCESS)
+  {
+    if (get_REG_SZ(hKey, SEARCHLIST, outptr))
+      replaceColonBySpace(*outptr);
+    RegCloseKey(hKey);
+    if (*outptr != NULL)
+      return 1;
+  }
+
+  /* 2. Connection Specific Search List composed of:
+   *  a. Primary DNS Suffix */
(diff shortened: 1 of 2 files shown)


ChristianAmmer:
@bagder: I'm a new contributor to c-ares and don't know who to ask, can you suggest someone who could review this Pull Request?
I would appreciate if this Pull Request would be merged into master, as I think the library would benefit from it.

ChristianAmmer:
@bagder @yangtse @daviddrysdale you are first listed in the contributers list to ares_init.c. Can I request you as reviewer for this change?

bradh352:
I don't like the fact that your code is keeping the search list in a comma separated list and scanning those.  I think it makes the code harder to audit.  I'd prefer seeing an array of search lists, then as the last step, it join them into the space-separated list that set_search() requires.

Otherwise, I don't see any real issues, though I haven't personally tested.

I will point out that this is a completely different take to accomplish a similar goal:
https://c-ares.haxx.se/mail/c-ares-archive-2017-01/0003.shtml

Though that patch, while smaller, doesn't seem to include the actual search list, only suffixes.

ChristianAmmer:
@bradh352: You are right, it is different to [c-ares-archive-2017-01](https://c-ares.haxx.se/mail/c-ares-archive-2017-01/0003.shtml), it is more like [c-ares-archiv-2013-06](https://c-ares.haxx.se/mail/c-ares-archive-2013-06/0010.shtml) but it is also different to this. I think this approach fits perfectly into the c-ares initialization step as it does not need to change existing interfaces (like the 2017-01 patch does) and it stops adding more suffixes (like the 2013-06 patch does), if the global DNS suffix search list is available, which is in my opinion the correct behavior.

I have to think about converting the search list into an array and as last step back to an whitespace separated list. I'm not sure if it makes the change more readable.

ChristianAmmer:
@bradh352 Regarding a domain array:
- The first part is reading the "Global DS Suffix Search List". If this list is not empty, we only have to replace the commas by spaces and can return the list. Introducing an array in the first place is only overhead as we only have to change the commas by spaces.
- Now the second part is the "Connection Specific Search List". There are some sources where we get suffixes. Each source can be only one single suffix or a comma separated list of suffixes. Because it can be a comma separated list, and we do not want duplicates in the result, we have to tokenize each source (which is done by `next_suffix`) and check if this suffix is contained in the result (which is done by `contains`). If it is not conained we can add it to the result (which is done by `add_suffix`) Now let us consider where an array could be a benefit:

  1. Well `next_suffix` has to operate on an input csv string, this function wouldn't change.
  2. If `contains` would get an array instead a csv string, it would be easy to iterate through the array and check if an suffix is already contained. Here an array would make sense. But taking into account, that we already have a function `next_suffix`, it does not make a big difference because we can use this function also to iterate through the csv string.
  3. If `add_suffix` would add a suffix to an array, it would have to reallocate the array size (because we don't know the array size at the beginning) and it would have to allocate space for the string. Currently the function increases the space for the resulting string and concatenates the suffix. Here I don't see a benefit of an array.
  4. After there are no suffixes left, the array has to be converted into a space separated string. Which needs one more allocation. In my version no additional string allocation is necessary just a replacing of the commas by spaces. An array is here rather a disadvantage.

I don't see any advantage of using an array, what do you think?

bradh352:
looping back around on this since we're looking to get another release out.

1. If we don't switch this to an array, at least you should use the existing "commajoin()" function instead of your "add_suffix()" that appears to do the same thing so we don't have that duplicate code.  I do understand that you'll need to modify commajoin() to take a src_len, and modify the 2 use-cases to pass strlen(src) as the 3rd parameter, but I believe this is more desirable than 2 very similar pieces of code.

2. What are the reasons your contains() and next_suffix() also split on spaces and not just commas?  Am I misreading the code and this list would ever be space delimited at the point these functions are called?

3. Your indentation on the "return 0" for 2b is off.

4. Why if 2a fails do you not continue on to 2b?

5. Should we not combine 1 and 2 results?

ChristianAmmer:
1. Yes, that of course makes sense.
2. The reason is to skip whitespaces.
E.g. `abc.def, hij.klm` should not result in a second suffix with a leading space `<space>hij.klm`. I'm not sure if it is allowed therefore this additional test is maybe superfluous.
3. Yes, Indentation is wrong, thank you.
4. In the TCP/IP settings there is the Radio Button "Append primary *and* connection specific DNS suffixes". I did not find a reference for this but I tried it many times with any reasonable registry combination and validated the result with what we get when calling `ipconfig /all`. And yes, if there is no primary DNS suffix, then the Suffixes from the Adapters are ignored.
5. No they should not be combined. For this I found following [reference](https://technet.microsoft.com/en-us/library/dd197418(WS.10).aspx):
> When the suffix search list is empty or unspecified, the primary DNS suffix of the computer is appended to short unqualified names, and a DNS query is used to resolve the resultant FQDN. If this query fails, the computer can try additional queries for alternate FQDNs by appending any connection-specific DNS suffix configured for network connections.

The "suffix search list" in this paragraph is the "global DNS suffix search list" (the result of 1), and **only** if it is empty or unspecified...

I will update this Pull request after I had some time to fix 1 and 3. Thank you for your review and I'm happy seeing this Pull request getting merged.

bradh352:
If you can rename replaceColonBySpace() to replaceCommaBySpace() to reflect what it really does, and contains() to contains_suffix() I think I'm ok with merging it.

@bagder any feedback?

bradh352:
looks good to me.  @bagder, you ok if I merge this?

bagder:
@bradh352: absolutely! 👍

bradh352:
@ChristianAmmer sorry, one more bug.  I was doing final testing with VS2005 which doesn't support C99 and it found a mixed code and declaration in  contains_suffix()  please move the declarations for beg and end to the top of the function to resolve the issue.

Then also in get_SuffixList_Windows() you use commajoin() when it should be commanjoin().

Make those fixes and I can merge.

It'd be helpful if AppVeyor actually worked :)

ChristianAmmer:
`commajoin` how could that compile with 3 arguments, strange (at least for a C++ programmer like me)
@bradh352 good catch, thank you!

ChristianAmmer:
@bradh352 Great, Thank you!

PR #107, opened 2017-05-12
configure.ac: Allow AR to be overidden

orzen:
If an archiver is configured in with AR in the shell environment, it will be overidden by AC_PATH_TOOL's resolved archiver. This is normally based on the configure option --host alternatively TARGET_SYS. However there are cases when these two differ and configure should not ignore the archiver specified with AR. Corresponding fix has already been made in libcurl.


gjasny:
This looks like a duplicate of #62

orzen:
@gjasny : That seems correct. I apologize for the duplicate.

PR #112, opened 2017-06-14
Adding feature ares_getaddrinfo

ChristianAmmer:
Currently AF_UNSPEC is not well supported by ares. This Pull Request is a first attempt of an implementation of `ares_getaddrinfo`. Many details of the RFC described `getaddrinfo` are not implemented, however I think it's good enough for a starting point. It gets tested through:

> [ RUN      ] AddressFamilies/MockChannelTestAI.FamilyV6/0
[       OK ] AddressFamilies/MockChannelTestAI.FamilyV6/0 (1 ms)
[ RUN      ] AddressFamilies/MockChannelTestAI.FamilyV4/0
[       OK ] AddressFamilies/MockChannelTestAI.FamilyV4/0 (1 ms)
[ RUN      ] AddressFamilies/MockChannelTestAI.FamilyV4_MultipleAddresses/0
[       OK ] AddressFamilies/MockChannelTestAI.FamilyV4_MultipleAddresses/0 (1 ms)
[ RUN      ] AddressFamilies/MockChannelTestAI.FamilyUnspecified/0
[       OK ] AddressFamilies/MockChannelTestAI.FamilyUnspecified/0 (2 ms)


pmarks-net:
I strongly recommend not accepting this change until it implements RFC 6724.  Encouraging people to migrate from the standard `getaddrinfo()` to one lacking connectivity-aware address sorting would be directly harmful to global IPv6 adoption.

Address sorting is `getaddrinfo()`'s _raison d'etre_.

ChristianAmmer:
I did not create the Pull Request for an early merge, I created it to discuss, comment and contribute in it (according https://help.github.com/articles/about-pull-requests/)

bagder:
@ChristianAmmer I appreciate it a lot. So much I even mentioned this on the mailing list. You're addressing a shortcoming and a function people have been missing in c-ares for a long time.

Please don't take random comments as a sign that we don't appreciate your work. A function that isn't fully feature-complete compared to the "native" `getaddrinfo()` can still serve a purpose, even if I of course hope we can get in in par over time.

I don't agree with @pmarks-net on this topic. c-ares *already* hampers good IPv6 use due to its lack of this function so I'd claim adding this *increases* the chances of applications behaving better.

pmarks-net:
Let's say that this function resolves `www.kernel.org` (which has one IPv4 and one IPv6 address) from:

1. an IPv4-only client
2. an IPv4 client with 6to4/Teredo-tunneled IPv6
3. an IPv6-only client

For each case, what is the first address?  If I'm reading the code correctly, the current answer is "it's random", because it depends on the ordering of the DNS response packets.  So we would have IPv4 clients trying to connect to IPv6 addresses and vice versa, with users blaming IPv6 when it fails.

bagder:
> we would have IPv4 clients trying to connect to IPv6 addresses and vice versa

The ordering of the addresses in the ares_getaddrinfo() response is not what prevents that from happening. It is a properly written client code that does. There's nothing that preventing a client to do stupid things even when the list is sorted according to all specs.

And again, *without* this function I claim a client has a much harder time to do the right thing IPv4/IPv6-wise so I claim that even a getaddrinfo() that returns addresses in a bad order is a better situation than what we have before now...

Besides, we can all help out and make sure this new function works exactly the way it should!

pmarks-net:
> properly written client code

Let's suppose that a typical client does the following:

1. Call `ares_getaddrinfo()`
2. ???
3. Initiate a TCP connection

Assuming that the IPv4/IPv6 ordering is not predictable, do you have an example of what a properly-written client might do for step 2?

bagder:
I claim that a typical client would iterate over the list, no matter which order it comes in. But it is also not a very productive discussion so I'll drop out of it now.

pmarks-net:
I agree that a typical client would iterate over the list, as in [Beej's guide](http://beej.us/guide/bgnet/output/html/singlepage/bgnet.html#simpleclient).  This is basically the best-case scenario (except for Happy Eyeballs, which few developers implement because it's relatively complex.)

If a client is strictly IPv4-only or IPv6-only, then incorrect ordering only costs a few microseconds, for the `connect()` to fail and try the next address (though users may see confusing error messages in the process.)

But even for this ideal client implementation, random IPv4/IPv6 preference creates two problems:

- When a crufty home router advertises IPv4 alongside a 6to4 tunnel (`2002::/16`), or the OS provides a Teredo tunnel (`2001:0::/32`), [RFC 6724](https://tools.ietf.org/html/rfc6724) says to prefer IPv4, but random preference will send ~half the traffic through the tunnel, which is likely to be slow or blackholed (`connect()` may hang for a few minutes.)
- On a native dualstack network, you get the [hampered eyeballs](https://labs.ripe.net/Members/emileaben/hampered-eyeballs) scenario: RFC 6724 says to prefer IPv6, but random preference will send ~half the traffic via IPv4.  When the IPv4 path includes a Carrier Grade NAT, this consumes unnecessary state, decreasing the incentive for ISPs to deploy IPv6.

Random preference just seems like bad news for every scenario I can think of.  Does it have any advantages?

bradh352:
I think a getaddrinfo() is a good idea, even if it isn't fully conformant at this point in time.  It would still be useful for those of us who query both ipv4 and ipv6 and have to wait on both replies to proceed.  At least for me, it would cut out a synchronization step waiting for both replies which would be nice.  I do think that a man page is needed, I didn't see one in the commit, and it would be good to know the known deficiencies of the implementation (I haven't yet performed a full review).  

The only other thing I'm not sure of is if all platforms have struct addrinfo, I would bet they don't, and it appears this implementation would mean c-ares couldn't work on systems that don't.  Perhaps it would be wise to make a struct cares_addrinfo?  Heck, it may not be a bad idea anyhow as I'd love to have extensions like the "TTL" be returned so client-side caching can behave a little nicer.

In our current use of c-ares, we currently query for both ipv4 and ipv6 results separately for each host and wait for a response to each query.  We then round-robin from each bin, which is presumably in server-preference order, starting with ipv6 bin (then ipv4, then ipv6 again, etc), to form a list to try.   We then qsort the list based on a prior attempt's weight (if available) ascending, followed by original index when the list was created, ascending.  A lower weight is better (0=known good, 1=unknown/not tried, 2=slow, 3=bad)

We then cycle through the list starting non-blocking connections such that either when an error occurs or 100ms has elapsed, it will go to the next in the list until one succeeds (or there are none left).  For each condition we encounter we update our feedback cycle for the address for future sorting ... and of course don't cache a result more than 10 minutes to handle network changes.

Pretty sure that's Happy Eyeballs, if not, it works for us anyhow :)

pmarks-net:
> starting with ipv6 bin (then ipv4, then ipv6 again, etc)

This algorithm is reasonable when you run the client in an environment with known-good IPv6 connectivity, but if you release it to random users, it will incorrectly prefer 6to4/Teredo when present.  If 6to4/Teredo are slow but not 100% broken, then most connections will go via the tunnel, and performance will suffer.

> Pretty sure that's Happy Eyeballs

There are some similarities, but the [published algorithm](https://tools.ietf.org/html/rfc6555#section-6) for Happy Eyeballs depends on `getaddrinfo` doing a proper sort:

<pre>
   1.  Call getaddinfo(), which returns a list of IP addresses <b>sorted by
       the host's address preference policy.</b>

   2.  Initiate a connection attempt with the <b>first address</b> in that list
       (e.g., IPv6).

   3.  If that connection does not complete within a short period of
       time (Firefox and Chrome use 300 ms), initiate a connection
       attempt with the first address belonging to the other address
       family (e.g., IPv4).

   4.  The first connection that is established is used.  The other
       connection is discarded.
</pre>

ChristianAmmer:
Many Thanks @bagder for your motivating words and mentioning it in the mailing list. The discussion in this thread made it clear to me how important the sorting is, which I was not aware off.

But next to this there are other open points such as the not yet handled arguments `service` and the `hints` members `ai_socktype`, `ai_protocol` and `ai_flags`. I don't know which points are most important for the next steps and it would please me if someone would take an issue.

@pmarks-net the sorting function which seems not to be that easy to implement. I tried to understand the algorithm but some parts are unclear to me, maybe you can shed some light into. Mostly I ask myself how to get the source addresses. According to [RFC 6724](https://tools.ietf.org/html/rfc6724#page-10)
> It is RECOMMENDED that the candidate source addresses be the set of
   unicast addresses assigned to the interface that will be used to send
   to the destination (the "outgoing" interface).  On routers, the
   candidate set MAY include unicast addresses assigned to any interface
   that forwards packets, subject to the restrictions described below.

But how do we know which interface will be used, should we just try to iterate through the network interfaces and collect their addresses or should we try to get the routing table?

pmarks-net:
@ChristianAmmer, you can test whether a destination is reachable in the local routing table, and obtain its source address, using UDP `connect()` + `getsockname()`.  Generally, destination address selection is the resolver's responsibility, while source address selection is the kernel's.

See [Bionic getaddrinfo.c](https://android.googlesource.com/platform/bionic/+/master/libc/dns/net/getaddrinfo.c) for an example of a BSD-licensed C implementation of RFC 6724.  (But note that it uses a hard-coded preference table, and doesn't support parsing `/etc/gai.conf`.)

ChristianAmmer:
@pmarks-net: Thank you for the Link and your explanation, the glibc getaddrinfo.c file also makes it this way. The RFC says _This is simple and gives the best results_. But the document describes also another strategy:
> Another implementation strategy is to call down to the network layer to retrieve source address information and then sort the list of addresses directly in the context of getaddrinfo(). [...]
In this approach, the implementation might not have knowledge of the outgoing interface for each destination, so it MAY use a looser definition of the candidate set during destination address ordering.

I understand this in that way that we can also collect the available source addresses through iterating over the available interfaces. Although I'm not sure what they mean with "calling down the network layer".

pmarks-net:
"This is simple and gives the best results" refers to an implementation where both source *and destination* address selection happens in the kernel.  As far as I know, Linux doesn't provide an API like that.

"Call down to the network layer to retrieve source address information" seems to be describing the UDP `connect()` + `getsockname()` operation, which doesn't actually tell you which outgoing interface the connection will use.  But you don't need to know the destination interface if you're only implementing destination (not source) address selection.

I don't think `getifaddrs()` would be of much use here, because the important information is in the routing table, not the interface list.

halx99:
Hi,guys, what about this feature staus?

bagder:
I'm strongly in favor of getting this function added, even if it isn't 100% perfect in its first iteration. I don't believe in reaching perfect before merging.

halx99:
AND, I find a little bug, it can't be cancelled by ares_cancel, and I have made some changes at my fork repo: https://github.com/halx99/c-ares

ki11roy:
Are there any critical problems with the feature? How about merging it?

bradh352:
I'd be ok merging it if it had a man page, and also maybe declared current caveats in said man page.

That was the only thing holding me up from going ahead and merging it.

bradh352:
well, maybe one more flaw, the headers added to ares.h aren't good.  I think only   netdb.h   is even necessary as that is where the addrinfo struct should be defined, and it should be included where netinet/in.h is included ... since windows doesn't have it.

That said, some legacy systems that don't even have proper ipv6 support probably don't have the struct addrinfo either ...  so it may be more proper to create an   ares_addrinfo  struct  for people to use.  It might provide some flexibility on extensions (personally I'd love to add the TTL to the result for local caching).

ki11roy:
> well, maybe one more flaw, the headers added to ares.h aren't good. I think only netdb.h is even necessary as that is where the addrinfo struct should be defined, and it should be included where netinet/in.h is included ... since windows doesn't have it.
> 
> That said, some legacy systems that don't even have proper ipv6 support probably don't have the struct addrinfo either ... so it may be more proper to create an ares_addrinfo struct for people to use. It might provide some flexibility on extensions (personally I'd love to add the TTL to the result for local caching).

I have tried to fix that (and some other things) in another request, https://github.com/c-ares/c-ares/pull/232

ChristianAmmer:
@bradh352 thank you for pointing out the issue with the platform specific headers in `ares.h`. I fixed this and also added a first version of man pages.

About the `addrinfo` structure, I think c-ares should use the definition from the OS headers like it does with the `hostent` structure. According to [pubs.opengroup.org](http://pubs.opengroup.org/onlinepubs/009696699/basedefs/netdb.h.html)
> The <netdb.h> header shall define the addrinfo structure

On platforms with netdb.h we should be safe and for Windows c-ares includes the winsock headers.

bagder:
> I think c-ares should use the definition from the OS headers

That might be hard, because
1. They may differ and then we need to have different code for different weirdo structs
2. c-ares builds on systems that lack getaddrinfo and thus the define structs, but I can't think of a reason why the c-ares version can't be made to work there anyway

Of course, item (2) can be solved by just not provide the c-ares version on those legacy systems. It shouldn't affect *that* many users....

ki11roy:
@ChristianAmmer @bagder guys I am terribly sorry for the huge commit, but there are manuals and ares_addrnfo added and many more in the branch above https://github.com/c-ares/c-ares/pull/232

ChristianAmmer:
I agree and it should also not be a downside to the customer of `ares_getaddrinfo` when he has to use an c-ares defined `addrinfo` struct.

@ki11roy I appreciate your work and your engagement on this function becoming a part of c-ares. Participating to this is quite important because I'm not sure when I'm able to implement the address sorting and the other features according to the `hints` argument which are currently not supported. But it's not up to me to decide if and which Pull Request to be merged.

bagder:
IMHO, missing some functionality is not a reason to block the merge. It should rather be documented clearly so that users won't be surprised. For lots of applications, said sorting is not that important anyway.

ki11roy:
@ChristianAmmer @bagder well, what I want the least is to measure which pull request is a better candidate to merge, anyway it doesn’t matter because as I said my branch is based on this one. Let’s have this functionality anyway.

bradh352:
Without a deeper code review at this point of #232 to see why decisions, this PR looks less invasive from a cursory glance, especially since #232 breaks ABI/API for existing public functions.

pmarks-net:
> For lots of applications, said sorting is not that important anyway.

It is unreasonable to dismiss a proper Internet Standard as "not that important."  I suggest naming the function `ares_getaddrinfo_unsorted()` to prevent confusion.

ki11roy:
As a side note it is better also to have a sort as a public function if possible, because some applications have addrinfo cached so no sorting applies anyway while looking into cache.

bradh352:
I'm pretty happy with this pull request after the latest changes, despite its non-conformance with the RFC as I see those as things that we can probably overcome prior to the 1.16 release.  Its hard to take massive PRs that would implement it all and expect a sane code review anyhow, which is really part of the problem with #232, along with the fact that it breaks API/ABI of public functions (regardless of if we think anyone is using them or not).

However, #232 does implement additional features like honoring of the service and hints parameters.  @k11roy if I accept this pull request, would you be willing to re-base those features without modification of public API/ABI?

ki11roy:
@bradh352 sure, no problem.

ChristianAmmer:
@bradh352 I'm happy to read that this Pull Request got a status were it's likely that it gets merged.
From my side it can be merged.

bradh352:
ok, great.  Merging now, then we'll get in some additional features in other PRs like from @ki11roy

PR #123, opened 2017-07-03
hiearchy: introducing lib/ docs/ examples/ include/

bagder:
lib/      - the library source code
 docs/     - man pages and documentation
 examples/ - ahost, acountry etc
 include/  - public header(s)

The rest remains as it was.


dimbleby:
I'm all in favour of this reorganisation, but: any chance of doing a pass through the existing open pull requests and either accepting, rejecting, or otherwise progressing them first?

I imagine this is going to cause some irritating merge conflicts and it would be better to sort those once rather than piecemeal.

(Of course I particularly have in mind my own #96, though there are others that should be no harder to advance and are ahead of me in the queue...)

bagder:
I still think this is a good idea, I just failed to deliver. Might come back on this another time.

PR #124, opened 2017-07-05
Fix NMake build due to missing ares_build.h

user-none:
ares_build.h is auto generated these days and is created by the build system. This change "generates" ares_build.h by copying the pre-generated one (ares_build.h.dist) which is used for build systems like NMake.

### DIFF
--- a/Makefile.msvc
+++ b/Makefile.msvc
@@ -390,7 +390,10 @@ install:
 ALL: c-ares acountry adig ahost
     @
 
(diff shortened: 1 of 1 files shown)


dimbleby:
doesn't `buildconf.bat` already [do that](https://github.com/c-ares/c-ares/blob/245edc4fa4ebaee2fa43d787924e4e1fd66af56b/buildconf.bat#L16)?  

Per [instructions](https://github.com/c-ares/c-ares/blame/245edc4fa4ebaee2fa43d787924e4e1fd66af56b/INSTALL.md#L28).

user-none:
@coveralls, You are correct. I missed the buildconf.bat.

PR #130, opened 2017-07-19
Distribute CMake files

gjasny:


bagder:
we should add a cmake build from tarball to travis to verify the cmake based distribution...

bradh352:
didn't see any reason not to merge and figured we could go ahead and close this before next release.

PR #132, opened 2017-07-22
allow cmake_debug_postfix

ksophocleous:
allow both debug and release versions of the library to be installed simultaneously

it adds a debug postfix (`d`) to the library names (incl. dynamic libraries)
this is an option and it is `OFF` by default.

this is useful for people who work in IDEs like Visual Studio; switching to debug will allow them to debug and step into this project without fuss.


ksophocleous:
That did not cross my mind at all.. I feel bad now.

closing this as it's unnecessary

PR #133, opened 2017-07-22
allow linking against the static msvc runtime library

ksophocleous:
the option `CARES_MSVC_STATIC_RUNTIME` only exists when configuring for Visual Studio compilers and it is `OFF` by default.


cmake/msvc_static_runtime.cmake
```diff
@@ -0,0 +1,8 @@
+FOREACH (flag_var
+    CMAKE_C_FLAGS CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
+    CMAKE_C_FLAGS_MINSIZEREL CMAKE_C_FLAGS_RELWITHDEBINFO)
+
+    IF (${flag_var} MATCHES "/MD")
+        STRING (REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
+    ENDIF (${flag_var} MATCHES "/MD")
```
gjasny: Please leave endif brackets empty.

cmake/msvc_static_runtime.cmake
```diff
@@ -0,0 +1,8 @@
+FOREACH (flag_var
+    CMAKE_C_FLAGS CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
```
gjasny: Instead of hardcoding configurations, could you please use CMAKE_CONFIGURATION_TYPES ?

ksophocleous: yes, very good idea.. done

cmake/msvc_static_runtime.cmake
ksophocleous: right, also done

ksophocleous:
@gjasny do you want me to squash these two commits?

ksophocleous:
the change is done, do you want me to squash the commits?

PR #138, opened 2017-08-14
Make gethostbyaddr fail with `ECANCELLED` for `ares_cancel()`

addaleax:
When `ares_cancel()` was invoked, `ares_gethostbyaddr()` queries would fail with `ENOTFOUND` instead of `ECANCELLED`.

It seems appropriate to treat `ares_cancel()` like `ares_destroy()`, but I would appreciate review of the correctness of this change.

Ref: https://github.com/nodejs/node/issues/14814

### DIFF
--- a/ares_gethostbyaddr.c
+++ b/ares_gethostbyaddr.c
@@ -157,7 +157,7 @@ static void addr_callback(void *arg, int status, int timeouts,
         }
       end_aquery(aquery, status, host);
     }
-  else if (status == ARES_EDESTRUCTION)
+  else if (status == ARES_EDESTRUCTION || status == ARES_ECANCELLED)
     end_aquery(aquery, status, NULL);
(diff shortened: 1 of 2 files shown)


dimbleby:
If this is correct, then presumably `ares_gethostbyname()` should have a similar fix [here](https://github.com/c-ares/c-ares/blob/33bf5ba35e2df09ecde9530816733965bb4053e4/ares_gethostbyname.c#L218).

addaleax:
@dimbleby I don’t think so, for `ares_gethostbyname()` the status is already passed along via `next_lookup()`; I *assume* there is a reason for that disparity in the first place.

addaleax:
Bump, anything I can do for this to be merged? Should I ping the mailing list?

bagder:
Thanks!

PR #142, opened 2017-09-06
Force using the ANSI versions of WinAPI functions

ngg:
When compiling c-ares with a build system that defines UNICODE,
bad versions of WinAPI functions are used causing failures or even
crashes. When windows.h is included in MBCS mode (like in the default
build system), the ..A versions are the same as using the one without
any suffix.


ngg:
We use a separate build system as we want to compile cares and curl for 10+ platforms with a single unified build system together with other libraries and our own project.
In the past we faced lots of issues when some libraries are built with UNICODE and some with MBCS mode, so we want to build everything in UNICODE mode. We patched this for ourselves but thought it might cause issues for other users as well, that's why I submitted this PR.

Issue #134, opened 2017-07-26
Windows: Slow resolution due to querying of non-existing autoconf IPv6 servers


nmaier:
Affected
---
- c-ares 0.13.0 w/ system DNS server list
- Windows 7 (other Windows versions not tested, but potentially affected as well)

Problem
---
c-ares regressed in lookup time from a few milliseconds to >10s.

Pesky IPv6 DNS Servers (or: The cause)
---
Starting with #81 (or rather 1.13.0) I reliably and always observe massive delays from c-ares (via https://github.com/aria2/aria2/) running on Windows 7 when letting c-ares pick the name servers itself by querying the Windows configuration.

I tried looking into it, and c-ares now picks up 3 additional IPv6 DNS server addresses:
- fec0:0:0:ffff::1
- fec0:0:0:ffff::2
- fec0:0:0:ffff::3

Those are site-local addresses that get autoconfigured in Windows for IPv6 whenever a NIC has no DHCP or static configured DNS servers, but IPv6 connectivity in theory. It seems it's Microsoft's attempt trying to implement https://tools.ietf.org/html/draft-ietf-ipv6-dns-discovery-07 (which they also authored)

So where are these coming from exactly?
Adding some `printf`s to `get_DNS_AdaptersAddresses` shows those DNS servers are configured for the "Loopback Pseudo-Interface 1" and "VirtualBox Host-Only Network" interfaces and picked up by c-ares from there.

It gets worse... c-ares "detects" a metric of `-1` for all DNS servers incl. the non-functioning ones, and `qsort`reliably sorts the bad ones first in my configuration, causing c-ares to first query three servers on sitelocal hosts that do not even exist, resulting in **multi-second delays of up to 14 seconds** in my tests. The final dns server string c-ares generates for me is:
`fec0:0:0:ffff::1,fec0:0:0:ffff::2,fec0:0:0:ffff::3,127.0.0.1`

Remedies
---

- Do not even consider the loopback interface, e.g. by amending the check for not-up interfaces to also skip loopbacks.
```c
    if (ipaaEntry->OperStatus != IfOperStatusUp ||
        ipaaEntry->IfType == IF_TYPE_SOFTWARE_LOOPBACK)
        continue;
```
- Addressing virtual interfaces such as "VirtualBox Host-Only Network" is a bit harder, tho
  - Ignore all interfaces without a `GatewayAddress`?
  - Explicitly blacklist those three IPv6 addresses?
[...]

PR #144, opened 2017-09-10
Win32 exclude legacy ipv6 subnets

bradh352:
win32 ipv6: add infrastructure to exclude ipv6 subnets that are known to cause issues
Fixes issue #134


bradh352:
@bagder @daviddrysdale any issues with me merging this?

PR #152, opened 2017-10-19
__system_property_get not available on arm64

eskimor:
Removing the code on arm64. It does not link otherwise (symbol not found).

As this code is only a fallback, I believe it should be fine to just remove it for arm64. But please check!


bradh352:
Thanks for the report. commit 2250b59 should accomplish what you are looking for but instead does so by checking for symbol availability.

Issue #155, opened 2017-11-27
ares_init(&channel) returns unexpected status ARES_EBADNAME


cy20lin:
While trying to initialize channel with following code,

```c
    #include <ares.h>
    #include <stdio.h>

    int main() {
        int status;
        ares_channel channel;
        status = ares_library_init(ARES_LIB_INIT_ALL);
        printf("ares_library_init() status : (%d) %s\n", status, ares_strerror(status));
        if (status) {
            return 1;
        }
        status = ares_init(&channel);
        printf("ares_init(&channel) status : (%d) %s\n", status, ares_strerror(status));
        if (status) {
            ares_library_cleanup();
            return 1;
        }
        ares_destroy(channel);
        ares_library_cleanup();
        return 0;
    }
```

the function ares_init(&channel) returns ARES_EBADNAME (status=8).

The outputs:

```txt
    ares_library_init() status : (0) Successful completion
    ares_init(&channel) status : (8) Misformatted domain name
    Error: init_by_defaults failed: Misformatted domain name
```

This returned status is not any one of the four possible values as described in the [document](https://c-ares.haxx.se/ares_init.html).
(i.e. ARES_SUCCESS, ARES_EFILE, ARES_ENOMEM, ARES_ENOTINITIALIZED)

I'm on Windows 7, [MSYS2](http://www.msys2.org) mingw64 Shell, using
mingw-w64-x86_64-gcc version 7.2.0 and 
mingw-w64-x86_64-c-ares version 1.13.0

cy20lin:
After further diving into the source, it reveals that the error status is returned if the gethostname() failed in the function init_by_default()

The error comes form the implementation, init_by_defaults() in file ares_init.c.

In the function init_by_defaults():

If HAVE_GETHOSTNAME is defined,
the impl would try to retrieve the hostname using gethostname() (ares_init.c:1969),
and set channel->domains if certain condition is met,
otherwise it would fail with ARES_EBADNAME if gethostname() fails for some reason. This is where the error ARES_EBADNAME comes from.

On the other hand, if HAVE_HOSTNAME is not defined,
it just simply sets the channel->ndomains to 0 and continues with successive init operations.

IMO, the function init_by_defaults() should not fail the entire initialization process just because of a bad hostname. Instead, it should continue and ignore the bad hostname if the hostname isn't helpful.

cy20lin:
Actually, the gethostname() fails because of lack of initializing winsock api.

[...]

PR #156, opened 2017-11-27
ares_init.c: Don't return ARES_EBADNAME

cy20lin:
Make init_by_defaults() continue and ignore the bad hostname if the hostname from gethostname() isn't helpful, rather than failing the whole initialization process with error ARES_EBADNAME.

See issue #155 for details


bradh352:
alternative fix committed

Issue #157, opened 2017-11-27
Suggest adding WSAStartup/WSACleanup into c-ares library init/cleanup procedure on Windows


cy20lin:
Recently, trying to use c-ares on Windows with no lucks (#155). After doing some research, it turns to out be the reason that winsock isn't initialized with WSAStartup() in advance, therefore c-ares functions don't work properly on Windows.

As a new user of c-ares, it is quite frustrating to run such simple code and only to find it doesn't works on Windows (though it works on linux). It is not explicitly mentioned in the [document](https://c-ares.haxx.se/ares_library_init.html) that WSASetup() should be called to initialize winsock before using c-ares functions. I just pretend this kind of setup would be taken care in the function ares_library_setup(), but it goes wrong, and takes me some times to look into the source codes and figure out what goes wrong.

Rather than letting users to initialize winsock on their own, it would be great if c-ares can take the responsibility to initialize winsock in ares_library_init() (through WSAStartup()) and clean it up in ares_library_clelanup() (throught WSACleanup()). As those are of platform dependent codes, by wrapping those into c-ares library init/cleanup would definitely increase protability across platforms. 

If there are any good reasons for c-ares not to take the responsibility of setting/cleaning up winsock, it will also helps to remind users in the document, that winsock should be manually started up before using any c-ares functions, and be manually cleaned up afterwards.

p.s. There've been discussions about this winsock init/deinit topic in mailing list back to the year 2005, but it seem to be with no conclusion. [reference: c-ares on windows - error 11](https://c-ares.haxx.se/mail/c-ares-archive-2005-06/0004.shtml)

bradh352:
In my personal opinion, since in order to perform other network operations on Windows, like connect to the host returned by c-ares after a DNS query, you must also perform WSAStartup() .... I think this should be up to the application developer to handle this.  

I could see updating the documentation for ares_library_init() to indicate on Windows that WSAStartup() is *not* called and that it is up to the caller to initialize system networking before invoking c-ares.

PR #158, opened 2017-11-27
Add WSAStartup/WSACleanup into library init/cleanup procedure on Windows.

cy20lin:
See issue #157


bradh352:
PR #180 is the right way to go

PR #164, opened 2017-12-20
Fix computation of IPv6 blacklist mask for values of netmask > 8.

b-spencer:
Upstream c-ares PR #144 attempts to filter out certain blacklisted IPv6 DNS resolver addresses from those returned by Windows.

The `ares_ipv6_subnet_matches` function computes the full netmask from a bit count argument via a loop.  However, this loop gives the wrong answer when mask > 8. It needn't even be a loop.  This PR fixes the code so it should compute the right netmask for all values from 0 to 128.

BTW, I discovered that the same bad IPv6 resolver addresses ("fec0::") can also appear in Android simulators.  Perhaps these addresses should be generally blacklisted for all operating systems?

### DIFF
--- a/ares_init.c
+++ b/ares_init.c
@@ -1009,8 +1009,8 @@ static int ares_ipv6_subnet_matches(const unsigned char netbase[16],
   memset(mask, 0xFF, netmask / 8);
 
   /* Set remaining bits */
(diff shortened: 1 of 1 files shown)


bradh352:
i agree, your method is much cleaner and the original solution was definitely broken, odd.

PR #168, opened 2017-12-30
CMake: Add tests

gjasny:


gjasny:
- [ ] Wait until #167 is merged
- [ ] Implement symbol hiding for CMake as it's done with autoconf, rebase, retest

bradh352:
@gjasny Thanks for doing this, I really had no idea where to start as I'm not familiar with gmock.  

It looks to me like if the compiler isn't C++11 compatible (which is obviously not a requirement for c-ares itself), the entire build will fail due to the tests directory being included.  Instead, I think if the compiler isn't C++11 compatible, that we should just skip including the tests directory.

I think it can be detected with  $<COMPILE_FEATURES:cxx_std_11> as per
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#manual:cmake-generator-expressions(7)
and
https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

Unfortunately, cxx_std_11 wasn't introduced until CMake 3.8 ... so I'm not exactly sure how to handle that ... maybe require CMake 3.8+ to include the tests directory in the first place ... or rely on a c++11-specific feature that did exist in older CMake releases -- but could be error prone for compilers that were in-process but not fully c++11-compliant?

gjasny:
I just notices that minimum CMake has been raised to 3.2.

bradh352:
@gjasny 

I think we should just rename `CARES_WITH_TESTS` to `CARES_TESTS` or `CARES_BUILD_TESTS`.

Also, we should hide `CARES_TESTS` behind:
`
IF (NOT CMAKE_VERSION VERSION_LESS 3.8 AND $<COMPILE_FEATURES:cxx_std_11>)
`
... I think that will work, I haven't personally tried the `$<COMPILE_FEATURES:...>` syntax, but it does says `cxx_std_11` was added in CMake 3.8.

gjasny:
@bradh352 I'm no big fan of automatically disabling tests if test requirements are not met. If the user does not notice the disabling, it might give a false sense of security.

My approach has two advantages: We could keep CMake 3.1 requirement and the user will get an error if CMake is not able to enable required C++11 mode. In case the users environment is not sufficient, the user will have to manually disable the tests.

Does that make sense?

bradh352:
@gjasny Then you probably need to disable tests by default.  This is what had to happen with the autotools build as enabled by default caused a lot of issues.  In autotools, we enabled by default *if* we could detect it was a C++11 compiler, otherwise we disabled by default but you could try to turn it back on.

So maybe the solution is to do the `IF (NOT CMAKE_VERSION VERSION_LESS 3.8 AND $<COMPILE_FEATURES:cxx_std_11>)`  JUST for the default value of the test enablement.

gjasny:
I guess I'll just disable them. The problem I see with your approach is that one has to enable c++ language before one could test for the feature macros. And enabling a language cannot be optional right now. I guess there are some embedded platforms around where you don't have a C++ compiler around.

bradh352:
looks good to me if you want to merge

PR #182, opened 2018-02-11
Use cmake3 package provided by Ubuntu

gjasny:
I think it's better to depend on a stock Ubuntu package than a PPA.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,9 @@ addons:
     apt:
         sources:
             - ubuntu-toolchain-r-test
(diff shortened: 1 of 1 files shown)


bradh352:
what version of cmake3 is that though, we need at least cmake 3.8 for PR #168 for running tests on travis

gjasny:
The `ppa:adrozdoff/cmake` repo just provides 3.7. I hope we could stick to `3.1`

bradh352:
you're right, I thought the ppa provided the latest cmake, but no, its quite old.  I think we can probably live with cmake 3.1.

PR #184, opened 2018-03-22
add CMAKE_DEBUG_POSTFIX so the debug lib is given a different name

JoeStaines:
useful for using both the release and debug libs in the same directory


bradh352:
Seems like a very windows-y thing to do.  I could see an IF (WIN32) around this or something being ok. Haven't really seen that common for Linux/Unix development methodologies though.

Do you have any background on why this is important to you?

JoeStaines:
Yeah it was for Windows pretty much.

Yeah it can easily be left as a command line argument for the user. I don't think its uncommon to add a debug postfix though (although in some libraries i've seen it applied at the install point rather than the build point). Its up to you guys, i can put the if WIN32 around it or close the PR if you don't think its suitable.

PR #193, opened 2018-05-17
Apply the IPv6 server blacklist to all nameserver sources, not just Windows

b-spencer:
For #164, I mentioned that it seemed like the IPv6 nameserver blacklist should apply to all OSes.  In [a mailing list post](https://c-ares.haxx.se/mail/c-ares-archive-2017-12/0001.shtml), @bradh352 agreed and suggested that I file a PR to make it so.  This is that PR.  Sorry for the delay.

This moves the blacklist check from being Windows-specific to being a general feature of `config_nameservers()`, no matter the nameserver source.  It also simplifies the `ares_ipv6_server_blacklisted()` implementation to not parse and re-parse the blacklisted IPv6 addresses from strings on every check.  I think they're almost as easy to read as a sequence of hex bytes in an array initializer, and it's definitely less work on each trip through the code.

The new unit test for this behaviour passes on Linux x86_64.

### DIFF
--- a/ares_init.c
+++ b/ares_init.c
@@ -997,63 +997,6 @@ static int compareAddresses(const void *arg1,
   return 0;
 }
 
-/* Validate that the ip address matches the subnet (network base and network
- * mask) specified. Addresses are specified in standard Network Byte Order as
- * 16 bytes, and the netmask is 0 to 128 (bits).
- */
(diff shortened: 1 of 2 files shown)


bradh352:
looks good, I'll merge once travis finishes successfully

PR #204, opened 2018-07-05
ares_gethostbyname: fake hostent for AF_UNSPEC

Seraphime:
Currently, the conditional which decides whether the specified host
is an IPv4 address verifies whether family is AF_INET or AF_INET6
which is equivalent to saying family is AF_UNSPEC, but users of
AF_UNSPEC cannot fake IPv4 hosts.


Seraphime:
The reason the coverage has fallen by 0.002% is because the if checking for IPv4 addresses is no longer executed for family == AF_INET6, which is nonsensical anyways.

PR #205, opened 2018-07-11
Remove tabs

ruslo:


bradh352:
Can you elaborate on why you think this change is proper.  It appears the file uses tabs for indention and space for alignment.  Is there some official best-practice from the makers of CMake to use spaces and not tabs?

ruslo:
C code use spaces everywhere as far as I can see:
* https://github.com/c-ares/c-ares/blob/master/acountry.c#L41-L47

I was thinking that it make sense to make it consistent with CURL CMake code (no tabs, lower case):
* https://github.com/curl/curl/blob/089833147420e1b6018c79d1929a8284b1b8aa8a/CMakeLists.txt#L258-L267

If you're not interested feel free to close.

PR #221, opened 2018-09-04
Add configurability for path to resolv.conf file (#220)

### LINKED ISSUE #220: allow path for resolv.conf file to be configurable
Hi Team,

While working on [this issue](https://github.com/nodejs/node/pull/22226), I came to understand that the path to the `resolv.conf` file is hardcoded, via `./configure`.

For testing purposes, we would like to launch a child process with a different path for the `resolv.conf` file (via an environment variable like `CARES_RESOLV_CONF_PATH`, say), where we would then be able to specify different `resolv.conf` files (with different `search`, `nameserver`, etc. stanzas, say).

I'd like to work on adding this support, but wanted to ask if such a PR would be welcome?

SirR4T:
- [x] By extending `struct ares_options` and `struct ares_channeldata`, and similarly `ares_init_options()`, `ares_save_options()`, and `init_by_resolv_conf()`
- [x] Added test case, for *nix, within `CONTAINED_TEST_F` (could not test locally, as local machine is darwin, which does not support linux LXC)
- [ ] Add support for windows (How? also, cannot test locally)
- [x] No new warnings with `--enable-maintainer-mode` and `--enable-debug`
- [ ] Documentation needs to be updated.

Fixes #220

### DIFF
--- a/ares_init.c
+++ b/ares_init.c
@@ -1662,6 +1662,7 @@ static int init_by_resolv_conf(ares_channel channel)
     size_t linesize;
     int error;
     int update_domains;
+    const char *resolvconf_path;
 
     /* Don't read resolv.conf and friends if we don't have to */
     if (ARES_CONFIG_CHECK(channel))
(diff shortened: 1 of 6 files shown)


ares_init.c
```diff
@@ -1652,43 +1670,90 @@ static int init_by_resolv_conf(ares_channel channel)
...
           break;
+        default:
+          DEBUGF(fprintf(stderr, "fopen() failed with error: %d %s\n",
+                        error, strerror(error)));
+          DEBUGF(fprintf(stderr, "Error opening file: %s\n", PATH_RESOLV_CONF));
+          status = ARES_EFILE;
+        }
       }
-      fclose(fp);
     }
```
bradh352: Too much duplication here of logic.  If someone explicitly configured a resolveconf_path, I don't see any reason to use the default PATH_RESOLV_CONF.  So really we should just do this once.  I'd think it makes more sense to error out anyhow.

bradh352:
regarding windows support, this would be non-windows only ... simply putting it in the docs is sufficient.

SirR4T:
Great, will do that. Also, anything to be done towards the Travis build failures? I'm facing the following:

```
libcares_la-windows_port.gcno:no functions found
{u'message': u"Couldn't find a repository matching this job.", u'error': True}
```

bradh352:
looks like travis succeeded to me.  Where did you see this?

SirR4T:
Ah, my bad. This was against my fork, in my travis account.

SirR4T:
Hi @bradh352 , just pushed some documentation to the man pages, please review.

bradh352:
Thanks

PR #239, opened 2019-01-16
Support for ares__sortaddrinfo

ki11roy:
This is a port of RFC 6724 compliant sorting function from Android Bionic project:
https://android.googlesource.com/platform/bionic/+/e919b116d35aa7deb24ddece69c491e24c3b0d6f/libc/netbsd/net/getaddrinfo.c

The latest version is essentially the same, except two additional parameters to test connection with (mark/uid):
https://android.googlesource.com/platform/bionic/+/master/libc/dns/net/getaddrinfo.c

Please note that even that version has some restrictions, which I stated in the man. It doesn't support some rules from RFC 6724:
 - Rule 3 (Avoid deprecated addresses)
 - Rule 4 (Prefer home addresses)
 - Rule 7 (Prefer native transport)

There are no tests, because it is a port.


bradh352:
What is the argument for making this a public function rather than just running it internally to ares_getaddrinfo() ?

ki11roy:
I do not really like the fact that to sort the results I need to make multiple connections somewhere, so I thought it would be a good idea to make a flag or a separate function for those who do not want this behavior. On the other hand there are no much use of getaddrinfo without sorting, because that is the main idea of the function and the original one is not public. Do you think it’ll be better to hide it?

bradh352:
If you'd like to add a flag to the ares_getaddrinfo() on whether or not the output should be sorted, in my mind, that's the best solution.  Most will probably want it sorted though.  I'd prefer the sort function itself to be strictly used internally, if there is a usecase in the future we could look at exposing it, but I don't see any reason to speculatively add to the public namespace.

bagder:
I'm totally with @bradh352 on this. Functions should only be added to the public API when there's a clear need and desire for them. We have not heard anyone express wanting to sort the results separate from invoking this function.

In this case, `ares_getaddrinfo()` feels like a function that primarily gets added to the c-ares API pool to offer as an in-place replacement for the regular `getaddrinfo()`. I believe a large portion of the future users of this function want as similar functionality as possible, including sorting.

gvanem:
@ki11roy 
I've not tried you patch, but I do note your use of `qsort()`. This is a potential **Danger zone** since I remember we  had problems with this before. Refs: 
* Pull https://github.com/c-ares/c-ares/pull/81
* And @dimbleby's comment at https://github.com/c-ares/c-ares/commit/39aeafd27ddabfcdcaa406ffb585d918bee7263e.

ki11roy:
@gvanem thanks for the warning! I think everything will be fine since there are:
```
   * Rule 10: Leave the order unchanged.
   * We need this since qsort() is not necessarily stable.
   */
  return a1->original_order - a2->original_order;
```

ki11roy:
@bradh352 sure

PR #20, opened 2014-09-05
Support DNSSEC + DANE

nmav:
This series of patches adds support for DNSSEC + parsing of DANE structures.

Changes:
    To add support for sending a query that will set the flags needed by dnssec I had to add the new ares_create_query2().

These patches do add raw DNSSEC support and DANE. They do not add support for reading the trusted dnssec servers (done as separate pull requests).

PR #243, opened 2019-04-12
Added support for DNS AD and EDNS DO flags

eriklax:
This pull requests adds functionality to allow a more fine grained control of which DNS header flags to set. This pull requests focuses on flags for requesting DNSSEC validation status.

- The AD flags on requests asks for the AD flag in the response [1].
- The EDNS DO flag on requests asks for the AD flag and DNSSEC RRs in the response [1].

There have been a previous attempt (https://github.com/c-ares/c-ares/pull/20) to include a similar feature in c-ares. That pull request added a new function ares_create_query2() and a generic DNSSEC flag (which set the EDNS DO flag).

This pull request changes the use of the "int rd" argument to ares_create_query/ares_mkquery() into a "flags" type of argument. This should work without regressions as long as the rd was treated as either 0 or 1. There is of course no guarantees that all external usages of ares_create_query() worked that way but a all usages I've found (on GitHub) used a constant of 0 or 1.

I'll happily update with documentation changes and tests (where we see fit) if this pull request would be accepted.

[1] https://labs.apnic.net/?p=316


bradh352:
I'm not too familiar with DNSSEC (I know, horrible), but wouldn't this trigger additional response data to be returned?  How is this to be handled or verified.  

The current docs say zero/non-zero not zero/one for the 'rd' flag, so it definitely has API and ABI implications to have this change, but I do agree most likely people really do use zero/one.  We'll need to think about it from that standpoint a bit more.

eriklax:
I agree with the concerns regarding the API change, but I do believe reusing this argument really makes sense, and I hope it can be justified. As "rd" is a DNS header flag so evolving this argument to a flag would allow c-ares to add this and more flags in the future.

As for any additional data returned (eg DNSSEC RRs); I'd suggest that c-ares do not do any verifications and just let it be ignored since c-ares is currently not a validating resolver and should probably not be.

If anyone would like to use this additional data, it's very easy to parse this data from the result in the callback using any DNS parsing framework available.

The important part of this pull request is _being able_ (emphasized) to set the AD flag (request DNSSEC status) and being able to read the AD flag (access the DNSSEC status). In the result callback I could simply do

```
bool dnssec = false;
HEADER* hdr = (HEADER*)abuf; // DNS header
if (hdr && hdr->ad == 1)
  dnssec = true;
```

An option could be to have a generic callback function available used by the ares_query code path to do last minute changes to the DNS packet before sending it out on the wire. But I do believe being able to set flags should not require raw packet modifications if possible.

bradh352:
Please see the new DNS message formatter and parser currently in main.  It should become public in the c-ares 1.22, and should give you control over these flags. You wouldn't use the ares_create_query() legacy function, you'd use the ares_dns_record_create() and all the associated helpers currently found in ares_dns_record.h.

As for the DNSSEC RR's you referred to, we don't currently have parsers for any of the DNSSEC RRs, if you had a list of RRs you're looking for we can add those onto the task list.  Writing new parsers is trivial with the new API.  You can reference #600 as an example.

PR #264, opened 2019-06-24
Fix leak and crash in ares_parse_a/aaaa_reply

ki11roy:
Fixes leak and crash in ares_parse_a/aaaa_reply. Borrowed fuzz files from #263.

@bradh352 By the way, I thought of moving common code somewhere in ares_parse_a_aaaa_reply.inc and including it into ares_parse_a_reply.c directly, what do you think of that?


bradh352:
I agree they're quite common.  I don't know what you mean by a .inc file, but you could definitely probably combine the functionality into a single conversion function, with a one-liner wrapper that is the parse_a/aaaa_reply.

PR #270, opened 2019-08-05
Add ttl field to ares_srv_reply

dschopf:
This commit simply adds a new field to the ares_srv_reply struct and fills it with the TTL field of the corresponding RR.

### DIFF
--- a/ares.h
+++ b/ares.h
@@ -534,6 +534,7 @@ struct ares_srv_reply {
   unsigned short          priority;
   unsigned short          weight;
   unsigned short          port;
+  unsigned int            ttl;
 };
 
 struct ares_mx_reply {
--- a/ares_parse_srv_reply.c
+++ b/ares_parse_srv_reply.c
@@ -49,7 +49,7 @@ int
 ares_parse_srv_reply (const unsigned char *abuf, int alen,
                       struct ares_srv_reply **srv_out)
 {
-  unsigned int qdcount, ancount, i;
+  unsigned int qdcount, ancount, i, rr_ttl;
   const unsigned char *aptr, *vptr;
   int status, rr_type, rr_class, rr_len;
   long len;
@@ -103,6 +103,7 @@ ares_parse_srv_reply (const unsigned char *abuf, int alen,
         }
       rr_type = DNS_RR_TYPE (aptr);
       rr_class = DNS_RR_CLASS (aptr);
+      rr_ttl = DNS_RR_TTL (aptr);
       rr_len = DNS_RR_LEN (aptr);
       aptr += RRFIXEDSZ;
       if (aptr + rr_len > abuf + alen)
@@ -145,6 +146,7 @@ ares_parse_srv_reply (const unsigned char *abuf, int alen,
           vptr += sizeof(unsigned short);
           srv_curr->port = DNS__16BIT(vptr);
           vptr += sizeof(unsigned short);
+          srv_curr->ttl = rr_ttl;
 
           status = ares_expand_name (vptr, abuf, alen, &srv_curr->host, &len);
           if (status != ARES_SUCCESS)
--- a/test/ares-test-parse-srv.cc
+++ b/test/ares-test-parse-srv.cc
@@ -12,7 +12,7 @@ TEST_F(LibraryTest, ParseSrvReplyOK) {
   pkt.set_qid(0x1234).set_response().set_aa()
     .add_question(new DNSQuestion("example.com", ns_t_srv))
     .add_answer(new DNSSrvRR("example.com", 100, 10, 20, 30, "srv.example.com"))
-    .add_answer(new DNSSrvRR("example.com", 100, 11, 21, 31, "srv2.example.com"));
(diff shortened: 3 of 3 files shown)


dschopf:
Both failed builds are from system setup failures, not from my changes. Unfortunately I don't know how to trigger a rebuild without changing the pull request.

bradh352:
I like the idea of having ttl availability... however, this does break ABI for c-ares which I can't recall the last time that happened.  Any thoughts on this ?

bagder:
The previous SONAME bump was made in 2007 in commit e64b10330300.

In general, we should avoid bumping it - if it can be avoided - as distros and many users consider it a huge nuisance.

dschopf:
So the alternative would be to add a new function as a kind of "overload" which takes an additional argument for the ttl. But that would make the ttl values disconnected from the ares_srv_reply linked list.

Or just add a completely unrelated function which takes a new struct ares_srv_reply_ttl which is defined in the same way as in my change.

Which approach would you prefer or do you have a better solution?

bradh352:
I prefer option #3, create new parser functions that rely on opaque structures for all the various parsers and provide 'getters' to fetch the values from the opaque structure to prevent future ABI breakage.

But... that's probably more than you want to do :)

dschopf:
I would be willing to put some time into it. Maybe I can start with the ares_parse_a_reply function and see how much work it actually is. 

One question: when looking at the hostent struct
```
           struct hostent {
               char  *h_name;            /* official name of host */
               char **h_aliases;         /* alias list */
               int    h_addrtype;        /* host address type */
               int    h_length;          /* length of address */
               char **h_addr_list;       /* list of addresses */
           }
```
how do you actually want to have get getters for arrays? One for each member which returns the actual member? Or something more sophisticated where you can ask for the number of items and one function where you can then retrieve the value?
So either:
```
char** get_aliases(opaque_ptr* p);
```
or
```
size_t get_aliases_count(opaque_ptr* p);
char* get_aliases_entry(opaque_ptr* p, size_t index);
```

bradh352:
The second would be the proper choice.  Of course the get should return const char *.

dschopf:
I have made a new branch with an attempt for a new opaque pointer interface here: 
[dschopf/c-ares/devel](/dschopf/c-ares/tree/devel)

Is this roughly what you had in mind? 
The new interface is very similar to the existing API, I don't know if this good or bad.

bradh352:
Yep, pretty much.  The only comments I'd make are 

1) typedef the struct (but do NOT hide the pointer in the typedef as that would prevent constification)

2) So we don't have 2 parsers, the old parser needs to be rewritten to just wrap the new parser (and yes, this does mean an internal memory allocation is performed that previously would not have been, but I think that's ok as the other interfaces we'd want to mark as legacy).

3) Internal usage of reply parsing should use the new functions not legacy.

4) I wonder if it makes sense to split out a separate allocation function for the ares reply, and instead of the parser allocating, you pass in the pre-allocated pointer.  This way someone conscious of memory allocations could use the same pointer over and over.  Would probably need a 'clear' too incase there are internal memory allocations due to chaining in some results.

5) I wonder if there's enough overlap to have a single 'reply' opaque structure that all parsers share.

Obviously we'd probably need others to chime in on 4 & 5....

dschopf:
While working on 2. I have come across the following issue:

Let's assume that the legacy functions will just be wrappers for the new functions which just call the new parse functions. Where can I store the pointer for the internally created ares_a_reply structure so it can be passed to the ares_free_a_reply function in the ares_free_hostent call? The only possibilities that I see is either before or after the hostent or ares_addrttl structures. Is that acceptable or do you have any other suggestions? To me this seems rather questionable.

bradh352:
You wouldn't cache the new result at all, you'd copy all the members into the legacy struct so the existing free functions (ares_free_hostent, ares_free_data) work as they have in the past.

dschopf:
New commit which addresses the remaining points. The code for the A reply should now be complete except for the callback from ares_gethostbyname, but that needs also AAAA support. And than there is the name, which probably should be changed.

Can you give some advice about the callback interface? If there is no common reply structure for A and AAAA records how would you design the callback function?

dschopf:
Looks like I was a little bit too eager. I thought that it would be nice to have a new ares_gethostbyname_ex with a new callback signature which returns a reply struct instead of a hostent. But looking back that was probably a bit too much.

bradh352:
getaddrinfo is the newer recommended way ... and it hasn't seen an official release yet

orivej:
For the reference, a copy of c-ares in pycares has added support for TTL in these commits:
https://github.com/saghul/pycares/commit/8d4783d1a4272cf96d300428bd2da7e7445c12ad
https://github.com/saghul/pycares/commit/908a6af807ffd9baff6c7a9aacf25d4429b985dd
https://github.com/saghul/pycares/commit/dbfeafca21e6f58e7d5b6f8254e8d021b8967127#diff-47a1012510ade8b182937f35e2385847

bradh352:
due to ABI breakage, this PR cannot be accepted.

PR #272, opened 2019-08-09
gethostbyname: reload resolv.conf values if file changed

nayana-sue:
There is unhandled ares use case when we call ares_init when /etc/resolv.conf
is empty and later do the DNS request when it is filled in.
The bug is that ares does not detect change of resolv.conf.
The result is that default DNS IP is used, which is 127.0.0.1

Solution for this problem is that we check if resolv.conf file changed
each time we call ares_gethostbyname function.
Similar solution is done in libc.

Another unhandled use case is that system changes resolve.conf for some reason
and ares would not be aware of this change silently doing wrong DNS requests.

Signed-off-by: Nayana Topolsky <nayana.topolsky@streamunlimited.com>

### DIFF
--- a/ares.h
+++ b/ares.h
@@ -339,6 +339,8 @@ CARES_EXTERN const char *ares_version(int *version);
 
 CARES_EXTERN int ares_init(ares_channel *channelptr);
 
+CARES_EXTERN int ares_reinit_by_resolv_conf_file(ares_channel channel);
+
 CARES_EXTERN int ares_init_options(ares_channel *channelptr,
                                    struct ares_options *options,
                                    int optmask);
--- a/ares_init.c
+++ b/ares_init.c
@@ -1683,6 +1683,21 @@ static int init_by_resolv_conf(ares_channel channel)
 
     fp = fopen(resolvconf_path, "r");
     if (fp) {
+      /* Every time this init function is called, update the stat values */
+      struct stat st;
+      if (stat(resolvconf_path, &st) == 0)
+      {
+#ifdef __APPLE__
+        channel->resolvconf_mtime = st.st_mtimespec;
+        channel->resolvconf_ctime = st.st_ctimespec;
+#else
+        channel->resolvconf_mtime = st.st_mtim;
+        channel->resolvconf_ctime = st.st_ctim;
+#endif
+        channel->resolvconf_size = st.st_size;
+        channel->resolvconf_ino = st.st_ino;
+      }
+
       while ((status = ares__read_line(fp, &line, &linesize)) == ARES_SUCCESS)
       {
         if ((p = try_config(line, "domain", ';')) && update_domains)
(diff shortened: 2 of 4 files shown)


nayana-sue:
I just have one little concern, I ifdef HAVE_SYS_STAT_H of including of sys/stat.h header but
do not ifdef all stat/timespec structure usage, should I do that?
The CI passed, but maybe you do not have CI for win32 etc.

ares.h
```diff
@@ -339,6 +339,8 @@ CARES_EXTERN const char *ares_version(int *version);
 
 CARES_EXTERN int ares_init(ares_channel *channelptr);
 
+CARES_EXTERN int ares_reinit_by_resolv_conf_file(ares_channel channel);
```
bagder: I think this should be just "ares_reinit"

- not all systems may have resolv_conf so implying that in the symbol name seems wrong
- not all systems may reinit from a"file" so lets not say so either

This then allows other systems to do whatever is necessary there to do the equivalent.

nayana-sue: Thanks for feedback.
Maybe the function has wrong name and is on wrong place.

I just wanted a function that would internally reload solely the resolv.conf file if it is used by that particular system. It probably does not belong to ares.h - I just wanted to access it easily from ares_gethostbyname.c.

But maybe other systems like Windows can change DNS information on the fly (in registers?) or something and it would have the same problem. 
But I do not own Windows or other platforms, just Linux.

So I can probably rename the function to ares_reload_resolv_conf() and move it to non public place (but where?). And then what do I do with other platform implementations?
I could make some effort and try to implement things blindly, not sure if its good idea.

bradh352:
Please make it not specific to just your needs, and make sure it re-initializes other systems like windows too.  I don't see getting this merged otherwise.

bradh352:
closed due to lack of activity based on provided feedback

ares_init.c
```diff
@@ -1845,6 +1860,87 @@ static int init_by_resolv_conf(ares_channel channel)
...
+  channel->nsort = -1;
+
+  /* clear options */
+  channel->ndots = -1;
+  channel->timeout = -1;
+  channel->tries = -1;
+  channel->rotate = -1;
+
+  int rc = init_by_resolv_conf(channel);
+  if (init_by_resolv_conf(channel) != ARES_SUCCESS) {
```
zhihaoy, line 1930: Why `init_by_resolv_conf` is called twice here?

zhihaoy (commented):
We applied the patch and confirmed that it works. But given the patched library based on 1.17.1, we saw a crash on this line: https://github.com/c-ares/c-ares/blob/cares-1_17_1/src/lib/ares_process.c#L1449
So `server->qhead` or `server->next` may be uninitialized or dangling in some situations. We don't know whether the crash is caused or enabled by this patch.

ares_init.c
nayana-sue: Sorry, that is is a typo.

nayana-sue:
> We applied the patch and confirmed that it works. But given the patched library based on 1.17.1, we saw a crash on this line: https://github.com/c-ares/c-ares/blob/cares-1_17_1/src/lib/ares_process.c#L1449
> So `server->qhead` or `server->next` may be uninitialized or dangling in some situations. We don't know whether the crash is caused or enabled by this patch.

Thanks for the info on the crash.
Any specific use case? When did you "saw" the crash?

ares_init.c
zhihaoy: I drafted a patch and attempt to make this part more transactional, i.e., don't update timestamp unless `init_by_resolv_conf` succeeds.
I'm not sure whether that can resolve the crash. In theory, if the first call in the typo succeeds and the second call fails, `servers[0]->qhead` could be uninitialized while `nservers > 0`, but this is only my guess.

zhihaoy:
> > We applied the patch and confirmed that it works. But given the patched library based on 1.17.1, we saw a crash on this line: https://github.com/c-ares/c-ares/blob/cares-1_17_1/src/lib/ares_process.c#L1449
> > So `server->qhead` or `server->next` may be uninitialized or dangling in some situations. We don't know whether the crash is caused or enabled by this patch.
> 
> Thanks for the info on the crash.
> Any specific use case? When did you "saw" the crash?

The stack trace looks like the following:
```
0   end_query (ares_process.c:1450)
1   next_server (ares_process.c:804)
2   ares__send_query (ares_process.c:873)
3   next_server (ares_process.c:791)
4   ares__send_query (ares_process.c:873)
5   next_server (ares_process.c:791)
6   ares__send_query (ares_process.c:873)
7   next_server (ares_process.c:791)
8   ares__send_query (ares_process.c:873)
9   ares_send (ares_send.c:137)
10  ares_query (ares_query.c:144)
11  ares_search (ares_search.c:73)
12  next_lookup (ares_gethostbyname.c:162)
13  ares_gethostbyname (ares_gethostbyname.c:135)
14  Curl_resolver_getaddrinfo
15  Curl_resolv
16  create_conn
17  Curl_connect
18  multi_runsingle
19  curl_multi_perform
```
The line numbers are off-by-one. You can tell that it recursively retries UDP and fails when finally gave up. Again, I don't know whether the patch causes this, but unfortunately, our test suite can't proceed far enough without the patch :/

zhihaoy:
Forgot to update here. We figured out the cause and fixed it. The problem is that `init_by_defaults` may add servers, but `init_by_defaults` can also add a local server (which doesn't work) when no network interface presents. In this case, needs to call `ares__init_servers_state`, otherwise `servers[0].qhead` is uninitialized, causing `end_query` to crash on dereferencing invalid pointers.

```diff
@@ -1911,11 +1911,13 @@ int ares_reinit_by_resolv_conf_file(ares_channel channel)
     channel->resolvconf_size = st.st_size;
     channel->resolvconf_ino = st.st_ino;
   }
-  ares__init_servers_state(channel);
 
   /* for any options that were not loaded from resolv.conf file,
      make sure they have at least default values */
-  return init_by_defaults(channel);
+  init_by_defaults(channel);
+
+  ares__init_servers_state(channel);
+  return ARES_SUCCESS;
```

nayana-sue:
> Forgot to update here. We figured out the cause and fixed it. The problem is that `init_by_defaults` may add servers, but `init_by_defaults` can also add a local server (which doesn't work) when no network interface presents. In this case, needs to call `ares__init_servers_state`, otherwise `servers[0].qhead` is uninitialized, causing `end_query` to crash on dereferencing invalid pointers.
> 
> ```diff
> @@ -1911,11 +1911,13 @@ int ares_reinit_by_resolv_conf_file(ares_channel channel)
>      channel->resolvconf_size = st.st_size;
>      channel->resolvconf_ino = st.st_ino;
>    }
> -  ares__init_servers_state(channel);
>  
>    /* for any options that were not loaded from resolv.conf file,
>       make sure they have at least default values */
> -  return init_by_defaults(channel);
> +  init_by_defaults(channel);
> +
> +  ares__init_servers_state(channel);
> +  return ARES_SUCCESS;
> ```

Thanks for this, we may pick this fix up.
Did you have some kind of good reproducible test case? I guess this was on linux?

zhihaoy:
macOS >= Big Sur (older version of macOS doesn't deterministically reproduce). It needs to be a long-running program that relies on auto-reloading DNS. I forgot the exact scenario. But you can try these:
1. Start the program with all network interface are off, do query, exit from the program that performs de-initialize.
2. Start the program with all network interface are off, do query, enable a network interface, query again, exit.

PR #274, opened 2019-09-07
Accept invalid /etc/resolv.conf lookup values

bnoordhuis:
Before this commit invalid `lookup` values resulted in c-ares not using
any lookups without any clear indication why. After this commit it uses
the default "fb".

(I'm all ears if there is a better way to fix this.)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -34,6 +34,10 @@ add_executable(arestest ${TESTSOURCES} ${TESTHEADERS})
 target_include_directories(arestest PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 target_link_libraries(arestest PRIVATE caresinternal gmock)
 
+IF (CARES_BUILD_CONTAINER_TESTS)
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ add_executable(arestest ${TESTSOURCES} ${TESTHEADERS})
 target_include_directories(arestest PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 target_link_libraries(arestest PRIVATE caresinternal gmock)
 
+IF (CARES_BUILD_CONTAINER_TESTS)
+  target_compile_definitions(arestest PRIVATE HAVE_USER_NAMESPACE HAVE_UTS_NAMESPACE)
+ENDIF ()
```
bnoordhuis: I figured it'd be safe to assume container namespaces are supported when people explicitly pass the option on the command line.

bradh352:
seems reasonable

PR #280, opened 2019-10-14
Rename local macros and remove unused local variables

oontvoo:
- Macros currently conflict with system ones.
- A couple of variables are unsued.

(Issues discovered once compiled with `[-Werror,-Wmacro-redefined,-Wunused-variable]`  )


oontvoo:
@daviddrysdale

PR #281, opened 2019-10-14
Move variables into the block where it is used to avoid unused-vars

oontvoo:
Warning uncovered with [-Werror, -Wunused-variables]


oontvoo:
@daviddrysdale Second PR (for moving variable defs )

bradh352:
This breaks c89 due to mixed code and declarations.  Please fix.  Either don't move the vars and wrap them in #ifdef HAVE___SYSTEM_PROPERTY_GET .... or move them into the if (status != ARES_EOF) block as they can be at the top of any block.

oontvoo:
@bradh352 Done. PTAL. Thanks!

PR #288, opened 2019-11-14
Fixing name resolution under win10 x64

pgurenko:
While using c-ares for internal project, got a problem of names not resolving on Windows 10 x64 native (Linux/Mac/.NET x86&x64 worked fine).

It turned out, some compilers just do not null the high part of the `size_t uns` in `nlen` union leading to ares_malloc failure (nlen.uns gets too big).

Simple memset fixes the problem.

### DIFF
--- a/ares_expand_name.c
+++ b/ares_expand_name.c
@@ -73,6 +73,8 @@ int ares_expand_name(const unsigned char *encoded, const unsigned char *abuf,
      size_t uns;
   } nlen;
 
+  memset(&nlen, 0, sizeof(nlen));
+
   nlen.sig = name_length(encoded, abuf, alen);
   if (nlen.sig < 0)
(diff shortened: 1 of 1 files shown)


bradh352:
This doesn't make sense, nlen and uns should be the same size. This to me implies ares_ssize_t and size_t aren't the same size on your system which would mean something is wrong with the build system in defining ares_ssize_t.

What compiler are you using?  Is this targeting 32bit or 64bit? What did CARES_TYPEOF_ARES_SSIZE_T get set to in ares_build.h?  Which build system are you using (nmake, cmake, autoconf)?

pgurenko:
Ah, my bad: I am using single cmake-generated vcxproj and ares_build.h + ares_config.h for both x86 and x64 for Windows, thanks for looking into it!

Issue #286, opened 2019-11-11
RFC conformance of `ares_getaddrinfo` regarding `AI_CANONNAME` and result structures.


ChristianAmmer:
An `ares_getaddrinfo` which is compliant regarding input and output (flags, names, result) of the function compared to other `getaddrinfo` implementations would greatly improve acceptence of this function. That should not be underestimated.

In this sense I suggest considering the naming of the `ares_getaddrinfo` result structures and the effect of the flag `ARES_AI_CANONNAME`.

The `ares_getaddrinfo` result structures are

- [`ares_addrinfo_node`](https://github.com/c-ares/c-ares/blob/master/ares.h#L585)
- [`ares_addrinfo_cname`](https://github.com/c-ares/c-ares/blob/master/ares.h#L601)
- [`ares_addrinfo`](https://github.com/c-ares/c-ares/blob/master/ares.h#L608)

The current implementation has the [RFC-3493](https://tools.ietf.org/html/rfc3493) `addrinfo` members stored in `ares_addrinfo_node`. The `ares_addrinfo` on the other hand is a structure which serves as container holding two lists. The first list contains a linked list of CNAME records, and the other contains the real `getaddrinfo` result.
This is confusing. I'm voting for renaming `ares_addrinfo_node` to `ares_addrinfo` and `ares_addrinfo` to some other name e.g. `ares_addrinfo_result`.

Additionally the flag `ARES_AI_CANONNAME` has the effect that a linked list of CNAME records is returned in the result. The RFC-3493 `AI_CANONNAME` on the other hand has following definition: 

>   If nodename is not null, and if requested by the AI_CANONNAME flag,
>   the ai_canonname field of the first returned addrinfo structure shall
>   point to a null-terminated string containing the canonical name
>   corresponding to the input nodename; if the canonical name is not
>   available, then ai_canonname shall refer to the nodename argument or
>   a string with the same contents.  The contents of the ai_flags field
>   of the returned structures are undefined.

Unfortunatelly `ai_canonname` was removed from the ares addrinfo structure and it's also not obvious how to get the RFC `ai_canonname` out of the `ares_addrinfo_cname` structure without deeper knownledge of CNAME chains. I would appreciate if the ares addrinfo structure would include a RFC conformant `ai_canonname` member.

[...]

PR #291, opened 2019-11-18
Rename `ares_getaddrinfo`

ChristianAmmer:
This PR addresses #286 and is suggesting some renamings regarding the `getaddrinfo` feature. One is the function name `ares_getaddrinfo` itself. I'm suggesting the new name `ares_getaddrinfo_ex` instead. The reason for this renaming is because 
> the only thing we'd accept that would accomplish your actual goal of "RFC Conformance" from an API prototype standpoint would be a wrapper to better emulate getaddrinfo(), and I'd think that wrapper should rely on the system's struct addrinfo if it exists, or the wrapper as a whole would not be available 

The most appropriate name for a future [RFC-3439](https://tools.ietf.org/html/rfc3493 "Basic Socket Interface Extensions for IPv6") conformant wrapper would be `ares_getaddrinfo`. And to have this name reserved, we need this renaming. 

The other renamings are addressing the structures:
- `ares_addrinfo` to `ares_addrinfo_ex` because for the same reason as for the function renaming
- `ares_addrinfo_node` to `ares_addrinfo` to be conformant to the RFC
- replace `ares_addrinfo_hints` by `ares_addrinfo` to be conformant to the RFC

FYI: @bradh352

### DIFF
--- a/ares.h
+++ b/ares.h
@@ -283,7 +283,7 @@ struct hostent;
 struct timeval;
 struct sockaddr;
 struct ares_channeldata;
-struct ares_addrinfo_result;
+struct ares_addrinfo_ex;
 struct ares_addrinfo;
 
 typedef struct ares_channeldata *ares_channel;
@@ -316,7 +316,7 @@ typedef int  (*ares_sock_config_callback)(ares_socket_t socket_fd,
 typedef void (*ares_addrinfo_callback)(void *arg,
                                    int status,
                                    int timeouts,
-                                   struct ares_addrinfo_result *res);
+                                   struct ares_addrinfo_ex *res);
 
 CARES_EXTERN int ares_library_init(int flags);
 
@@ -388,7 +388,7 @@ CARES_EXTERN void ares_getaddrinfo_ex(ares_channel channel,
                                    ares_addrinfo_callback callback,
                                    void* arg);
 
-CARES_EXTERN void ares_freeaddrinfo(struct ares_addrinfo_result* ai);
+CARES_EXTERN void ares_freeaddrinfo(struct ares_addrinfo_ex* ai);
 
 /*
  * Virtual function set to have user-managed socket IO.
@@ -605,7 +605,7 @@ struct ares_addrinfo_cname {
   struct ares_addrinfo_cname *next;
 };
 
-struct ares_addrinfo_result {
+struct ares_addrinfo_ex {
   struct ares_addrinfo_cname *cnames;
   struct ares_addrinfo  *nodes;
 };
--- a/ares.h
+++ b/ares.h
@@ -283,8 +283,8 @@ struct hostent;
 struct timeval;
 struct sockaddr;
 struct ares_channeldata;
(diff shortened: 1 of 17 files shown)


ki11roy:
Why that hints removal? How it will simplify things from the user perspective? Now there are more fields to set. Result postfix seems misleading either.

I suggest renaming the whole function set into something different and not bound to “addrinfo” at all, like “lookup” for example, as in https://node.readthedocs.io/en/latest/api/dns/ for example.

ChristianAmmer:
@ki11roy it will make it more transparent for the user to replace the function. Consider a user switching to c-ares. For example he has some code like:
```cpp
struct addrinfo hints = { .ai_family = PF_UNSPEC };
getaddrinfo("www.example.com", NULL, &hints, &result);
```
It's transparent to switch because he only has to append the "ares" namespace and of course the additional callback arguments, which can't be changed:
```cpp
struct ares_addrinfo hints = { .ai_family = PF_UNSPEC };
ares_getaddrinfo(channel, "www.example.com", NULL, &hints, callback, arg);
```
And as you see, there are not more fields to set.

ki11roy:
For example (using the before-the-renaming terms):
ares_addrinfo -> ares_lookup_result
ares_addrinfo_node -> ares_lookup_record or ares_lookup_node
ares_addrinfo_cname -> ares_lookup_cname

So it will look like:
```
struct ares_lookup_hints hints = { .ai_family = PF_UNSPEC };
ares_lookup(channel, "www.example.com", NULL, &hints, callback, arg);
ares_freelookup(lookup)
```

So there is no connection with addrinfo anymore and we can keep hints as is just to not mess with the ares_lookup_record.

It will be ok then to reintroduce wrappers which will match the original RFC API.

ChristianAmmer:
@ki11roy Your "lookup" suggestion goes to far, the `getaddrinfo` connection would be lost.

ki11roy:
Correct, that’s the main idea and it matches with the https://node.readthedocs.io/en/latest/api/dns/
I believe the guys had the same naming problem and their naming seems pretty good to me.

bradh352:
I never personally did understand the hints being shared with getaddrinfo(), it tells the user less about the interface looking at the prototype, as you don't know what members need to be set or why without referencing the docs explicitly.

That said isnt ai_flags an input param only for hints, I see it in the output of the original too which i'm not sure makes sense.

ChristianAmmer:
Even if `getaddrinfo` supports a special `hints` structure containing only the input parameters, I would have to read the documentation of `hints`. Because even if I know, that there is an input parameter "ai_flags", I really don't know what to do with it. 
But as a user, familiar with `getaddrinfo`, I would be more satisfied if `hints` is just similar (also by name) to what I know from `getaddrinfo`. Therefore I completly removed this special structure in favor for the regular `addrinfo` structure.

bradh352:
since we released 1.16, the API/ABI is now set in stone.  Closing this.

Issue #300, opened 2020-01-31
Regression in ares_parse_a_reply/ares_parse_aaaa_reply


AdamMajer:
When `ares_parse_a_reply` or `ares_parse_aaaa_reply` is called in case where an A **and** AAAA responses exist, the resulting ares_addrttl count is invalid and the structure points to gibberish.

There is no problem in latest released c_ares 1.15. It only exist on the master branch. Running reproducer under valgrind with old version produces no errors while complains about access to uninitialized memory under master.

Reproducer at https://gist.github.com/AdamMajer/fa1808b6c4751638f30e1bd7085bc64b

Edit: Updated to include data in the example

PR #302, opened 2020-02-03
Only count valid addresses when response parsing

AdamMajer:
When ares_parse_a_reply or ares_parse_aaaa_reply is called in case
where another AAAA and A responses exist, the resulting ares_addrttl
count is invalid and the structure points to gibberish.

This is a regression since 1.15.


AdamMajer:
This fixes https://github.com/c-ares/c-ares/issues/300

ChristianAmmer:
@ki11roy could you review this PR I think you know the code well.
fyi: @bradh352

ki11roy:
The code is already reviewed, anyway looks good to me.

PR #304, opened 2020-02-26
Return ENODATA if no valid A or AAAA record found

### LINKED ISSUE #303: CNAME with no associated A record returns success status
I have a DNS record which returns a CNAME pointing to a non-existent domain: an example is `broken.thomaswalton.co.uk`.

Using `ares_gethostbyname` with an A record request for this domain returns a status of `ARES_SUCCESS` even though no record was found.  The status should surely be `ARES_ENODATA` instead.

This can be fixed by something like https://github.com/tjwalton/c-ares/commit/06fcfd0a49f1c2bf4884d42e86c64fcb7c551a03 but I suspect something more nuanced may be better.

tjwalton:
Leave `ares_parse_a_reply` behaviour unchanged.
Update tests to check for the now intended behaviour.
Fixes #303.

### DIFF
--- a/ares_gethostbyname.c
+++ b/ares_gethostbyname.c
@@ -211,6 +211,13 @@ static void host_callback(void *arg, int status, int timeouts,
           if (host && channel->nsort)
             sort6_addresses(host, channel->sortlist, channel->nsort);
         }
+      if (status == ARES_SUCCESS && host && host->h_addr_list[0] == NULL)
+      {
+        /* The query returned something but had no A/AAAA record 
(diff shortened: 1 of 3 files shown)


bradh352:
in `TEST_F(LibraryTest, ParseAReplyNoData) {` it looks like you're missing a `ares_free_hostent(host);` at the bottom of the function.  This missing means the tests are failing with a memory leak.

tjwalton:
@bradh352: thanks. Should be fixed now.

PR #306, opened 2020-03-03
cmake: lower required version to 2.8.12

sergepetrenko:
There is no need to require CMake version >= 3.1.0
CMake 2.8.12 shipped with CentOS 6 and CentOS 7 still works just fine, as can be checked here:
https://gitlab.com/tarantool/tarantool/-/jobs/410010014 (We use cmake to build ares as part of tarantool. Search for lines containing `bundled-ares-project`)


bradh352:
I'm pretty sure there are significant issues on other platforms if trying to use cmake 2.8.  So while it might be ok on linux, I'm pretty sure that's not the case on Windows and possibly MacOS.   I don't have any interest in trying to support a long EOL version of CMake, even if it is the default on CentOS7 ... EPEL has had cmake3 forever ...

Totktonada:
We didn't expirienced any build problems on Linux, Mac OS and FreeBSD. We don't support Windows, however.

It is okay for us to keep the lower cmake version requirement within our fork, so please consider this PR as 'we tried cmake 2.8 on various platforms and everything works'. If there are known problems on, say, Windows, I understood your point.

Our motivation is to keep cmake 2.8 is to simplify build for a user / a distro maintainer: cmake 2.8 is available in system repositories on all distros / OSes.

PR #322, opened 2020-04-08
CMake for ios armv7 support

halx99:
Build iOS library with architectures: arm64 arm64e armv7 armv7s


bradh352:
We can't just unconditionally override this.  Some people may be setting it externally and passing it via `-DCMAKE_OSX_ARCHITECTURES` to cmake.  You at least need to check to make sure this is empty before overriding it.  CMake's documentations says to do this: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos

halx99:
Agree with you, I will close this PR. thanks

PR #323, opened 2020-04-09
Populate the INCLUDE_DIRECTORIES property of installed targets

dmitigr:
This is important to let CMake properly set include path after find_package(c-ares) when linking with c-ares by using target_link_libraries().


bradh352:
According to this:
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages
You should not use absolute paths, instead use the relative path.

dmitigr:
@bradh352 Agreed. Done.

PR #325, opened 2020-04-12
fix clang compiler warning

gengjiawen:
Found when building Node.js using clang
https://github.com/gengjiawen/node-github-workflow/runs/576197901?check_suite_focus=true#step:6:3783

```console
[2695/2882] CC obj/deps/cares/src/cares.ares__sortaddrinfo.o
../../deps/cares/src/ares__sortaddrinfo.c:494:2: warning: no newline at end of file [-Wnewline-eof]
}
 ^
1 warning generated.
```


bradh352:
It looks like your patch is no longer relevant with the other warnings that were fixed in b523541

Issue #327, opened 2020-04-29
Failure building c-ares on Windows for v1.16.0


codebytere:
Electron experienced a build failure in `c-ares` for `v1.16.0` as a Node.js dependency when building on Windows:

<details>
<summary> Stacktrace </summary>

```
../../third_party/electron_node/deps/cares/src/ares_getaddrinfo.c(459,48): error: incompatible pointer types passing 'char [51]' to parameter of type 'LPCWSTR' (aka 'const unsigned short *') [-Werror,-Wincompatible-pointer-types]
          if (RegOpenKeyEx(HKEY_LOCAL_MACHINE, WIN_NS_NT_KEY, 0, KEY_READ,
                                               ^~~~~~~~~~~~~
../../third_party/electron_node/deps/cares/src/ares_private.h(56,30): note: expanded from macro 'WIN_NS_NT_KEY'
#define WIN_NS_NT_KEY        "System\\CurrentControlSet\\Services\\Tcpip\\Parameters"
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
..\..\..\..\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um\winreg.h(705,22): note: passing argument to parameter 'lpSubKey' here
    _In_opt_ LPCWSTR lpSubKey,
                     ^
../../third_party/electron_node/deps/cares/src/ares_getaddrinfo.c(463,42): error: incompatible pointer types passing 'char [13]' to parameter of type 'LPCWSTR' (aka 'const unsigned short *') [-Werror,-Wincompatible-pointer-types]
              RegQueryValueEx(hkeyHosts, DATABASEPATH, NULL, NULL, (LPBYTE)tmp,
                                         ^~~~~~~~~~~~
../../third_party/electron_node/deps/cares/src/ares_private.h(61,30): note: expanded from macro 'DATABASEPATH'
#define DATABASEPATH         "DatabasePath"
                             ^~~~~~~~~~~~~~
..\..\..\..\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um\winreg.h(862,22): note: passing argument to parameter 'lpValueName' here
    _In_opt_ LPCWSTR lpValueName,
                     ^
[...]

PR #328, opened 2020-04-29
fix: windows LPCWSTR/LPWSTR cast incompatibilities

codebytere:
Closes https://github.com/c-ares/c-ares/issues/327.

Fixes the following compatibility and casting issues:
* Use [`RegQueryValueExA`](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa) instead of `RegQueryValueEx`
* Use [`ExpandEnvironmentStringsA`](https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-expandenvironmentstringsa) instead of `ExpandEnvironmentStrings`
* Use [`RegOpenKeyExA`](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regopenkeyexa) instead of `RegOpenKeyExA`
* Use [`GetWindowsDirectoryA`](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getwindowsdirectorya) instead of `GetWindowsDirectoryA`

Previous build errors can be [seen here](https://ci.appveyor.com/project/electron-bot/electron-ia32-testing/builds/32537988).

### DIFF
--- a/ares_getaddrinfo.c
+++ b/ares_getaddrinfo.c
@@ -456,18 +456,18 @@ static int file_lookup(struct host_query *hquery)
           char tmp[MAX_PATH];
           HKEY hkeyHosts;
 
-          if (RegOpenKeyEx(HKEY_LOCAL_MACHINE, WIN_NS_NT_KEY, 0, KEY_READ,
+          if (RegOpenKeyExA(HKEY_LOCAL_MACHINE, WIN_NS_NT_KEY, 0, KEY_READ,
                            &hkeyHosts) == ERROR_SUCCESS)
             {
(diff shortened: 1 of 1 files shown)


bradh352:
Your fix will only work if built with UNICODE.  Our build scripts do not use unicode which is why it works.  The proper thing to do here is like the rest of c-ares, use the 'A' variants of the respective functions, e.g. RegOpenKeyExA()

codebytere:
@bradh352 done - thanks!

Issue #82, opened 2017-01-10
Building with 'CURLDEBUG' fails


gvanem:
Currently building `cares` with `-DCURLDEBUG` fails:
```
ares__close_sockets.c
f:\MingW32\src\inet\curl\lib\../lib/memdebug.h(38): error C2054: expected '(' to follow 'CURL_EXTERN'
f:\MingW32\src\inet\curl\lib\../lib/memdebug.h(38): error C2085: 'curl_domalloc': not in formal parameter list
...
```
Seems to be a simple fix though:
```
--- a/ares_private.h 2016-04-30 10:37:07
+++ b/ares_private.h 2017-01-10 18:12:19
@@ -356,6 +356,7 @@
    libcurl lowlevel code from within library is ugly and only works when
    c-ares is built and linked with a similarly curldebug-enabled libcurl,
    but we do this anyway for convenience. */
+#include <curl/curl.h>
 #define HEADER_CURL_SETUP_ONCE_H
 #include "../lib/memdebug.h"
 #endif
```

bradh352:
@bagder can you confirm this is the right thing to do?  I don't know anything about curl internals.

bagder:
I rather think it is time to drop all the `CURLDEBUG` support that remains in c-ares.

Once upon the time that system made sure that c-ares got some testing and verifying when used by curl, as part of curl testing, but with the introduction of the real and proper c-ares test suite I think that need is long gone and is now a "hack" that can be cleaned up.

bradh352:
fixed in abe1ff1

PR #329, opened 2020-05-03
Fix for CURLDEBUG code

gvanem:
Since libcurl's changed some names in `memdebug.[ch]`.

Ref. https://github.com/curl/curl/commit/76b63489495ee1d49182a20525d7e6629cbab493


bradh352:
See #82 ... think we should just be removing CURLDEBUG

bradh352:
no longer relevant, removed as per #82 in abe1ff1

PR #334, opened 2020-05-26
Updated help information for adig, acountry, and ahost.

anonymoushelpishere:
Information from the man pages was added to the -h option of each of the commands.


bradh352:
please move code out of headers

anonymoushelpishere:
I was able to get Travis to pass with code in header but not in the '.c' (see be57554..., 6e9bae7...) not sure if this has to do with the 'Quick hack...' where it looks like ares_get_opt is manually included in some of the makefiles (e.g. Makefile.m32 at line 42) or if this is some other subtlety that I am not familiar with. In the case that I need to manually include in makefiles via the hack, this seems like more editing than should be needed for a simple help printout, thus perhaps I missed something. Currently all checks have passed.

bradh352:
your header file still has the full implementation.

you should just move the respective print functions to adig.c/ahost.c/etc instead of creating new external files.

bradh352:
should probably merge -? and -h to do the same output

anonymoushelpishere:
Note that, to stay consistent with 'man -h' (which I hope is a good example to follow) I omitted the -? from the usage note but the programs now respond to -? and -h.

adig.c
```diff
@@ -214,7 +215,12 @@ int main(int argc, char **argv)
           dbug_init();
 #endif
           break;
-
+        case 'h':
+          print_help_info_adig();
+          break;
+        case '?':
+          print_help_info_adig();
+          break;
```
gvanem, line 223: One could collapse these info:
```c
        case 'h':
        case '?':
          print_help_info_adig();
          break;
```

adig.c
```diff
@@ -825,3 +831,26 @@ static void append_addr_list(struct ares_addr_node **head,
   else
     *head = node;
 }
+
+
+/* Information from the man page. Formatting taken from man -h */
+static void print_help_info_adig() {
+    printf("adig, version %s \n\n", ARES_VERSION_STR);
+    printf("usage: adig [-h] [-d] [-f flag] [-s server] [-c class] [-t type] [-T|U port] name ...\n\n"
+    "  d : Print some extra debugging output.\n"
```
gvanem, line 840: Why ` d ` and not ` -d ...` as is more common in help-screens?
(same in all 3 `a*.c` files).

PR #339, opened 2020-07-18
Fix build with GCC 4.0.1

noloader:
The old Apple PowerMac G5 has GCC 4.0.1. The compiler is c89 by default.


bradh352:
I saw the mailing list post, but didn't realize there was a PR too before committing.  Closing this since its already fixed.

PR #346, opened 2020-08-24
support for system-assigned ports during test

apenn-msft:
Allow ares-test to use a system-assigned port (by passing port value 0) which will use the next available port assigned by the system.
Context: In a unit test sandbox environment, it is difficult to control what ports will be available at the time of running the test, and a default or manually-selected port can fail to bind if it is in use. The binding to a system-assigned port guarantees the port is available.

### DIFF
--- a/test/ares-test.cc
+++ b/test/ares-test.cc
@@ -171,7 +171,7 @@ void DefaultChannelModeTest::Process() {
 }
 
 MockServer::MockServer(int family, int port)
-  : udpport_(port), tcpport_(udpport_), qid_(-1) {
+  : udpport_(port), tcpport_(port), qid_(-1) {
   // Create a TCP socket to receive data on.
   tcpfd_ = socket(family, SOCK_STREAM, 0);
   EXPECT_NE(-1, tcpfd_);
--- a/test/ares-test.cc
+++ b/test/ares-test.cc
@@ -200,7 +200,7 @@ MockServer::MockServer(int family, int port, int tcpport)
     // retrieve system-assigned port
     if (udpport_ == 0) {
         int len = sizeof(addr);
-        int result = getsockname(udpfd_, (struct sockaddr*) & addr, &len);
+        int result = getsockname(udpfd_, (struct sockaddr*)&addr, &len);
         EXPECT_NE(SOCKET_ERROR, result);
         udpport_ = ntohs(addr.sin_port);
         EXPECT_NE(0, udpport_);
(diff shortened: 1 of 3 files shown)


apenn-msft:
> [Coverage Status](https://coveralls.io/builds/32966274)
> 
> Coverage decreased (-61.2%) to 27.93% when pulling **[9d6cd7b](https://github.com/c-ares/c-ares/commit/9d6cd7bd9770102650d384438bacef1f35465737) on apenn-msft:system-ports** into **[2c32de2](https://github.com/c-ares/c-ares/commit/2c32de2aa313494e065538cf05ab5f66208ecd91) on c-ares:master**.

this could be covered by setting up a new test command line that passes '-p 0', however since it's test code and this is up to the test runner (which port to use), I'd prefer to leave it to user choice.

bradh352:
Looks like the travis builds are all failing.  A quick glance says its due to int vs socklen_t.  You should probably use  ares_socklen_t  instead.

Also, instead of EXPECT_NE(SOCKET_ERROR, result);  use  EXPECT_EQ(0, result);  ... its more portable.

apenn-msft:
> Looks like the travis builds are all failing. A quick glance says its due to int vs socklen_t. You should probably use ares_socklen_t instead.
> 
> Also, instead of EXPECT_NE(SOCKET_ERROR, result); use EXPECT_EQ(0, result); ... its more portable.

on it, thanks!

comparing SOCKET_ERROR based on the only existing example I could find of getsockname in ares_sortaddinfo.c
if (getsockname(sock, src_addr, &len) == -1)

bradh352:
Just trying to evaluate your changeset a little further.  I'm not actually seeing where the port could be specified as zero to trigger your new logic except if you pass   -p 0  to the test harness maybe.  Is that how you intend on this to be tested?

What about when MockChannelOptsTest() passes a value > 1 for count.  I see a few tests that pass a value of '3' for count, so that would pass ports   0, 1, 2  in those tests if you used -p 0.

apenn-msft:
> Just trying to evaluate your changeset a little further. I'm not actually seeing where the port could be specified as zero to trigger your new logic except if you pass -p 0 to the test harness maybe. Is that how you intend on this to be tested?

Yep, that's the intention.

> What about when MockChannelOptsTest() passes a value > 1 for count. I see a few tests that pass a value of '3' for count, so that would pass ports 0, 1, 2 in those tests if you used -p 0.

good point. let me look into that

bradh352:
It might be better to go ahead and see what happens if you default to 0 too.  There's probably no reason to use a specific port with your change (but allow the -p to override it). And just handle that loop case differently if the mock_port is 0, to always pass 0 instead of incrementing so it gets a system-assigned port.

apenn-msft:
I've fixed the multi-server base port scenario; I'll try changing the default to 0 now. I'm actually curious if the 0-mean-system-assigned port is a universally agreed-upon (i.e. POSIX?) concept or if there is a define I should be using for it.

test/ares-test.cc
```diff
@@ -169,8 +170,8 @@ void DefaultChannelModeTest::Process() {
   ProcessWork(channel_, NoExtraFDs, nullptr);
 }
 
-MockServer::MockServer(int family, int port, int tcpport)
-  : udpport_(port), tcpport_(tcpport ? tcpport : udpport_), qid_(-1) {
```
apenn-msft, line 173: should be port, not udpport

bradh352:
Yes, 0 is portable for a system-assigned port to my knowledge.

PR #349, opened 2020-09-08
Reorg source tree

bradh352:
@bagder started a project a couple of years ago to reorganize the source tree.  I've now attempted to do this myself.  There should be quite a few similarities with #123


bagder (approved):
good stuff! :+1:

PR #356, opened 2020-09-12
[adig] add '-x' option.

gvanem:
Added a `dig-style* `-x` option. Also support `-xx` for a IPv6 bit-string `PTR` query.

From `man dig`:

>  Reverse lookups -- mapping addresses to names -- are simplified by the
>  -x option.  addr is an IPv4 address in dotted-decimal notation, or a
>  colon-delimited IPv6 address. When this option is used, there is no
>  need to provide the name, class and type arguments.  dig automatically
>  performs a lookup for a name like 11.12.13.10.in-addr.arpa and sets the
>  query type and class to PTR and IN respectively. By default, IPv6
>  addresses are looked up using nibble format under the IP6.ARPA domain.

Thus for an IPv4-address:
  * `c:\> adig -x -t ptr 1.1.1.1`
   question:  `1.1.1.1.in-addr.arpa. PTR`
   answer:    `one.one.one.one.`

And for an IPv6-address:
 *  `c:\> adig -x -t PTR 2001:67c:21e0::16`
  question:  `6.1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.E.1.2.C.7.6.0.1.0.0.2.IP6.ARPA.  PTR`
  answer:    `www.vg.no`

 *  `c:\> adig -xx -t PTR 2001:67c:21e0::16`
  question: `[x610000000000000000000E12C7601002].IP6.ARPA. PTR`
  answer:   *none*

The last example (with the more compact [RFC-2673](https://tools.ietf.org/html/rfc2673) notation; a bit-string),
returns no answer with the DNS-servers I've tested against. Seems it's not widely supported.

### DIFF
--- a/src/tools/adig.c
+++ b/src/tools/adig.c
@@ -171,6 +171,7 @@ static const unsigned char *display_question(const unsigned char *aptr,
                                              int alen);
 static const unsigned char *display_rr(const unsigned char *aptr,
                                        const unsigned char *abuf, int alen);
+static int convert_query (char **name, int use_bitstring);
 static const char *type_name(int type);
 static const char *class_name(int dnsclass);
 static void usage(void);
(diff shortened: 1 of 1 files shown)


bradh352:
Can you also update docs/adig.1 ?

gvanem:
> Can you also update docs/adig.1 ?

I suck at Groff syntax, but I'll try.

PR #359, opened 2020-09-12
Update acountry.c

gvanem:
* Use C99 var-arg macro for `TRACE()` macro. Are there still C89 compilers around?
* Fix the `usage:` output.


bradh352:
I personally don't think it worth breaking c89 compliance for something this minor.  Outside of gcc and clang, there are a lot of compilers that don't support c99 ... SCO OpenServer 5/6 comes to mind if I recall.  Yes, they're legacy systems, but since this project seems mostly to be in maintenance mode and not so much active development with new features we should probably keep it as c89

gvanem:
I'm okay with C89. No big deal. Closing.

PR #363, opened 2020-09-26
fuzz CAA parsing

dimbleby:


dimbleby:
'*' is not expanded inside double quotes so the fuzzing was broken anyway

PR #376, opened 2020-11-18
put back ares_dns.h as a public header

ffontaine:
`ares_dns.h` is not part of the public API since commit 0bf721cdd7f64b893c4ad015c341de680617411b resulting in the following build failure with resiprocate:

```
In file included from dns/DnsCnameRecord.cxx:7:
dns/AresCompat.hxx:5:10: fatal error: ares_dns.h: No such file or directory
 #include "ares_dns.h"
          ^~~~~~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/cbf158f0c037d44ef293a8804d18c84e3b731059


bagder (commented):
Install? This is not a public header, it should not be "installed" ...

ffontaine:
Prior to commit 0bf721cdd7f64b893c4ad015c341de680617411b, `ares_dns.h` was packaged as a public header through:

```
SET (CARES_HEADERS ares.h ares_version.h ares_dns.h "${PROJECT_BINARY_DIR}/ares_build.h" ares_rules.h)
```

and

```
# what headers to install on 'make install':
libcares_la_HEADERS = ares.h ares_version.h ares_dns.h \
	ares_build.h ares_rules.h
```

As a result, users of c-ares such as resiprocate and libeXosip used it (e.g. resiprocate uses `DNS__16BIT`, `DNS_HEADER_ANCOUNT`, `DNS_RR_LEN` ...). Should I understand from your answer that you intentionally decided to remove `ares_dns.h` from your public headers?

bagder:
Hm, I think I was wrong. It was indeed a public header so we should bring it back as such.

I think however that we should do so by also moving it to `include/` to make it obvious in the source tree.

bagder:
whatdoyousay @bradh352 ?

bradh352:
@bagder I don't think it was ever truly intended to be public, but I did look at the source and they're using those macros so we shouldn't break external integrations ... so moving to /include would be the right thing to do.

PR #385, opened 2020-11-25
Compile fix taken from https://github.com/nodejs/node/pull/36207

lxdicted:
Compile fix taken from https://github.com/nodejs/node/pull/36207

T_OPT is not defined on all platforms, see ares_create_query.c:35

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -60,6 +60,9 @@
 #include "ares_nowarn.h"
 #include "ares_private.h"
 
+#ifndef T_OPT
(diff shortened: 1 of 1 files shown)


bradh352:
Can you share the actual build failure when using c-ares 1.17.1 ?

T_OPT is defined if not found on the system in nameser.h which is included in ares_process.c .... so I'm a little surprised this issue is being reported as this appears to be a fix introduced between 1.16.1 and 1.17.0 with commit 535763a

lxdicted:
Thanks for pointing that out, the update of namerser.h was missing.

PR #389, opened 2020-12-08
Cares uri support

Prasek-prtes:
Added dns uri reply support. It's based on code of server reply.


src/lib/arec_parse_uri_reply.c
```diff
@@ -0,0 +1,184 @@
+
```
gvanem, line 1: The file-name is wrong; `ares_parse_uri_reply.c`.

src/tools/adig.c
```diff
@@ -152,6 +155,7 @@ static const struct nv types[] = {
   { "DNSKEY",   T_DNSKEY },
   { "CAA",      T_CAA },
   { "ANY",      T_ANY }
+  { "URI",      T_URI }
```
gvanem, line 158: Add a `,` above that line. You probably have noted the AppVeyor jobs have failed.

bradh352:
Seems like a number of build failures.  We'll also need a test case.

bradh352:
moved to #411

PR #390, opened 2020-12-13
zos: successfully build c-ares codebase on z/OS systems

dhrumilrana:
This PR is for building c-ares on z/OS systems. The PR makes changes couple of header files to introduce MVS checks for z/OS specific code.


bradh352:
looks pretty good.  Just need a couple of fixups.

1. Remove the assert
2. Conform to c89 (basically put any variable declarations at the top of a code block, not sprinkled throughout)
3. CARES_USE_LIBRESOLV should be fixed to be defined via the autotools and cmake build systems rather than doing this in the code directly ... if you search for other usages of CARES_USE_LIBRESOLV if its set it disables other sections of code you're leaving enabled (unless this is really intentional).

Then a question ... why the default_malloc override?  Are there actually instances where you've seen c-ares call malloc(0) and it matter?

dhrumilrana:
> looks pretty good. Just need a couple of fixups.
> 
> 1. Remove the assert
> 2. Conform to c89 (basically put any variable declarations at the top of a code block, not sprinkled throughout)
> 3. CARES_USE_LIBRESOLV should be fixed to be defined via the autotools and cmake build systems rather than doing this in the code directly ... if you search for other usages of CARES_USE_LIBRESOLV if its set it disables other sections of code you're leaving enabled (unless this is really intentional).
> 
> Then a question ... why the default_malloc override? Are there actually instances where you've seen c-ares call malloc(0) and it matter?

Yes, on z/OS malloc(0) returns 0, while on others it returns a valid pointer. As it is used in the linked list overriding it was safer as the other platforms would not be affected and this also allows future code to be added without worrying about z/OS functionality.

Also for CARES_USE_LIBRESOV, I have added a check for ZOS in CMakeLists.txt as per your request, but I'm confused as to how to add a check for z/OS in configure.ac where CARES_USE_LIBRESOV is being set for apple.

bagder:
`malloc(0)` is documented to return *either* NULL or 0 so no code in c-ares should depend on one or the other. in fact, malloc(0) should rather be completely avoided, IMHO.

src/lib/ares_init.c
```diff
@@ -50,7 +50,11 @@
 #define MAX_DNS_PROPERTIES    8
 #endif
 
-#if defined(CARES_USE_LIBRESOLV)
+#if defined(__MVS__)
+#include <assert.h>
+#endif
+
+#if defined(CARES_USE_LIBRESOLV) || defined(__MVS__)
```
bradh352, line 57: remove  "|| defined (__MVS__)"

src/lib/ares_init.c
```diff
@@ -50,7 +50,11 @@
 #define MAX_DNS_PROPERTIES    8
 #endif
 
-#if defined(CARES_USE_LIBRESOLV)
+#if defined(__MVS__)
+#include <assert.h>
```
bradh352, line 54: can this be removed?

src/lib/ares_init.c
```diff
@@ -1472,7 +1476,57 @@ static int init_by_resolv_conf(ares_channel channel)
...
+      rc = res_init();
+    }
+    if (rc == -1) {
+      return ARES_ENOMEM;
+    }
+    res = __res();
+  }
+
+  int count4, count6;
+  __STATEEXTIPV6 *v6 = res->__res_extIPv6;
```
bradh352, line 1494: These 3 variable declarations should be moved up to the top of the block for c89 compliance.

src/lib/ares_init.c
```diff
@@ -1472,7 +1476,57 @@ static int init_by_resolv_conf(ares_channel channel)
...
+  }
 
+  nservers = count4 + count6;
+  servers = ares_malloc(nservers * sizeof(struct server_state));
+  if (!servers)
+    return ARES_ENOMEM;
+
+  memset(servers, 0, nservers * sizeof(struct server_state));
+
+  struct server_state *pserver = servers;
```
bradh352, line 1509: This declaration should be moved to the top of the block for c89 compliance.

src/lib/ares_init.c
```diff
@@ -1472,7 +1476,57 @@ static int init_by_resolv_conf(ares_channel channel)
...
+    struct sockaddr_in *addr_in = &(res->nsaddr_list[i]);
+    pserver->addr.addrV4.s_addr = addr_in->sin_addr.s_addr;
+    pserver->addr.family = AF_INET;
+    pserver->addr.udp_port = addr_in->sin_port;
+    pserver->addr.tcp_port = addr_in->sin_port;
+  }
+
+  for (int j = 0; j < count6; ++j, ++pserver) {
+    struct sockaddr_in6 *addr_in = &(v6->__stat_nsaddr_list[j]);
+    assert(sizeof(pserver->addr.addr.addr6) == sizeof(addr_in->sin6_addr));
```
bradh352, line 1520: is this assert really necessary?  Are there some systems where this isn't true?

src/lib/ares_library_init.c
```diff
@@ -45,6 +45,14 @@ static int          ares_init_flags;
 static void *default_malloc(size_t size) { return malloc(size); }
 static void *default_realloc(void *p, size_t size) { return realloc(p, size); }
 static void default_free(void *p) { free(p); }
+#elif defined(__MVS__)
```
bradh352, line 48: This block should be removed.  If there are really any malloc(0) issues in c-ares, those need to be fixed instead.

src/lib/ares_setup.h
```diff
@@ -179,7 +179,8 @@
  * Android does have the arpa/nameser.h header which is detected by configure
  * but it appears to be empty with recent NDK r7b / r7c, so we undefine here.
  */
-#if (defined(ANDROID) || defined(__ANDROID__)) && defined(HAVE_ARPA_NAMESER_H)
+#if (defined(ANDROID) || defined(__ANDROID__) || defined(__MVS__)) && \
```
bradh352, line 182: Can we add a comment as to why this may be necessary on z/OS?

bradh352 (changes requested):
more detailed review

bradh352:
I haven't seen the CMakelists.txt modification, once that is added to this PR, I can then tell you what needs to happen in autotools to do the equivalent check...

src/lib/ares_init.c
```diff
@@ -1472,7 +1472,57 @@ static int init_by_resolv_conf(ares_channel channel)
     /* Catch the case when all the above checks fail (which happens when there
        is no network card or the cable is unplugged) */
     status = ARES_EFILE;
+#elif defined(__MVS__)
+
+  static struct __res_state *res = 0;
+  int count4, count6;
+  __STATEEXTIPV6 *v6 = res->__res_extIPv6;
```
bradh352, line 1479: pretty sure you need to set   v6=res->__res_extIPv6;  after setting res as otherwise you're dereferencing a NULL ptr here.

src/lib/ares_init.c
```diff
@@ -1472,7 +1472,57 @@ static int init_by_resolv_conf(ares_channel channel)
     /* Catch the case when all the above checks fail (which happens when there
        is no network card or the cable is unplugged) */
     status = ARES_EFILE;
+#elif defined(__MVS__)
+
+  static struct __res_state *res = 0;
```
bradh352, line 1477: Should this really be static?  What if system configs change?

bradh352 (changes requested):

bradh352:
Regarding autotools support for z/OS, I'd think in configure.ac after this:

```
dnl resolv lib for Apple (MacOS and iOS)
AS_IF([test "x$host_vendor" = "xapple"], [
  AC_SEARCH_LIBS([res_servicename], [resolv], [
    AC_DEFINE([CARES_USE_LIBRESOLV], [1], [Use resolver library to configure cares])
  ], [
    AC_MSG_ERROR([Unable to find libresolv which is required for iPhone targets])
  ])
])
```

you could use

```
dnl resolv lib for z/OS
AS_IF([test "x$host_vendor" = "xibm" -a "x$host_os" = "xopenedition" ], [
  AC_SEARCH_LIBS([res_init], [resolv], [
    AC_DEFINE([CARES_USE_LIBRESOLV], [1], [Use resolver library to configure cares])
  ], [
    AC_MSG_ERROR([Unable to find libresolv which is required for z/OS])
  ])
])
```

src/lib/ares_init.c
dhrumilrana: Yes it only needs to be done once.

bradh352: So the system config can't change?  Some apps detect system changes and will re-run ares_init() to pick up the changes and start using the new channel.  It would seem to me that declaring this static would mean that any apps attempting to re-run ares_init() after a system change wouldn't see it... Perhaps I'm missing something here.

dhrumilrana:
Added the autotools update. Regarding malloc(0) calls I see that malloc(0) is being called when in two of the node.js FVT test cases, I have linked the two tests: [test 1](https://github.com/nodejs/node/blob/master/test/parallel/test-dns.js) and [test 2](and https://github.com/nodejs/node/blob/master/test/parallel/test-dns-resolveany.js). Let me know if I need to add in the previous block of code.

bradh352 (approved):

bradh352:
> Added the autotools update. Regarding malloc(0) calls I see that malloc(0) is being called when in two of the node.js FVT test cases, I have linked the two tests: [test 1](https://github.com/nodejs/node/blob/master/test/parallel/test-dns.js) and [test 2](https://github.com/nodejs/node/blob/master/test/parallel/test-dns-resolveany.js). Let me know if I need to add in the previous block of code.

Can you be more specified in regards to the test failures?  Is there a backtrace?  Can you show what line in the tests they failed on?  We'd need to know the code path if it is crashing due to this to know if its something in nodejs's assumptions or something deep inside c-ares.

bradh352:
Also, does the c-ares test suite work for you on z/OS the way it is right now (without the malloc workaround)?

I just performed a local hack to have ares_malloc(0) return NULL always and print out a message if this condition is hit.  At least none of the conditions in the test suite appear to have hit this at all.  That doesn't mean its not possible in some code path not in the test suite, of course.

dhrumilrana:
Yes, I ran the c-ares test-suite with and without the malloc workaround and there are no hits in either cases. I can provide you with a more detailed analysis of the failure soon.

bradh352:
I'm going to go ahead and merge this and treat the malloc(0) as a different issue.  Infact, I'll probably follow-up with a commit to make the default malloc implementation return NULL on 0 always to catch any possible bugs.  Maybe we can get a report of an issue on another system with NodeJS.

Issue #387, opened 2020-12-06
[Feature request] Expose TTL values for all rrtypes.


lxdicted:
c-ares does not expose TTL values except for A and AAAA rrtypes. nodejs has a desire for this, see https://github.com/nodejs/node/issues/14713.

bradh352:
This would break ABI to expose since the data structure are not opaque.  At some point we should figure this out, but it isn't going to be easy unfortunately.

bagder:
I think the cleanest way to provide TTL to users is to introduce new RR parsing functions that return other (extended) structs that contain TTL and possibly even a way to extend the struct in a future.

kylebevans:
@markdingo offered a patch for this back in 2013 here: https://c-ares.haxx.se/mail/c-ares-archive-2013-07/0005.shtml

Looks like it didn't get traction at the time, and looks like that patch changed the structs directly which as you mentioned would break ABI.

I'm not an expert, is something like this what you mean for a new struct that could be extended?

```
struct ares_srv_reply_with_ttl {
  struct ares_srv_reply  *next;
  char                   *host;
  unsigned short          priority;
  unsigned short          weight;
  unsigned short          port;
  int                     ttl;
};

typedef struct ares_srv_reply_with_ttl *ares_srv_reply_with_ttl_ptr;

int
ares_parse_srv_reply_with_ttl (const unsigned char *abuf, int alen,
                      struct ares_srv_reply_with_ttl_ptr srv_out)
{
/* same code as ares_parse_srv_reply but also add ttl */
}
```

bradh352:
Better yet would be to add a new function that takes an opaque pointer/object, and accessors to the pointer/object to future-proof it in case we want to add anything else in the future.  Having a structure as public is always a bad thing to do :/

kylebevans:
Bear with me, that's what I had hoped to achieve with the typedef of `*ares_srv_reply_with_ttl_ptr;` and then the new function that takes it as a parameter: `ares_parse_srv_reply_with_ttl`. Is that the wrong way to do it?  I guess I need to have the actual struct definition in an internal header file and an incomplete definition in the normal header file.  Is there anything else to make it private?

I guess I didn't put accessors in there, so they would be like this, right?

```
char* ares_srv_reply_with_ttl_get_host (ares_srv_reply_with_ttl_ptr srv_reply)
char* ares_srv_reply_with_ttl_set_host (ares_srv_reply_with_ttl_ptr srv_reply)
```

[...]

PR #393, opened 2021-01-25
add a function to parse a srv reply with ttl

kylebevans:
This PR is in response to issue #387.
This adds a new private srv record struct with a ttl field.
It also adds public accessor functions to set all of the
fields.  Finally, it replicates the ares_parse_srv_reply
function to use the new struct and accessors.

So far, I just added one test as a POC.  If this is the right direction, I can add more tests as well as implement the other RR parsing functions.


kylebevans:
I made this to confirm whether this was how you wanted to approach #387.  If it is, then I will implement the parsing for the other RR records on this PR too, so no need to merge it yet.  I also need to add more tests.  Thanks.

bradh352:
Looks like its going in the right direction.

I would think though that we should make ares_parse_srv_reply() call ares_parse_srv_reply_ext() so we don't have 2 basically identical code paths.  Basically make the old one a legacy wrapper.

Infact, I'd almost opt for naming   ares_parse_srv_reply_ext() -> cares_parse_srv_reply() ... and start making a new parallel API that is more modern and extensible prefixed with cares_ ... afterall, this is a fork of the original ares project.

kylebevans:
That makes sense to me and I considered that, but after thinking it through it seems like the new function will still have to duplicate much of the logic anyway:

* new function calls ares_parse_srv_reply
* ares_parse_srv_reply calls ares_expand_name to get past the name part of the question part and then skip past the question
* then it loops through the RRs
* for each RR, it calls ares_expand_name again to get past the name, which is discarded
* then it starts parsing the srv record, getting the weight, priority, port
* finally it calls ares_expand_name again for every RR to get the srv record name
* it fills the ares_srv_reply struct and adds it to the linked list
* at the end looping through the RRs the linked list is finished and ares_parse_srv_reply cleans up and returns

Now we need to get the TTLs out of all the RRs again so we'll have to do this:
* call ares_expand_name again to skip past the question
* loop through the RRs again and simultaneously loop through the linked list 
* call ares_expand_name again for every RR to get past the name
* parse the srv record fields to pull out the TTL
* then make our new struct with the info from the old struct and the TTL we found and add it to a new linked list
* cleanup and return

So after thinking through that, I kind of thought it might be preferable to just replace ares_parse_srv_reply completely and deprecate it.  It would save on the complexity and overall work the function does, but I maybe I'm making too much out of it.  I will do it whichever way you think is best.  Thanks for helping me with this.  Either way, I do think a new parallel API is a good idea and "cares_" is a good naming convention.

Also, for the life of me I can't figure out why the NMAKE build is failing.  Do you have any idea there?

bradh352:
anytime you're dealing with message parsing itself, its always better not to do it too often ... most bugs tend to be in that sort of code.  So making the legacy function a wrapper is preferred, even if its still tedious and a good amount of code.  It also makes the legacy function easier to maintain in the long run.

Regarding the windows build failure, my guess is you have some additional whitespace in Makefile.inc that is hard to see.  Nmake is very white-space sensitive.  It sort of looks like you used spaces between the ares_reply_ext.c and the \ terminating the line, but everything else is using tabs.

src/lib/cares_reply.c
```diff
@@ -0,0 +1,89 @@
...
+}
+
+char* cares_srv_reply_get_host(const struct cares_srv_reply* srv_reply)
+{
+  char *newhost = NULL;
+  if ((newhost = malloc(strlen(srv_reply->host) + 1)) != NULL) {
+    strcpy(newhost, srv_reply->host);
+  }
+  return newhost;
+}
```
kylebevans, line 32: I changed this to allocate a new `char*` and return it.  Just returning the raw pointer had the side effect of allowing direct modification of the host field of the struct using the pointer.  The downside to this method is that the caller will now have to both check for NULL and free the `char*`.  Let me know if there is a better way.

Also, what checks need to be done on srv_reply being passed here? Just check for NULL?  If it's not NULL then is it safe to assume I can `strcpy` without an overflow?

kylebevans:
Thanks for your help; I made ares_parse_srv_reply a wrapper for cares_parse_srv_reply.  Is this on the right track for what you envisioned? I want to make sure before I start working on other parse functions.  Thanks for all your help.

src/lib/cares_reply.c
bradh352: I don't see a reason to do that, instead just make the return value   const char * and return the pointer ... that designates the caller isn't allowed to modify the return value.

bagder:
Note that two sanitizer builds fail tests now.

src/lib/cares_reply.c
kylebevans: That makes sense, but in our case back in ares_parse_srv_reply, we will have to do this copy there to get host out of the cares_srv_reply struct and into the ares_srv_reply struct so that we can free the cares_srv_reply.  Is that okay then?

src/lib/ares_parse_srv_reply.c
```diff
@@ -38,128 +38,80 @@ int
...
-      if (srv_head)
-        ares_free_data (srv_head);
-      return status;
+      status = ARES_ENOMEM;
+      break;
     }
+    if ((newhost = malloc(strlen(tmphost) + 1)) != NULL) {
+      strcpy(newhost, tmphost);
+    }
+    srv_curr->host = newhost;
```
kylebevans, line 97: Here is where the copy happens now.  Everything check out?

bagder: 1. Why not use strdup() ?
2. If the malloc fails, it must return error and fail the entire function call (without leaking anything)

src/lib/ares_parse_srv_reply.c
```diff
@@ -38,128 +39,76 @@ int
...
+      srv_head = srv_curr;
+    }
+    srv_last = srv_curr;
+
+    /* copy the host to newhost so we can free csrv_out */
+    char *newhost = ares_strdup(cares_srv_reply_get_host(csrv_curr));
+    if (!newhost) {
+      status = ARES_ENOMEM;
+      break;
     }
```
kylebevans, line 92: 1. Thanks for telling me about strdup, it is better and simplifies the logic. On the last commit, I was missing the "else" part to set the status to an error if the malloc failed, but it's simpler and clearer now.
2. There are 3 places where allocations occur: csrv_out at the top and then srv_curr and newhost inside the loop for each RR.  I think in each case I check for null, set an error, make sure all three of them are freed (newhost is freed as part of the new ares_srv_reply linked list), and return the error status. Does it look right to you?

kylebevans: also, do you want me to add each RR as a separate PR?  Or put them all in one?

src/lib/cares_reply.c
```diff
@@ -0,0 +1,85 @@
...
+ * without express or implied warranty.
+ */
+
+#include "ares_setup.h"
+#include "ares.h"
+#include "ares_private.h"
+#include "string.h"
+
+struct cares_srv_reply*
+cares_srv_reply_get_next(const struct cares_srv_reply* srv_reply)
```
kylebevans, line 20: Do you want tests on these getters and setters?

bradh352:
Sorry for the delay, been slammed lately.  I think the setters should be private, and only the getters be public.  The setters would be tested already in the internal implementation, and the getters would be tested by the legacy function wrappers when those are called for the most part.

And yes, I think everything in one large PR would be best as it may uncover things that hadn't been thought about previously.

src/lib/cares_parse_ptr_reply.c
```diff
@@ -0,0 +1,188 @@
...
+      ares_free(rr_name);
+      rr_name = NULL;
+
+      /* Move on to the next record */
+      aptr += rr_len;
+      if (aptr > abuf + alen)
+        {  /* LCOV_EXCL_START: already checked above */
+          status = ARES_EBADRESP;
+          break;
+        }  /* LCOV_EXCL_STOP */
```
kylebevans, line 168: Some parse functions have this check after moving aptr, and some don't.  Should all of them have it?

src/lib/cares_reply.c
```diff
@@ -0,0 +1,534 @@
...
+cares_srv_reply*
+cares_srv_reply_get_next(const cares_srv_reply* srv_reply)
+{
+  return srv_reply->next;
+}
+
+const char* cares_srv_reply_get_host(const cares_srv_reply* srv_reply)
+{
+  return srv_reply->host;
+}
```
kylebevans, line 29: Should we check for null on the input variables in these getters? Like on srv_reply here? It will be UB if a caller passes a null pointer, right?

kylebevans:
I think this is getting pretty close, let me know what you think.  I left some questions above, and I also want to point out that I made new types for cares_ns_reply and cares_ptr_reply. The parse functions return a linked list of those types rather than a hostent.  I hope that's okay; it felt more consistent to me.

src/lib/cares_parse_srv_reply.c
```diff
@@ -0,0 +1,177 @@
...
+          cares_srv_reply_set_port(srv_curr, DNS__16BIT(vptr));
+          vptr += sizeof(unsigned short);
+          cares_srv_reply_set_ttl(srv_curr, rr_ttl);
+
+          status = ares_expand_name (vptr, abuf, alen, &srv_host, &len);
+          if (status != ARES_SUCCESS)
+            break;
+          cares_srv_reply_set_host(srv_curr, srv_host);
+        }
+      else if (rr_type != T_CNAME)
```
dkikidis, line 147: Could you please explain why T_CNAME is excluded? A short comment in the code would be great.
(Same for CAA, MX, NAPTR, NS, PTR, SRV, TXT)

bradh352:
sorry, life's been crazy.  I'll really try to get to this review this week.

src/lib/ares_parse_caa_reply.c
```diff
@@ -36,163 +36,96 @@
...
+    if (ccaa_out)
+      ares_free_data(ccaa_out);
     return status;
+  }
+
+  /* iterate through the cares_caa_reply list and
+   * create a new ares_caa_reply */
+  for(ccaa_curr = ccaa_out; ccaa_curr;
+      ccaa_curr = cares_caa_reply_get_next(ccaa_curr))
+  {
```
kylebevans, line 68: For #4 on your list, I'm having trouble figuring out a way to do this loop with cares_caa_reply_get_next and others like it returning a const pointer.  The compiler throws a warning that the const value is getting ignored since I'm assigning it to a non-const variable ccaa_curr.  Is that warning okay?

If trying to make ccaa_curr const, then I would have to make it scoped to just the for loop, and make the loop an infinite loop that has an if statement inside that breaks if ccaa_curr is null.  However, then I have a problem with the very first iteration where I need to assign ccaa_curr from ccaa_out instead of cares_caa_reply_get_next.

kylebevans:
So far this addresses numbers 1 and 2 on your list.  I had a question about 4 which i put inline on the code above.  I'll also continue working on 3 and 5.  For 5, maybe I'll update one of them to that pattern and see how you like it.

src/lib/ares_parse_caa_reply.c
bradh352: ccaa_curr should definitely be defined as const.  Not sure what you mean by scoped to the loop, you can just define it const where it is.  What error are you getting when you do, perhaps there's something missing in the quick review i just did..

kylebevans: Okay, i tried exactly what you just said and I didn't get an error.  I think I was getting confused by constant pointer vs pointer to constant, and I was hung up thinking about how I'm mutating ccaa_curr, but really we're making the value pointed to const here, not the pointer itself. Right?

kylebevans:
Also, i made the ttl field of ares_uri_reply an unsigned int instead of an int, which isn't strictly related to this PR but I think it was a bug that could lead to a signed integer overflow.

src/lib/ares_parse_caa_reply.c
bradh352: const in this case means the data pointed to is read-only.  The pointer itself can be switched to another read-only address.

bradh352:
> Also, i made the ttl field of ares_uri_reply an unsigned int instead of an int, which isn't strictly related to this PR but I think it was a bug that could lead to a signed integer overflow.

Just make sure the code doesn't use a ttl of -1 to indicate no ttl has yet been set or something along those lines, otherwise I'd think it should be safe.

bradh352:
> So far this addresses numbers 1 and 2 on your list. I had a question about 4 which i put inline on the code above. I'll also continue working on 3 and 5. For 5, maybe I'll update one of them to that pattern and see how you like it.

Most likely for #5, just make sure you don't realloc each time you add a member.  Either pre-count or use some growable array logic (e.g. grows by power of 2 when it needs to expand).

kylebevans:
@bradh352 I've reworked the srv replies to use an array and an iterator.  Let me know if you like that pattern better or if you like the old linked list pattern.

bradh352:
Hi @kylebevans sorry for the late reply here, it may be a couple of weeks before I have time to review this.  Thanks for doing this though, I'm sure we'll be pulling this in as I think it greatly modernizes the c-ares API... it just might take a while to get through since the patch set is absolutely huge :)

bassmandja:
Putting a note on this to attempt to resuscitate the pending `srv_dns` extension cluster in Envoy

kylebevans:
@bradh352 are you still interested in this PR? I could try to get it to a finish if there is still a desire for it.  I think the current question is whether to rewrite all the replies with the array/iterator pattern like the srv reply is now, or to keep the current api pattern.

bradh352:
thanks for your work on this, but i ended up going an easier to maintain way in c-ares 1.21.0 with ares_dns_record.h.  Probably in c-ares 1.22 I'll make these interfaces public.  It would be nice to have feedback on this.

PR #396, opened 2021-02-17
CMake: fix compile error when RANDOM_FILE is defined

demian51:
When `RANDOM_FILE` is defined in CMake, compile error occurs:
> /home/demian51/cares/src/lib/ares_init.c: In function ‘randomize_key’:
/home/demian51/cares/src/lib/ares_init.c:2474:30: error: expected expression before ‘,’ token
2474 |   FILE *f = fopen(RANDOM_FILE, "rb");
      |                              ^

`RANDOM_FILE` is used as file path(e.g., "dev/urandom") but it defined as empty value in `ares_config.h.cmake`. As a result the code is preprocessed like `FILE *f = fopen(, "rb");` and compile error occurs.

So fixed the macro to be defined as a string.

### DIFF
--- a/src/lib/ares_config.h.cmake
+++ b/src/lib/ares_config.h.cmake
@@ -347,7 +347,7 @@
 #cmakedefine NEED_MEMORY_H
 
 /* a suitable file/device to read random data from */
-#cmakedefine RANDOM_FILE
+#cmakedefine RANDOM_FILE "@RANDOM_FILE@"
 
 /* Define to the type qualifier pointed by arg 5 for recvfrom. */
 #define RECVFROM_QUAL_ARG5 @RECVFROM_QUAL_ARG5@


demian51:
uh-oh, sorry I modified the author of commit.

bradh352:
Just taking a quick look at this I don't see how when using cmake `RANDOM_FILE` could ever be defined at all.  `ares_config.h` that gets generated should always end up doing   `/* #undef RANDOM_FILE */`  as `CMakeLists.txt` never attempts to determine what value it should actually be.

What system are you on that it did something other than this, and what version of cmake?

Granted, it appears to be bad that cmake doesn't attempt to set RANDOM_FILE as it falls back to a less secure RC4 key generation mechanism.

demian51:
I'm building c-ares in Ubuntu 20.04(WSL) with libcurl. It seems that libcurl sets `RANDOM_FILE`:

https://github.com/curl/curl/blob/f6d13090d4b108af75426f1ba225e31c9d4f61cf/CMakeLists.txt#L975

bagder:
`./configure --with-random=PATH` gets that define set in the build, if I read the code correctly.

bradh352:
Yeah, autotools does it properly, looks like this was missed in CMake.  Also, probably need to use a prefixed variable like CARES_RANDOM_FILE to prevent clashes like this anyhow.

bagder:
But this is a private define, right? How can it collide with anyone else's definition of it?

bradh352:
Sorry, should have been more specific, the variable in the ares_config.h.cmake that is used for the `#cmakedefine` is what needs to be renamed. The define that actually gets generated in ares_config.h isn't really relevant other than the fact that cmake doesn't make it easy to use a different name than an internal variable name. When you chain-build with CMake it inherits variables set by other CMakelist.txt's ... so in this case, it inherited the variable set by curl's CMakelist.txt which set RANDOM_FILE.  So this is purely a CMake inheritance thing during chain building, and not a C-code compilation thing.

bradh352:
moving to PR #397

Issue #398, opened 2021-03-01
[Bug] Using search domains if number of dots in the name <= ndots


vmishenev:
If dots number  of the domain name is greater than or equal to ```ndots```, no search domains should not using (even in failure case of a dns query for "as-is" name).
This is how the Linux resolver actually works.

bradh352:
Can you provide an example where ndots is not being honored?  Which function is being used, e.g. ares_getaddrinfo() vs ares_gethostbyname() and so on.

vmishenev:
Example: 
```ndots==1``` (default)
The search domain is set to ```example.com```.
A call ```ares_gethostbyname``` for  non-existent domain  ```ssssssssssss.com```:
1. DNS query for name ```ssssssssssss.ru```, server returned NXDOMAIN
2. DNS query for name ```ssssssssssss.com.example.com``` --- But this query should not do. It is excess.

bradh352:
https://linux.die.net/man/5/resolv.conf :
`ndots:n
sets a threshold for the number of dots which must appear in a name given to res_query(3) (see resolver(3)) before an initial absolute query will be made. The default for n is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute name before any search list elements are appended to it. The value for this option is silently capped to 15.`

So reading that, it ONLY controls how many dots are in the provided name before the absolute name is searched *first*.  It does not prevent using the search list at all.

Another reference that has a slightly different description that I think is more clear:
https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html
`sets a threshold for the number of dots which must appear in a name before an initial absolute query will be made. The default for n is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute name before any search list elements are appended to it.`

If your intent is to NOT use the search domains at all, then your query needs to contain a '.' as the last character.

PR #404, opened 2021-04-15
End query if dots in a domain >= ndots

vmishenev:
This PR is in response to issue #398.


bradh352:
Forgive me if I missed something as I took all of 2 minutes to look at the commit, but how is squery->status_as_is ever going to be equal to -1?  It looks like this gets set to the callback value from ares_query()'s status argument, which should be one of the ARES_E* values which are all positive.

Also based on what is trying to be resolved, I'd have expected an ndots check of some sort.

vmishenev:
Thanks for the answer. I hurried. This was not the best and most obvious idea. I've fixed, but i ain't sure about ares_search logic.

bradh352:
This seems to be in violation of the actual spec, so its not something we can accept.

PR #406, opened 2021-06-14
Validate hostnames in DNS responses and discard from malicious servers

bradh352:
DNS servers could intentionally return malformed DNS responses that could cause client-side issues (like XSS).  Attempt to identify such responses and throw them away and return EBADRESP.

It is not clear if there are any legitimate use cases where a DNS response might return a non-validly formatted hostname.

### DIFF
--- a/src/lib/ares_expand_name.c
+++ b/src/lib/ares_expand_name.c
@@ -30,7 +30,7 @@
 #define MAX_INDIRS 50
 
 static int name_length(const unsigned char *encoded, const unsigned char *abuf,
-                       int alen);
+                       int alen, int is_hostname);
 
 /* Reserved characters for names that need to be escaped */
 static int is_reservedch(int ch)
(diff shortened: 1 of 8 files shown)


bradh352:
@bagder any reservations about doing this?

src/lib/ares_expand_name.c
```diff
@@ -52,6 +52,31 @@ static int is_reservedch(int ch)
...
+  /* [A-Za-z0-9-.]
+   * Don't use isalnum() as it is locale-specific
+   */
+  if (ch >= 'A' && ch <= 'Z')
+    return 1;
+  if (ch >= 'a' && ch <= 'z')
+    return 1;
+  if (ch >= '0' && ch <= '9')
+    return 1;
+  if (ch == '-' || ch == '.')
```
AdamMajer, line 74: `.` is not a valid character for hostnames

bradh352: function name is misleading, its validating a full domain name response which can include .

PR #411, opened 2021-07-28
Added support for URI(Uniform Resource Identifier) records.

martin-256:
I'm finishing work from pull request https://github.com/c-ares/c-ares/pull/389 created by my former colleague https://github.com/Prasek-prtes.

Added tests and forgotten adjustments.

### DIFF
--- a/src/lib/ares_parse_uri_reply.c
+++ b/src/lib/ares_parse_uri_reply.c
@@ -149,6 +149,11 @@ ares_parse_uri_reply (const unsigned char *abuf, int alen,
           uri_curr->weight = DNS__16BIT(vptr);
           vptr += sizeof(unsigned short);
           uri_curr->uri = (char *)ares_malloc(rr_len-3);
+	  	  if (!uri_curr->uri)
+	      	{
+	      	  status = ARES_ENOMEM;
+			  break;
+		    }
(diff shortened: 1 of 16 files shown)


src/lib/ares_parse_uri_reply.c
```diff
@@ -0,0 +1,190 @@
...
+#endif
+#ifdef HAVE_ARPA_NAMESER_H
+#  include <arpa/nameser.h>
+#else
+#  include "ares_nameser.h"
+#endif
+#ifdef HAVE_ARPA_NAMESER_COMPAT_H
+#  include <arpa/nameser_compat.h>
+#endif
+
```
bradh352, line 37: Please just include ares_nameser.h always, and never directly include arpa/nameser.h or arpa/nameser_compat.h ... see c2f3235 on this change we made to the rest of c-ares

bradh352 (commented):
at a cursory glance, the only issue I see is the headers mentioned in the review

src/lib/ares_parse_uri_reply.c
martin-256: Thank you. I didn't notice this change.

bradh352:
I'll merge this after the 1.17.2 release which is just about out.

Issue #416, opened 2021-08-17
Build is broken on Debian 11


dpvpro:
The same error like in #409

Build output:

```
user@debian-build:/veil_rebuild/test$ apt source -b c-ares
Reading package lists... Done
NOTICE: 'c-ares' packaging is maintained in the 'Git' version control system at:
https://salsa.debian.org/debian/c-ares.git
Please use:
git clone https://salsa.debian.org/debian/c-ares.git
to retrieve the latest (possibly unreleased) updates to the package.
Need to get 1,531 kB of source archives.
Get:1 http://security.debian.org/debian-security bullseye-security/main c-ares 1.17.1-1+deb11u1 (dsc) [2,301 B]
Get:2 http://security.debian.org/debian-security bullseye-security/main c-ares 1.17.1-1+deb11u1 (tar) [1,519 kB]
Get:3 http://security.debian.org/debian-security bullseye-security/main c-ares 1.17.1-1+deb11u1 (asc) [488 B]
Get:4 http://security.debian.org/debian-security bullseye-security/main c-ares 1.17.1-1+deb11u1 (diff) [9,960 B]
Fetched 1,531 kB in 0s (3,756 kB/s)
dpkg-source: info: extracting c-ares in c-ares-1.17.1
dpkg-source: info: unpacking c-ares_1.17.1.orig.tar.gz
dpkg-source: info: unpacking c-ares_1.17.1-1+deb11u1.debian.tar.xz
dpkg-source: info: using patch list from debian/patches/series
dpkg-source: info: applying disable-cflags-rewrite.diff
dpkg-source: info: applying ares_expand_name-should-escape-more-characters.patch
dpkg-source: info: applying ares_expand_name-fix-formatting-and-handling-of-root.patch
dpkg-buildpackage: info: source package c-ares
dpkg-buildpackage: info: source version 1.17.1-1+deb11u1
dpkg-buildpackage: info: source distribution bullseye-security
dpkg-buildpackage: info: source changed by Salvatore Bonaccorso <carnil@debian.org>
dpkg-buildpackage: info: host architecture amd64
 dpkg-source --before-build .
 fakeroot debian/rules clean
dh clean
   dh_clean
 debian/rules build
dh build
   dh_update_autotools_config
   dh_autoreconf
aclocal: installing 'm4/ax_ac_append_to_file.m4' from '/usr/share/aclocal/ax_ac_append_to_file.m4'
[...]

PR #418, opened 2021-08-17
Fix building with latest autotools

felixonmars:
Turns out the second build error in #416 comes from `@CODE_COVERAGE_RULES@` being left out at the file:

```
# Makefile.in generated by automake 1.16.4 from Makefile.am.
# src/lib/Makefile.  Generated from Makefile.in by configure.
...
```

ax_code_coverage.m4 needs an update and Makefile.am needs to be updated
accordingly. Resolves the second failure in #416.


bradh352:
Ah, yes, it does appear the current `ax_code_coverage.m4` dropped `@CODE_COVERAGE_RULES@` ... interesting.
Makes sense if `autoreconf -fi` is installing the later version that dropped that, it would break everything.

Issue #399, opened 2021-03-03
ares_gethostbyname fails to resolve localhost on Windows


c29reid:
I'm using curl + ares on windows and it fails to resolve localhost. Curl with the windows resolver works.

It looks like this issue came up a couple times in the past (in both https://daniel.haxx.se/blog/2011/02/21/localhost-hack-on-windows/ and #100).
There was a patch that was made a while back on the mailing list that seems to have fallen through https://c-ares.haxx.se/mail/c-ares-archive-2011-03/0021.shtml.
The main concerns with the patch seemed around behaviors with XP https://c-ares.haxx.se/mail/c-ares-archive-2011-04/0001.shtml.

Is the localhost workaround with windows something that's still open to being added?
If so, would it be okay to try and re-land those original patches and are those concerns with XP still relevant?

bradh352:
If you want to submit a PR, I think that would be accepted.  There's definitely no concerns about XP any longer.  Any XP-based systems are long since EOLd including the POSReady-based systems.

bradh352:
@c29reid, @nijikokun.  I'm seeing some issues reference this ticket, but those issues appear to be unrelated  to this issue that was opened.  The issue that was opened was about resolving the hostname of localhost using c-ares on Windows (which c-ares queries the Windows /etc/hosts equivalent which removed the entry after XP I think), but the references to this ticket appear to be talking about using a resolver running on localhost not being queried.

Can more information be provided on the latter issue?  Or if you think they are inter-related somehow, I'd need to know how.

PR #428, opened 2021-10-11
Reimplement ares_gethostbyname() by wrapping ares_getaddrinfo()

bradh352:
`ares_gethostbyname()` and `ares_getaddrinfo()` do a lot of similar things, however `ares_getaddrinfo()` has some desirable behaviors that should be imported into `ares_gethostbyname()`.  For one, it sorts the address lists for the most likely to succeed based on the current system routes.  Next, when `AF_UNSPEC` is specified, it properly handles search lists instead of first searching all of `AF_INET6` then `AF_INET`, it searches in parallel.  Therefore, this PR should also resolve the issues attempted in #94.

A few things this PR does:
1. `ares_parse_a_reply()` and `ares_parse_aaaa_reply()` had very similar code to translate `struct ares_addrinfo` into a `struct hostent` as well as into `struct ares_addrttl/ares_addr6ttl`  this has been split out into helper functions of `ares__addrinfo2hostent()` and `ares__addrinfo2addrttl()` to prevent this duplicative code.
2. `ares_getaddrinfo()` was apparently never honoring `HOSTALIASES`, and this was discovered once `ares_gethostbyname()` was turned into a wrapper, the affected test cases started failing. 
3. A slight API modification to save the query hostname into `struct ares_addrinfo` as the last element of `name`.  Since this is the last element, and all user-level instances of `struct ares_addrinfo` are allocated internally by c-ares, this is not an ABI-breaking change nor would it impact any API compatibility.  This was needed since `struct hostent` has an `h_name` element.
4. Test Framework: MockServer tests via TCP would fail if more than 1 request was received at a time which is common when `ares_getaddrinfo()` queries for both A and AAAA records simultaneously.  Infact, this was a long standing issue in which the `ares_getaddrinfo()` test were bypassing TCP alltogether.  This has been corrected, the message is now processed in a loop.
5. Some tests had to be updated for overall correctness as they were invalid but somehow passing prior to this change.

### DIFF
--- a/src/lib/Makefile.inc
+++ b/src/lib/Makefile.inc
@@ -1,5 +1,5 @@
 
-CSOURCES = ares__addrinfo2hostent.c \
+CSOURCES = ares__addrinfo2hostent.c	\
   ares__close_sockets.c	\
   ares__get_hostent.c			\
   ares__parse_into_addrinfo.c		\


bradh352:
If there's no comments, I'll probably just merge it and deal with the fallout.  This blocks me from working on #429 & #399

bagder:
> I'll probably just merge it and deal with the fallout.

That's fine with me. I've started a review and it looks fine but I haven't managed to get through it all...

ChristianAmmer:
@bradh352 sorry for the late reply. Your PR accomplishes the goal from PR #94, thanks for taking care. I'm going to do a review, but I'm not that familiar with the code anymore.

PR #436, opened 2021-11-02
Fix cross-compilation from Windows

reynoldsbd:
When determining value for CPACK_PACKAGE_ARCHITECTURE, prefer to use
value from CMAKE_SYSTEM_PROCESSOR before falling back to uname output.

Additionally, if building from a Windows host, emit a fatal error
instead of attempting to call uname.


bradh352:
This is cross compiling from Windows to Linux ?

reynoldsbd:
Yes that's correct. To be precise, my team is using vcpkg with a custom triplet and toolchain, building from Windows, targeting Linux.

PR #445, opened 2021-11-30
Windows: Clean out old pre-vista code that very likely was no longer working anyhow

bradh352:
There was a lot of windows initialization code specific to the era that predates Windows Vista such as reading DNS configuration from the registry, and dynamically loading libraries to get access to functions that didn't exist in XP or earlier releases. 

Vista was released in January 2007, and was EOL'd in 2017, and support for Vista is still maintained with this patch set. 

XP was EOL'd in Apr 8 2014.

I believe the last OS based on something earlier than Vista was POSReady 2009, as it was XP based for some reason, and that was EOL'd in January 2019.  Considering any POS system falls under the PCI-DSS rules, they aren't allow to run POSReady 2009 any more so there is no reason to try to continue supporting such systems.

We have also targeted with our build system Vista support for the last few years, and while developers could change the target, we haven't had any reports that they have.


bradh352:
no comments for over a week, guess I'll merge, seems safe to me

gvanem:
A good commit. I'd which Windows-CE support will be removed too.

Issue #462, opened 2022-04-14
Will support IPv6 link-local dns server address?


lianhansheng:
Dns server address like  fe80::cd6:1e60:fe42:b8b4%wlan0

bradh352:
@sergvpurik since you provided the original PR, can you see if PR #646 does what is needed?

@lianhansheng not sure if you can test this

sergvpurik:
@bradh352 
> can you see if PR https://github.com/c-ares/c-ares/pull/646 does what is needed?

Yes, it works.


For testing, I used IPv6 link-local IP of my router.
I use OpenWrt on my route, so it is easy to get console and find IPs.

More detailed steps:
1. Create "/etc/test-resolv.conf" with a router IP and specify a local interface name
```
echo nameserver fe80::xxxx:xxxx:xxxx:ae3e%wlx04xxxx > /etc/test-resolv.conf
```
2. Point system resolv.conf to /etc/test-resolv.conf
```
ln -s /etc/test-resolv.conf /etc/resolv.conf
```
3. Use c-ares adig tool to check DNS resolving
```
./adig -d www.facebook.com
```

4.  Restore default resolv.conf
```
cd /etc
ln -fsv ../run/systemd/resolve/stub-resolv.conf ./resolv.conf
```

Also, MacBook internet sharing can return IPv6 link-local for DNS and gateway.
Maybe, as I remember, the NAT64 option should be enabled.

bradh352:
You mentioned Mac here, have you actually been able to get your mac's dns configuration to show a link-local ipv6 address?
I have a MacOS vm and an Ubuntu VM both bridged to the same network adapter.  My Ubuntu VM picks up the router advertisement for DNS, but my Mac VM doesn't appear to at least according to:

`scutil --dns`

That said, I'm also getting an ipv4 dns server from DHCP so I don't know if that's causing the link-local RA to be ignored.

I have a similar issue on Windows, I do see it picking up the RA for the ipv6 default route, but not the RDNSS record.

Both Windows and MacOS work if I specify the link-local server to 'adig' with the appropriate interface, just not getting the system to actually record the dns server to properly test that aspect.

sergvpurik:
I don't have Mac for now. But I have an old log where I used rdisc6 and rdnssd.
I've tested this case on Linux, so I can say nothing about Mac or Windows clients.

The actual configuration of my Linux client was the next:
```
IP:       2001:2:0:xxxx:xxxx:xxxx:xxxx:b9
[...]

PR #463, opened 2022-04-18
IPv6 link-local servers processing

sergvpurik:
Enable support of IPv6 link-local servers. 
Tested on Linux only.

https://github.com/c-ares/c-ares/issues/462

### DIFF
--- a/include/ares.h
+++ b/include/ares.h
@@ -257,6 +257,7 @@ struct apattern;
    duplicates this new option.
 
  */
+
 struct ares_options {
   int flags;
   int timeout; /* in seconds or milliseconds, depending on options */
@@ -700,6 +701,9 @@ struct ares_addr_node {
     struct in_addr       addr4;
     struct ares_in6_addr addr6;
   } addr;
+  int ll_family;
+  unsigned int ll_scope;
+  struct ares_in6_addr ll_addr6;
 };
 
 struct ares_addr_port_node {
@@ -711,6 +715,9 @@ struct ares_addr_port_node {
   } addr;
   int udp_port;
   int tcp_port;
+  int ll_family;
+  unsigned int ll_scope;
+  struct ares_in6_addr ll_addr6;
 };
 
 CARES_EXTERN int ares_set_servers(ares_channel channel,
--- a/src/lib/ares__close_sockets.c
+++ b/src/lib/ares__close_sockets.c
@@ -22,6 +22,7 @@
 void ares__close_sockets(ares_channel channel, struct server_state *server)
 {
   struct send_request *sendreq;
+  server->ll_addr.family = AF_UNSPEC;
 
   /* Free all pending output buffers. */
   while (server->qhead)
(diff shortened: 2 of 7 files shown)


bradh352:
Can you elaborate more on what this patch is trying to accomplish?  Are you trying to implement LLMNR or something else?

sergvpurik:
@bradh352 
c-ares uses 'ares_inet_pton' to parse nameserver records from OS DNS configuration (/etc/resolv.conf).
So c-ares can't parse a DNS record with IPv6 addresses if a port or/and a network interface are specified.

Examples:
IPv6 record format with a port
[XXXX::XXXX:XXXX]:P

IPv6 record format with an interface
XXXX::XXXX:XXXX%IFNAME

IPv6 record format with a port and an interface
[XXXX::XXXX:XXXX]:P%IFNAME

Also in the case of IPv6 link-local DNS record, a local socket should be bound to an IPv6 link-local address of the local interface.

This PR fixes parsing of the IPv6 records if the port or/and an interface are specified.
Also, it fixes the local socket binding in case of using the IPv6 link-local address.

bradh352:
Can you share the specification enhancements for resolv.conf that allow that syntax?  I know systemd has some 'resolved.conf' file that supports a DNS syntax like that (and also allows things like  #example.com at the end to state to direct queries for that domain to that server).  I can't find where any docs say resolv.conf allows this though.

sergvpurik:
>Can you share the specification enhancements for resolv.conf that allow that syntax?
https://www.ietf.org/rfc/rfc2732.txt

sergvpurik:
About systemd.
[nameserver - is not only an IP address, It also may include a port and/or an interface name.](https://www.freedesktop.org/software/systemd/man/resolved.conf.html)

bradh352:
I don't see anywhere that says resolv.conf supports a bracketed syntax, that's typically only used if you're going to append a port, which resolv.conf doesn't support as far as I can tell.  We would really need to see a specification that updates the resolv.conf format with the modifications you are trying to make.

sergvpurik:
>I can't find where any docs say resolv.conf allows this though.

Actually to fix IPv6 link-local DNS support systems use nameserver format with an interface name specified.

http://lists.cluenet.de/pipermail/ipv6-ops/2014-July/010032.html
https://askubuntu.com/questions/819343/link-local-fe80-addresses-without-scope-make-their-way-into-etc-resolv-conf

bradh352:
@bagder any insight into this?

sergvpurik:
Another example is rdnssd.
As I know it can produce nameserver records with an interface name specified.
https://linux.die.net/man/8/rdnssd
```
rdnssd parses RDNSS options and keeps track of resolvers to write nameservers entries to a [resolv.conf](https://linux.die.net/man/5/resolv.conf)(5) configuration file.
```

69Bitch69 (commented):
http://www.flibustahezeous3.onion

bradh352:
We did end up adding port and bracketed syntax support as we found an openbsd spec for resolv.conf which explicitly allowed it in PR #534.  We also added support for this on MacOS and Windows.

That said, the specific syntax you're stating, while I do find references to it, its just not ever referenced as valid for resolv.conf.  I really think what we need to do is start reading files generated by systemd or network manager on Linux instead of resolv.conf as it seems like resolv.conf is essentially being deprecated and only exists for legacy applications.

bradh352:
Closing for PR #646

PR #465, opened 2022-05-03
Configurable hosts path for file_lookup

mmehra:
This changeset adds support for configurable hosts file
ARES_OPT_HOSTS_FILE (similar to ARES_OPT_RESOLVCONF).

c-ares does support configurable hosts file via an environment
varibale (ARES_AI_ENVHOSTS). However, if application wants to
support multiple VRF/VPN/namespaces, this environment variable
does not help much

### DIFF
--- a/docs/ares_init_options.3
+++ b/docs/ares_init_options.3
@@ -41,6 +41,7 @@ struct ares_options {
   int nsort;
   int ednspsz;
   char *resolvconf_path;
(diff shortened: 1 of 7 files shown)


dn-codes:
does this issue have/require a CVE-ID?

bradh352:
@dn-codes this is a feature, not a bug or security fix

PR #468, opened 2022-05-27
Don't include version info in the static library

marc-groundctl:
The static library should not contain version info, since it would be linked into an executable or dll with its own version info.


bradh352:
The static library should not contain version info, since it would be linked into an executable or dll with its own version info.

Fix By: @marc-groundctl

Issue #488, opened 2022-10-18
DNS over TCP: write_tcp_data() does not check number of iovec entries


dreibh:
write_tcp_data() does not check number of iovec entries is creates for queries. If the number exceeds __IOV_MAX (Linux: 1024), the following writev() will fail. In this case, the queries fail with ARES_ENOTFOUND although none of these queries have been sent to a DNS server.

The fix is simply to stop when the number of iovec entries has reached __IOV_MAX:
```
          for (sendreq = server->qhead; sendreq; sendreq = sendreq->next)
            {
              vec[n].iov_base = (char *) sendreq->data;
              vec[n].iov_len = sendreq->len;
              n++;
              if(n >= __IOV_MAX)
                break;
            }
```
The following call to advance_tcp_send_queue() is based on the actual number of bytes sent.

dreibh:
Note that filling the queue with more than __IOV_MAX requests leads to a failure of sending any request, as long as the queue size is not reduced (timeout?). This may be used in a denial of service attack.

bradh352:
Thanks, merged

PR #489, opened 2022-10-18
Fix for issue #488: ensure that the number of iovec entries does not exceed __IOV_MAX.

dreibh:
Fix for issue #488: ensure that the number of iovec entries does not exceed __IOV_MAX.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -256,6 +256,8 @@ static void write_tcp_data(ares_channel channel,
               vec[n].iov_base = (char *) sendreq->data;
               vec[n].iov_len = sendreq->len;
               n++;
+              if(n >= __IOV_MAX)
+                break;
             }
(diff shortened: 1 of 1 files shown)


gvanem:
This is Linux specific. On Windows, I get:
```c
src/lib/ares_process.c(259,23): error: use of undeclared identifier '__IOV_MAX'
              if(n >= __IOV_MAX)
                      ^
```

dreibh:
Indeed, this was not platform-independent. It also failed on FreeBSD. I have no Windows system for testing.

I updated the code, according to the documentation in the writev() manpage. IOV_MAX is used on FreeBSD, on modern Linux the limit is from sysconf(_SC_IOV_MAX), on very old Linux it is 16. The default of 16 is also used by the BOOST libraries, in case of IOV_MAX and sysconf(_SC_IOV_MAX) not being possible.

PR #491, opened 2022-10-19
Add include guards to ares_data.h

bsergean:
I am getting a compilation error somehow telling me about this problem. All the other header files in the src/lib folder do have an include guard so it look like an overthought.


bsergean:
Thanks for merging !

PR #495, opened 2022-11-08
ares_strsplit: add an underscore for internal name

bagder:
We use the `ares__` prefix for internal functions. One underscore for public functions. This is an internal function.


bradh352:
i committed the change separately as this was missing tests

PR #497, opened 2022-12-13
Add str len check in config_sortlist to avoid stack overflow

### LINKED ISSUE #496: Potential stack overflow in ares_set_sortlist
In `ares_set_sortlist`, it calls `config_sortlist(..., sortstr)` to parse 
the input str and initialize a sortlist configuration.

However, `ares_set_sortlist` has not any checks about the validity of the input str. 
It is very easy to create an arbitrary length stack overflow with the unchecked 
`memcpy(ipbuf, str, q-str);` and `memcpy(ipbufpfx, str, q-str);`
statements in the config_sortlist call, which could potentially cause severe 
security impact in practical programs.

For example, the litter case could produce a stack overflow:
```
#include "ares.h"
#include<stdlib.h>
typedef int8_t  i8;
typedef int32_t i32;
int main() {
    struct ares_channeldata *v0 = NULL; // channel
    i32 v2 = ares_init(&v0); // $relative
    i8 v5[] = "111.111.111.111*shellcode*shellcode*shellcode*/"; // sortstr
    i32 v6 = ares_set_sortlist(v0, v5); // $target
}
```

As `ares_set_sortlist` is a public interface of a widely-used library,
i think it is necessary to add some checks to forbid further severe security bugs.

hopper-vul:
In `ares_set_sortlist`, it calls `config_sortlist(..., sortstr)` to parse the input str and initialize a sortlist configuration.

However, `ares_set_sortlist` has not any checks about the validity of the input str. It is very easy to create an arbitrary length stack overflow with the unchecked `memcpy(ipbuf, str, q-str);` and `memcpy(ipbufpfx, str, q-str);` statements in the config_sortlist call, which could potentially cause severe security impact in practical programs.

This commit add necessary check for `ipbuf` and `ipbufpfx` which avoid the potential stack overflows.

fixes #496

Signed-off-by: hopper-vul <hopper.vul@gmail.com>

### DIFF
--- a/src/lib/ares_init.c
+++ b/src/lib/ares_init.c
@@ -1913,6 +1913,8 @@ static int config_sortlist(struct apattern **sortlist, int *nsort,
       q = str;
       while (*q && *q != '/' && *q != ';' && !ISSPACE(*q))
         q++;
+      if (q-str >= 16)
+        return ARES_EBADSTR;
       memcpy(ipbuf, str, q-str);
       ipbuf[q-str] = '\0';
       /* Find the prefix */
@@ -1921,6 +1923,8 @@ static int config_sortlist(struct apattern **sortlist, int *nsort,
           const char *str2 = q+1;
           while (*q && *q != ';' && !ISSPACE(*q))
             q++;
+          if (q-str >= 32)
+            return ARES_EBADSTR;
           memcpy(ipbufpfx, str, q-str);
           ipbufpfx[q-str] = '\0';
           str = str2;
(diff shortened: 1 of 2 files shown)


bradh352:
I'm not aware of any users of this interface, but in general it just looks broken to me.

It looks like via the documentation its supposed to *only* take ip addresses or ip addresses with subnet masks, however looking at the pre-existing test cases, its accepting garbage and actively being tested to ensure it accepts garbage ???

mmuehlenhoff:
JFTR, this was assigned CVE-2022-4904 (via https://bugzilla.redhat.com/show_bug.cgi?id=2168631)

PR #502, opened 2023-02-01
Fix stop to use memory in stack

Chilledheart:
char PATH_HOSTS[MAX_PATH] at L448 is memory in stack and becomes invalid outside the scope. It is unsafe to use it.

Found by doing ares_getaddrinfo on resolving "localhost" in Release build (win32).

### DIFF
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -430,16 +430,20 @@ static int file_lookup(struct host_query *hquery)
   FILE *fp;
   int error;
   int status;
-  const char *path_hosts = NULL;
+  char *path_hosts = NULL;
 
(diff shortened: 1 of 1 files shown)


bradh352 (changes requested):
The prototype for path_hosts would need to be char * not const char *, and every place it gets set would need to have an ares_strdup() ... there's one missing path.  Most likely this is what is causing the windows test failures.

Chilledheart:
> The prototype for path_hosts would need to be char * not const char *, and every place it gets set would need to have an ares_strdup() ... there's one missing path. Most likely this is what is causing the windows test failures.

Done. Add ares_strdup to the missing path (compared to 1.18.1). And it will return `ARES_ENOMEM` error code correctly if ares_strdup fails.

PR #503, opened 2023-02-01
Update cmake minimum version

2021-lex:


bradh352:
Can you describe why you think we need to bump the cmake minimum version?  Is there some feature we're using not available in 3.1.0?  Also, 3.20 seems pretty aggressive, likely many OSs won't have that recent a version.

2021-lex:
Ok, 3.20 maybe was a bit much... 
It is for the CMAKE_CXX_VISIBILITY_PRESET that I like minimum version to be above 3.3

bradh352:
It doesn't appear we're using CMAKE_CXX_VISIBILITY_PRESET, and if we were, it would only be for the test cases as c-ares itself doesn't use any C++ code, but the test cases are using gmock which is C++.

2021-lex:
Sorry it should be:
CMAKE_C_VISIBILITY_PRESET
https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html?highlight=cmake_cxx_visibility_preset

bradh352:
closing since this causes issues

Issue #458, opened 2022-03-07
Optimize ares__readaddrinfo (or find a way not to call it)


bsergean:
Hi there,

We just started to use c-ares (through libcurl), and we noticed a fair amount of time spent in ares__readaddrinfo.

It's not clear to me what is expensive yet, whether it's the disk IO or some computation inside the function itself.

```
$ wc -l /etc/hosts
7878 /etc/hosts
$ wc -c /etc/hosts
432830 /etc/hosts
```

Our host file is close to 500k, so definitely not a trivial amount of bytes to read.

```
>>> 432830 / 7878
54
```

1. Is it feasible to only read that file once, keep either a list of entries, or a hashmap if applicable ?
2. We can also just provide the content of the path as a memory buffer, and it would be up to the caller to detect when the file is changed, and re-read the /etc/host file from disk ?

I looked at the init options, and maybe we could already  for https://c-ares.org/ares_init_options.html

curl offers to tweak the c-ares resolver behavior thanks to https://curl.se/libcurl/c/CURLOPT_RESOLVER_START_FUNCTION.html
Is there anything we could do there.

I'd be happy to try something if time allows, if I can get some directions on a good thing to try it would be good.

bradh352:
I don't think I necessarily have a problem with the idea of caching the /etc/hosts data.  Afterall, we cache the server configs (/etc/resolv.conf).  That said, I'd like to see an overall solution that would also reload the rest of the configuration as it has been a common request (Issue #301 remains open to track this).

bsergean:
The linked ticket reference reloading the /etc/resolv.conf file, I looked at the proposed patch which was closed for lack of activity (https://github.com/c-ares/c-ares/pull/272/files).

Maybe a first piece of work would be to store the /etc/host data in memory (say a list of entries), and have are__readaddrinfo consume that. I don't know where the info from parsing the resolv.conf file lives, but maybe it could live close by.

bsergean:
Since ares__readaddrinfo is internal, maybe we have more room in modifying it to the needs of this ticket.

bradh352:
PR #272 did go stale and was a linux-only solution which unfortunately won't work for a cross-platform system.  It also would likely cause crashes or some undefined behavior if there were any outstanding requests due to changing the servers object out from under active requests that hadn't yet received replies.

bsergean:
Got it. Maybe the simplest approach to make progress would be to be able to pass in the host data as memory buffer (a char*), rather than a file path.

[...]

PR #514, opened 2023-02-09
remove union type from ares_expand functions

createyourpersonalaccount:
The union type in this context is used for the purpose of type punning. However, there is no need for it, and it is also not used correctly: Union members must have equivalent types, and there is no named signed integer type equivalent to size_t (not in ISO C nor POSIX.)

Theoretically, one can run into trouble if ares_ssize_t is not equivalent to size_t, as some bytes in size_t would be indeterminate, possibly affecting its value.

The union code is replaced with bounds-checked integer conversion code.

### DIFF
--- a/src/lib/ares_expand_name.c
+++ b/src/lib/ares_expand_name.c
@@ -31,8 +31,10 @@
 /* Maximum number of indirections allowed for a name */
 #define MAX_INDIRS 50
 
-static int name_length(const unsigned char *encoded, const unsigned char *abuf,
-                       int alen, int is_hostname);
+static int compute_lengths(const unsigned char *encoded,
+                           const unsigned char *abuf,
+                           size_t alen, int is_hostname,
+                           size_t *name_length, size_t *enclen);
 
 /* Reserved characters for names that need to be escaped */
 static int is_reservedch(int ch)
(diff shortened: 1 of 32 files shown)


createyourpersonalaccount:
Some force-pushes because I got the header for `SIZE_MAX` wrong. Apparently it's not in `limits.h` but `stdint.h`, which is missing from C89, thus the CPP check.

bradh352 (commented):
I think the only odd thing I see here is name_length() returns 'int' and not ares_ssize_t, so the whole bounds check just looks weird to me.  I'm pretty sure given a signed integer, there is no way on any system that nlen >= SIZE_MAX.

Also since ares_malloc() is going to take a size_t, and requires nlen + 1, I'd think that means we need SIZE_MAX-1 as the upper bounds for nlen, otherwise it will overflow.

createyourpersonalaccount:
@bradh352 it is checked that nlen < SIZE_MAX with the nlen >= SIZE_MAX assertion so nlen + 1 is safe.

My concern about union is that it is a complicated bit of code, which the reviewer must consider for every possible definition of ares_ssize_t. The alternative with casts is safer. The SIZE_MAX check is a bounds check, which in general you want, but can have less of by considering each system separately and having knowledge of type sizes.

Obviously optimizations come and go, but I tend to target correctness first, and leave optimization for when the need arises. For example I am currently working to provide a fix for https://github.com/c-ares/c-ares/issues/458 I think this is a more important optimization than another if-branch in the bounds-checking code in this commit (since there is someone who wants it), and in general the union bug may not be important as it hasn't been yet demonstrated, so altogether this commit can be ignored if necessary.

createyourpersonalaccount:
Also, I considered modifying the signature of name_length as you said, and that might happen too. I provided this quick patch now to fix the union issue.

bradh352:
how does this relate to #458?  I'm not seeing the connection

createyourpersonalaccount:
@bradh352 It does not relate. My message was confusing. I'm working on #458 so I was eager to let you know. Progress will be slow-paced because I only have a bit of free time during the week, but I anticipate that in a few weeks it will be done.

As for this PR, your criticism is valid, the name_length function should be modified.

bradh352:
@createyourpersonalaccount any chance you could go ahead and make the corrections mentioned so I can merge this?

src/lib/ares_expand_name.c
```diff
@@ -120,24 +126,20 @@ int ares__expand_name_validated(const unsigned char *encoded,
...
   const unsigned char *p;
-  union {
-    ares_ssize_t sig;
-     size_t uns;
-  } nlen;
 
-  nlen.sig = name_length(encoded, abuf, alen, is_hostname);
-  if (nlen.sig < 0)
+  nlen = name_length(encoded, abuf, alen, is_hostname);
+  if (nlen < 0 || nlen >= SIZE_MAX)
```
bagder, line 134: I'm not aware of a single architecture where `int` is larger than `size_t`, so this seems to check for something that cannot possibly happen?

createyourpersonalaccount:
I'm trying to work on this but I've noticed that main is not compiling, e.g. I get

    make[4]: *** No rule to make target 'ares_llist.c', needed by 'libcares_la-ares_llist.lo'.  Stop.

so I can't test my code right now on top of main. Is this something you are aware of?

dfandrich:
Try 'make clean' and rebuilding the autoconf files.

createyourpersonalaccount:
@dfandrich Thanks, moving on!

createyourpersonalaccount:
I want to make some comments on this patch.

I had some issues comparing my `ares__expand_name_validated()` against the old one. The test `LibraryTest.ParseRootName` was failing (among some others.) It turned out to be a difference in how the label `01 00` was being treated; the response is the "root" label. I could be wrong, but I think that RFC 1035 does not mention such a response as valid. Does it occur anyhow?

So these are the two issues I've encountered:

1. Using autoconf, I was not able to debug the test binary. I tried:

        autoreconf -fi
        mkdir build && cd build
        ../configure CFLAGS="-ggdb -g3 -Og" CXXFLAGS="-ggdb -g3 -Og" && make && make check
        ./libtool --mode=execute gdb --args test/arestest --gtest_filter=LibraryTest.ParseRootName

    But when I set a breakpoint for `ares_parse_a_reply()` and the subsequent `ares__parse_into_addrinfo()` I got the message from gdb that there was no line number information.

    I had to build with CMake to get debugging symbols, e.g.

        cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DCARES_BUILD_TESTS=ON && cmake --build build


2. The result of `name_length()` for the `01 00` response surprised me: it calculated `enclen` to be `3`. Shouldn't it be 2? Or is the full response `01 00 00` with the final byte being the terminating NUL byte? Where is this documented in RFC 1035?

dfandrich:
> I had to build with CMake to get debugging symbols, e.g.

You could be able to get this with autoconf using configure --enable-debug, or
by setting CFLAGS='-g'

createyourpersonalaccount:
> You could be able to get this with autoconf using configure --enable-debug, or by setting CFLAGS='-g'

Thank you, --enable-debug worked!

createyourpersonalaccount:
I'm not satisfied with this patch because I keep missing places where `int` or `long` pointers are passed to `size_t *` and I've turned up at least one occurrence where the type `int` is used both for its positive and negative values, which means even more code must be refactored. I don't know if it is worth it, there seems to be checks for the value ranges too, I saw at least one such comment. Therefore I'm closing this to save you the time of review.

PR #526, opened 2023-05-25
rand: add support for getrandom()

bnoordhuis:
glibc provides arc4random_buf() but musl does not and /dev/urandom is not always available.


src/lib/ares_rand.c
```diff
@@ -241,6 +241,21 @@ static void ares__rand_bytes(ares_rand_state *state, unsigned char *buf, size_t
...
+          ssize_t rv = getrandom(buf + bytes_read, n > 256 ? 256 : n, 0);
+          if (rv <= 0)
+            break;
+
+          bytes_read += rv;
+          if (bytes_read == len)
+            return;
+        }
+        break;
 #else
```
bradh352, line 259: I think the only problem here is the failure condition breaks but that doesn't cause it to go to another subsystem.  That said, according to the docs I don't think its actually possible to fail with 256 bytes (but might hang).

I'd be ok if for clarity you just changed the break on rv <= 0 to a continue as that ends up being the same actual behavior, but shows that its intentional.

bradh352: Not terribly sure how realistic that would be to really occur and be expected to still do something reasonable.  But if you want to fall back, that's fine ... but it should _remember_ that and not ever try getrandom() again ... so that if ARES_RAND_FILE fails it doesn't fall back to getrandom() instead of falling through to the random of last resort of rc4 seeded with rand().

But do remember, this is really a best effort thing, it doesn't actually need to be cryptographically secure, just not trivially predictable (which previously it was basically trivially predictable).  So cost/benefit analysis of doing too much logic here may not be worth it.

bnoordhuis: Fair point, I'll just `s/break/continue` then.

> Not terribly sure how realistic that would be to really occur and be expected to still do something reasonable.

You would not believe how many bug reports we get over at libuv that boil down to overzealous seccomp filters 🤦

PR #529, opened 2023-06-06
provide SPDX identifier and a REUSE CI job to verify

bagder:
All files should get their licenses and copyright information clearly identifiable.

## Licenses used

- BSD-3-Clause
- FSFAP
- GPL-3.0-or-later with Autoconf-exception-3.0
- LGPL-2.1-or-later
- MIT


bagder:
Current complication: the remaining m4 files.

bradh352:
For files from an upstream provider (specifically some of the m4 files), I'm not sure it makes sense to add the SPDX identifier as it will likely be overwritten when it syncs with upstream.  Plus not keeping them 1:1 seems a bit strange.

bagder:
Good point. I'll add them separately in the `.reuse/dep5` file

PR #531, opened 2023-06-22
fix: -Wunreachable-code warning in ares__init_rand_engine()

ckerr:
I'm one of the Electron maintainers and came across this issue when updating our Node.js snapshot, which uses c-ares. Since Electron builds Node.js with `-Werror` and `-Wunreachable-code`, this caused a FTBFS on Win32 until we patched it.

This doesn't show up under the default build flags; but since it does show up for anyone using those flags, I thought I'd upstream the patch.

### DIFF
--- a/src/lib/ares_rand.c
+++ b/src/lib/ares_rand.c
@@ -58,6 +58,8 @@ BOOLEAN WINAPI SystemFunction036(PVOID RandomBuffer, ULONG RandomBufferLength);
 #endif
 
 
+#if defined(CARES_RANDOM_FILE) && !defined(HAVE_ARC4RANDOM_BUF) && !defined(HAVE_GETRANDOM) && !defined(_WIN32)
+
 #define ARES_RC4_KEY_LEN 32 /* 256 bits */
 
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


bradh352:
This doesn't look right.  The rc4 code doesn't depend on CARES_RANDOM_FILE being set at all.  And you have a dangling #endif.

ckerr:
Hi @bradh352, thanks for the review.

On the second point, it's not a dangling `#endif`; it's just moved from a few lines further up.

But I I think your comment about `CARES_RANDOM_FILE` correct and more important than the `#endif` --  `ares_rc4_init()` needs to exist as the fallback impl both when `CARES_RANDOM_FILE` isn't defined _and_ if when it's defined but opening the random file fails. I'll update the patch.

ckerr:
Ugh.

Instead of adding several more `#ifdef`s to completely isolate `ares_rc4_*` and `ares_rand_rc4`, I tried just making it always available as a fallback RNG in `ares__init_rand_engine()` even in cases where Plan A can't fail e.g. `ARES_RAND_OS` but this just moves the warnings elsewhere.

Marking as a draft; I'm going to tinker with this to try & find something that's less intrusive but also solves the warnings.

PR #535, opened 2023-07-17
Fix Segment Fault in `ares_process.c`

lixin-wei:
I met 2 segment faults in my project. Crashed in the `if` statement at line 576 and 620:
<https://github.com/c-ares/c-ares/blob/bad62225b7f6b278b92e8e85a255600b629ef517/src/lib/ares_process.c#L572-L583>

At line 576, GDB shows that `query` is NULL, so `query->timeout` will cause a segment fault.

<https://github.com/c-ares/c-ares/blob/bad62225b7f6b278b92e8e85a255600b629ef517/src/lib/ares_process.c#L616-L625>

At line 620, GDB shows that `q` is NULL, so `q->qid` will cause a segment fault.

------
I fixed them intuitively: check whether it is NULL first.

```cpp
// added a `query &&`
if (query && query->timeout.tv_sec && ares__timedout(now, &query->timeout))
// added a `q &&`
if (q && (q->qid == id) && same_questions(q->qbuf, q->qlen, abuf, alen))
```
Then everything works fine in my project.

But I don't know why `query` and `q` are NULL here. Is there a further root bug? I'll appreciate it if anyone could explain to me.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -617,7 +617,7 @@ static void process_answer(ares_channel channel, unsigned char *abuf,
        list_node = list_node->next)
     {
       struct query *q = list_node->data;
-      if ((q->qid == id) && same_questions(q->qbuf, q->qlen, abuf, alen))
+      if (q && (q->qid == id) && same_questions(q->qbuf, q->qlen, abuf, alen))
         {
           query = q;
           break;
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -573,7 +573,7 @@ static void process_timeouts(ares_channel channel, struct timeval *now)
         {
           query = list_node->data;
           list_node = list_node->next;  /* in case the query gets deleted */
-          if (query->timeout.tv_sec && ares__timedout(now, &query->timeout))
+          if (query && query->timeout.tv_sec && ares__timedout(now, &query->timeout))
             {
               query->error_status = ARES_ETIMEOUT;
               ++query->timeouts;


bradh352:
Is it possible you have 2 threads operating on the same ares channel at the same time ?

lixin-wei:
I do use multiple threads, but I'm not using `ares` directly. I'm using `curl`'s curl_easy_perform, with multiple threads. 

Should this cause the problem?

bradh352:
Unfortunately I don't have visibility into how curl handles multithreading in relation to c-ares.  If you've initialized a separate curl handle per thread, I'd have to imagine its not an issue, but if you're sharing handles, it could be.

I will say based on what you're saying "fixes" your issue, it definitely looks like another thread is manipulating the c-ares channel simultaneously as I'm pretty sure what you're saying is happening should otherwise be impossible.

Maybe @bagder has more insight.

lixin-wei:
> If you've initialized a separate curl handle per thread, I'd have to imagine its not an issue, but if you're sharing handles, it could be.

I think I use curl in the right way. I called `curl_easy_init` to get a handle on each thread.

>  I'm pretty sure what you're saying is happening should otherwise be impossible.

One more piece of information.  This segment fault only happens when I set `http_proxy` to a very unstable server. Maybe this server sent some strange response to `ares` and caused the error?

bradh352:
I wouldn't think if you're using an http proxy that c-ares would be in use at all since it would be the proxy server doing the remote server dns resolution.  Have you tried running this through AddressSanitizer or Valgrind to see if there are any memory corruption issues reported?

bradh352:
this code has recently changed with a PR so it can't apply, and it is likely a multi-threading related issue anyhow since the linked list would never store a null query pointer.

Issue #444, opened 2021-11-30
Need ability to limit maximum queries per UDP port


crrodriguez:
at least when using
 adig foo.com bar.com 

it appears c-ares uses the same socket twice to send the queries out.. at least for UDP it shouldn't, as the little security DNS has left depends on each query coming from a different source port, hopefully with a good randomization provided by the OS.

This is more like a suggestion to change the default beahviour to use one socket per query and close agressively.

bradh352:
The problem with this request is you could quickly eat up all file descriptors available to the process and would be a denial of service attack.  Its not uncommon for a server to have hundreds of outstanding DNS requests under high load depending on what they're doing.  I could see closing after X requests or similar, but not every request.

crrodriguez:
Wel.. yeah. and those people have almost certainly changed the RLIMIT_FILE of their systems .. avoided select.. avoided ares_getsock() .. a middle ground would be starting reuse after $ARBITRAYNUMBER of outstanding DNS requests,,

TXTT2016:
I have met the same problem.In some special network env, when the v4 and v6 udp requires packages were sent from the same port, the firewall sometimes will block one of them. This make remote only receive the v4 or v6 packages and only reply one, then the client dns resolve will timeout.

This happened on iOS platform, sometimes windows also. And when use system interface to do the dns resolve, this won't happened, the v4 and v6 dns packages will send from two different ports.

bradh352:
What firewall/IDS is blocking that behavior?  I'd like to hear from the vendor.  By no means should multiple DNS queries on the same port be considered invalid.  Yes, having a maximum number of queries per port would make sense from a security standpoint, but limiting it to 1 query per port does not.

TXTT2016:
> What firewall/IDS is blocking that behavior? I'd like to hear from the vendor. By no means should multiple DNS queries on the same port be considered invalid. Yes, having a maximum number of queries per port would make sense from a security standpoint, but limiting it to 1 query per port does not.

Maybe c-ares can provide an option to choose use one or two port to send the V4 and V6 UDP packages ?

crrodriguez:
i am not buying the fd number argument really.. one can easily rise the number of fds except if we are going to hit the select-brick-wall. (macos users can avoid it by  _DARWIN_UNLIMITED_SELECT but that's it)

@TXTT2016 bind has an option like this but it actually shouldn't. there is no way to select the correct source port from userspace at least on linux, because only the kernel knows what is right.

[...]

Issue #301, opened 2020-01-31
channel: re-resolved servers under certain circumstances


junr03:
Right now a channel's servers do not get updated after initialization. There are certain circumstances under which it would be useful to have c-ares functions re-resolved the channel's servers. For example, https://github.com/envoyproxy/envoy/issues/4543#issuecomment-580915046. This PR started on the right track: https://github.com/c-ares/c-ares/pull/272.

sterchelen:
Hey, 
I'm working on re-using the PR #272 to implement a `reinit` function.
The idea is on the `ares_gethostbyname` fct, call a new fct named `is_ares_reinit_necessary`. Based on the result, calling `ares_reinit`.

Basically, `is_ares_reinit_necessary` would allow to implement platform specific scenarios to trigger a reinit... 
- Is the resolv.conf file has changed?
-

bradh352:
fixed in #614

Issue #135, opened 2017-07-27
Add timeout for each query?


XadillaX:
Any plan to add a timeout parameter for each query function?

bradh352:
can you elaborate on what you are asking?  There are already configurable timeouts

XadillaX:
maybe I missed somewhere on the document? I didn't find how to configure it.

bradh352:
ares_init_options() ARES_OPT_TIMEOUT/ARES_OPT_TIMEOUTMS combined with ARES_OPT_TRIES determines the overall time any query can take.  If that's not enough, you can always handle the timeouts yourself by adjusting the timeout parameters passed to select() and possibly use ares_cancel() depending on your implementation.

XadillaX:
I meant for each query, not a channel.

If I use `ares_cancel`, will all active query in that channel be canceled?

bnoordhuis:
Yes, `ares_cancel()` will do that.

@bradh352 Is it currently possible to cancel a query without resorting to a channel-per-query model?  My understanding after reading of the code is that you can't and that is the feature that @XadillaX is asking for.

cjihrig:
It's also my understanding that the current code cannot cancel individual queries.

I put together a small proof of concept that I think would allow it at https://github.com/cjihrig/c-ares/commit/4103105aca3526ee465e2e7aca26b260403676ad. It would involve exposing queries to userland. I'm not familiar enough with this project and its maintainers to know if this is a route worth exploring in more detail or not.

PR #540, opened 2023-08-10
Modernization: Implement base data-structures and replace usage

bradh352:
c-ares currently lacks modern data structures that can make coding easier and more efficient.  This PR implements a new linked list, skip list (sorted linked list), and hashtable implementation that are easy to use and hard to misuse.  Though these implementations use more memory allocations than the prior implementation, the ability to more rapidly iterate on the codebase is a bigger win than any marginal performance difference (which is unlikely to be visible, modern systems are much more powerful than when c-ares was initially created).  

The data structure implementation favors readability and audit-ability over performance, however using the algorithmically correct data type for the purpose should offset any perceived losses.

The primary motivation for this PR is to facilitate future implementation for Issues #444, #135, #458, and possibly #301

A couple additional notes:
  - The ares_timeout() function is now O(1) complexity instead of O(n) due to the use of a skiplist.
  - Some obscure bugs were uncovered which were actually being incorrectly validated in the test cases.  These have been addressed in this PR but are not explicitly discussed.
  - Fixed some dead code warnings in ares_rand for systems that don't need rc4


bradh352:
FYI, windows build failed due to whitespace issues in Makefile.inc that I'll fix.  The darwin build issue I've researched extensively, it appears to be something related to some DNS server in Cirrus-CI not being able to handle large responses so it times out then returns a server failure.  It all depends on which Cirrus-CI node you hit.  At first I thought it was my issue, but that's how I uncovered some bugs in what response code c-ares was returning and this PR fixes that too.

bradh352:
@bagder don't suppose you can review today?  If not, I'll probably just go ahead and merge it so I can start work on the tickets referenced in the description.  I can't really imagine there's anything objectionable here.

bagder (approved):
:+1:

Issue #541, opened 2023-08-11
On PF_UNSPEC, skip retry on second address class if we have a reply on one


sachusachin622:
Hi,

**I am using curl 8.2.1 with c-ares 1.19.0**

The issue I am noticing with c-ares is, when we don't specify the ip address to query for a domain name using the curl command and there is a timeout associated to it the resolution fails for the domain and curl as a consequence also fails.

root@2d142b34d5a2:/# curl -m20 -v http://www.google.com
* processing: http://www.google.com
06:57:00.807787 IP 2d142b34d5a2.60042 > 10.174.75.167.domain: 29950+ A? www.google.com. (32)
06:57:00.807844 IP 2d142b34d5a2.60042 > 10.174.75.167.domain: 43180+ AAAA? www.google.com. (32)
06:57:00.808314 IP 2d142b34d5a2.47661 > 10.174.75.167.domain: 8116+ PTR? 167.75.174.10.in-addr.arpa. (44)
06:57:00.808714 IP 10.174.75.167.domain > 2d142b34d5a2.60042: 29950* 1/0/0 A 142.250.195.228 (49)
06:57:00.809018 IP 10.174.75.167.domain > 2d142b34d5a2.60042: 29950* 1/0/0 A 142.250.195.228 (49)
06:57:05.809237 IP 2d142b34d5a2.60042 > 10.174.75.167.domain: 43180+ AAAA? www.google.com. (32)
06:57:05.813398 IP 2d142b34d5a2.47661 > 10.174.75.167.domain: 8116+ PTR? 167.75.174.10.in-addr.arpa. (44)
06:57:15.811191 IP 2d142b34d5a2.60042 > 10.174.75.167.domain: 43180+ AAAA? www.google.com. (32)
* Resolving timed out after 20001 milliseconds
* Closing connection
curl: (28) Resolving timed out after 20001 milliseconds
root@2d142b34d5a2:/# 

If we don't pass the timeout for resolving the domain the curl passes by using the ipv4 address. But this generally take a long time to timeout and use the ipv4 address. In my testing its over a minute always before returning with the resolved address. Some program using curl generally doesn't wait so long. So is it possible to send the partial query response back to curl before it time's out, so that it can use the ipv4 address at least and continue with its operations?

My test environment is ubuntu 16.04

Thanks

bradh352:
[...]

PR #544, opened 2023-08-28
C-ares timeout fix

sachusachin622:
Fix for #541
@bradh352 @bagder Please help me to review
Thanks

### DIFF
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -576,6 +576,10 @@ static void host_callback(void *arg, int status, int timeouts,
   if (status == ARES_SUCCESS)
     {
       addinfostatus = ares__parse_into_addrinfo(abuf, alen, 1, hquery->port, hquery->ai);
+      if (hquery->hints.ai_family  == AF_UNSPEC && addinfostatus == ARES_SUCCESS && hquery->ai->nodes)
+        {
+          hquery->channel->ares_succ_resp_flag = 1;
+        }
     }
 
   if (!hquery->remaining)
@@ -748,6 +752,7 @@ void ares_getaddrinfo(ares_channel channel,
   hquery->next_domain = -1;
   hquery->remaining = 0;
   hquery->nodata_cnt = 0;
+  hquery->channel->ares_succ_resp_flag = 0;
 
   /* Start performing lookups according to channel->lookups. */
   next_lookup(hquery, ARES_ECONNREFUSED /* initial error code */);
--- a/src/lib/ares_init.c
+++ b/src/lib/ares_init.c
@@ -462,12 +462,20 @@ static int init_by_options(ares_channel channel,
   /* Easy stuff. */
   if ((optmask & ARES_OPT_FLAGS) && channel->flags == -1)
     channel->flags = options->flags;
-  if ((optmask & ARES_OPT_TIMEOUTMS) && channel->timeout == -1)
-    channel->timeout = options->timeout;
-  else if ((optmask & ARES_OPT_TIMEOUT) && channel->timeout == -1)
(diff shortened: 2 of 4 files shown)


bradh352:
I don't see how changing a per-attempt timeout would solve this problem.

sachusachin622:
> I don't see how changing a per-attempt timeout would solve this problem.

In the context of using Curl, if a timeout of 20 seconds (-m 20 option) is set, while C-ares employs DEFAULT_TIMEOUT as 5000 (5 seconds) and DEFAULT_TRIES as 4, the outcome is as follows: C-ares initiates the sending of requests in a sequence, with intervals determined by an exponential progression. Specifically, C-ares transmits the first request after a 5-second interval, the second request after a 10-second interval, the third request after a 20-second interval, and the fourth request after a 40-second interval. The cumulative time taken to receive a partial query response (IPV4 address) from C-ares to pass it to Curl totals 75 seconds (1 minute and 15 seconds). During this period, Curl, having already elapsed the 20-second timeout, has concluded with a failure response.

**Solution**
From the above explanation of C-ares timeout logic, C-ares should return the partial query response to curl within curl timeout. So that we can derive the timeout of C-ares from timeout of curl.
Timeout of C-ares = ( Timeout of curl / 2^number of retries )

Dividing the curl timeout by 2^number of retries to get Timeout of C-ares because C-ares follows an exponential pattern for sending successive requests.

Example:
Consider curl timeout is 20 seconds and c-ares number of retries is 4 then C-ares timeout value is 

Timeout of C-ares = (Timeout of curl/2^number of retries)

Timeout of C-ares = (20/2^4) = (20/16)

Timeout of C-ares = 1.25 seconds.

C-ares will send request every 1.25 seconds, 2.50 seconds , 5 seconds, and 10 seconds total 18.75 seconds required to send back response to curl. 
C-ares able to send the partial query response back to curl before curl gets time's out, so that it can use the ipv4 address at least and continue with its operations.

**Here is the packet capture for the above example:**

root@[bd05f1ebe90d:/curl-8.2.1#](http://bd05f1ebe90d/curl-8.2.1) time curl -m20 -v http://www.google.com/
* processing: http://www.google.com/
07:08:22.567944 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 58166+ A? [www.google.com](http://www.google.com/). (32)
07:08:22.568075 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
07:08:22.568758 IP 10.174.75.167.domain > bd05f1ebe90d.44543: 58166* 1/0/0 A 142.250.195.228 (49)
07:08:22.568767 IP 10.174.75.167.domain > bd05f1ebe90d.44543: 58166* 1/0/0 A 142.250.195.228 (49)
07:08:23.818355 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
07:08:26.319154 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
07:08:31.320595 IP bd05f1ebe90d.44543 > 10.174.75.167.domain: 37862+ AAAA? [www.google.com](http://www.google.com/). (32)
*   Trying 142.250.195.228:80...
* Connected to [www.google.com](http://www.google.com/) (142.250.195.228) port 80
> GET / HTTP/1.1


**Please guide me If anything wrong or missing here**

Thanks

bradh352:
your solution is definitely not right, a right-shift of the timeout by number of tries doesn't make any sense.  the proper solution is discussed in #541

sachusachin622:
Closing this PR and raising new PR #548 
Thanks

PR #548, opened 2023-09-14
Fix for ares dns timeout fix and Returning the available address

sachusachin622:
#544 As part of this PR raising this new PR.

C-ares should promptly return the partial query response to curl within the curl timeout. In other words, C-ares should immediately provide either an IPv4 or IPv6 address without waiting for a potentially failed query to complete.

Two scenarios illustrate this behavior:

Single Server Scenario: In the first scenario, when there is only a single server available, and the PF_UNSPEC case is encountered, C-ares will initiate queries for both IPv4 and IPv6 address classes. If the IPv4 query succeeds, it will immediately return the IPv4 address, bypassing the need to complete the IPv6 query.
Multiple Servers Scenario: In the second scenario, suppose there are two servers—S1 and S2. When facing the PF_UNSPEC case, C-ares will initiate queries for both IPv4 and IPv6 address classes. If the IPv4 query is successful but the IPv6 query fails with S1 server, C-ares will not proceed to retry the IPv6 query with S2 server. Instead, it will promptly return the available IPv4 address.
This approach ensures that C-ares does not unnecessarily delay the response to curl, enhancing efficiency and responsiveness.

**Changes:**
In this file [src/lib/ares_private.h](https://github.com/c-ares/c-ares/compare/main...sachusachin622:dev/ares-dns-timeout-fix?expand=1#diff-99be976f40a9f77bf713f4f1f878ba101d3d01fa9b9fbb41cc27a721935a9d53) 
we are declaring one new flag **ares_succ_resp_flag** to set this flag when we get the successful response.

In this file [src/lib/ares_getaddrinfo.c](https://github.com/c-ares/c-ares/compare/main...sachusachin622:dev/ares-dns-timeout-fix?expand=1#diff-8f022ca768aa72d2ef3c6774683fd3591eb1639966829d5157ea5cfdbc61d7b1)
We are setting this ares_succ_resp_flag

In this file [src/lib/ares_process.c](https://github.com/c-ares/c-ares/compare/main...sachusachin622:dev/ares-dns-timeout-fix?expand=1#diff-1cbf06a764b0c1e45d9090473de26423e9d0586021f6295caf4e5a785e700f3a)
We are checking the flag is set or not if it is one it will not send new query

Please help me to review. Is my changes are correct.

Thanks,
Sachin

### DIFF
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -576,6 +576,10 @@ static void host_callback(void *arg, int status, int timeouts,
   if (status == ARES_SUCCESS)
     {
       addinfostatus = ares__parse_into_addrinfo(abuf, alen, 1, hquery->port, hquery->ai);
+      if (hquery->hints.ai_family  == AF_UNSPEC && addinfostatus == ARES_SUCCESS && hquery->ai->nodes)
+        {
+          hquery->channel->ares_succ_resp_flag = 1;
+        }
     }
 
   if (!hquery->remaining)
@@ -748,6 +752,7 @@ void ares_getaddrinfo(ares_channel channel,
   hquery->next_domain = -1;
   hquery->remaining = 0;
   hquery->nodata_cnt = 0;
+  hquery->channel->ares_succ_resp_flag = 0;
 
   /* Start performing lookups according to channel->lookups. */
   next_lookup(hquery, ARES_ECONNREFUSED /* initial error code */);
--- a/src/lib/ares_private.h
+++ b/src/lib/ares_private.h
@@ -331,6 +331,7 @@ struct ares_channeldata {
 
   /* Path for hosts file, configurable via ares_options */
   char *hosts_path;
+  int ares_succ_resp_flag;
 };
 
 /* Does the domain end in ".onion" or ".onion."? Case-insensitive. */
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -764,7 +764,7 @@ static void next_server(ares_channel channel, struct query *query,
    * servers to try. In total, we need to do channel->nservers * channel->tries
    * attempts. Use query->try to remember how many times we already attempted
    * this query. Use modular arithmetic to find the next server to try. */
-  while (++(query->try_count) < (channel->nservers * channel->tries))
+  while ((++(query->try_count) < (channel->nservers * channel->tries)) && (!channel->ares_succ_resp_flag))
     {
       struct server_state *server;
 


sachusachin622:
@bradh352 @bagder Please help me to review
Thanks

bradh352:
Setting a status flag on the channel isn't going to be the right course of action, you could have thousands of queries on the same channel.

Instead, I think the better route would be to modify ares_send() to return the either the query pointer or the qid and store those as part of ares_getaddrinfo() (qid might be safer as otherwise you'd have to invalidate the query pointers that already returned results somehow).  Store that, within the ares_getaddrinfo's host_query structure, then once one result comes back, set a flag on any other spawned queries to not perform any more retries regardless of cause (server error, timeout, etc), in ares_process.c's next_server().  So it'll still delay until the *first* timeout is hit, but there won't be any other retries. Realistically this is the best you could do, short of maybe manipulating the timeout for the other queries to be more aggressive once you already have a partial result.

sachusachin622:
Hi @bradh352 ,

I have a doubt regarding a scenario involving two servers: S-1 (10.174.75.167) and S-2 (10.174.75.168). 
When using C-ares to send a query with the UNSPEC option, we receive a DNS response code of 0 (DNS_NOERROR). However, for the IPv4 address, we receive nothing (0.0.0.0), and IPv6 resolution fails when attempting query form S-1

**Here is the pcap**
root@9003533ae932:/# curl -m300 -v http://www.google.com
* processing: http://www.google.com
06:59:20.063010 IP 9003533ae932.45025 > 10.174.75.167.domain: 55528+ A? www.google.com. (32)
06:59:20.063078 IP 9003533ae932.45025 > 10.174.75.167.domain: 12503+ AAAA? www.google.com. (32)
06:59:20.063882 IP 10.174.75.167.domain > 9003533ae932.45025: 55528* 1/0/0 A 0.0.0.0 (49)
06:59:20.063896 IP 10.174.75.167.domain > 9003533ae932.45025: 55528* 1/0/0 A 0.0.0.0 (49)
06:59:25.064140 IP 9003533ae932.56854 > 10.174.75.168.domain: 12503+ AAAA? www.google.com. (32)
06:59:30.065619 IP 9003533ae932.45025 > 10.174.75.167.domain: 12503+ AAAA? www.google.com. (32)
06:59:40.067299 IP 9003533ae932.56854 > 10.174.75.168.domain: 12503+ AAAA? www.google.com. (32)
06:59:50.068468 IP 9003533ae932.45025 > 10.174.75.167.domain: 12503+ AAAA? www.google.com. (32)
07:00:10.069873 IP 9003533ae932.56854 > 10.174.75.168.domain: 12503+ AAAA? www.google.com. (32)
07:00:30.071284 IP 9003533ae932.45025 > 10.174.75.167.domain: 12503+ AAAA? www.google.com. (32)
07:01:10.072072 IP 9003533ae932.56854 > 10.174.75.168.domain: 12503+ AAAA? www.google.com. (32)
*   Trying 0.0.0.0:80...
* connect to 0.0.0.0 port 80 failed: Connection refused
* Failed to connect to www.google.com port 80 after 150020 ms: Couldn't connect to server
* Closing connection
curl: (7) Failed to connect to www.google.com port 80 after 150020 ms: Couldn't connect to server
root@9003533ae932:/# 

Here its not trying the ipv4 query with second server still it sending only ipv6 address query to server, after timeout curl will try to connect to the resolved ipv4 address and it will fail.

Could you please help me understand if this is a bug or if there's a way to resolve this issue when the resolved address is nothing from server ?

Thanks,
Sachin

bradh352:
Well, according to that, its not that you're receiving nothing, you're receiving an A record with an address of 0.0.0.0, which is considered a "black hole" address.  Often DNS servers will return this if they are configured as a split DNS, especially when it comes to things like VPN connections.   They may use it to intentionally to try to lock down traffic to only their supported domains while on VPN, or they may advertise a DNS domain suffix that they expect only those names to be routed to that particular DNS server and other names need to go to other DNS servers (please note that c-ares does NOT currently support this).

bradh352:
See PR #551

PR #551, opened 2023-09-25
ares_getaddrinfo(): Fail faster on AF_UNSPEC if we've already received one address class 

bradh352:
As per #541, when using AF_UNSPEC with ares_getaddrinfo() (and in turn with ares_gethostbynam()) if we receive a successful response for one address class, we should not allow the other address class to continue on with retries, just return the address class we have.

This will limit the overall query time to whatever timeout remains for the pending query for the other address class, it will not, however, terminate the other query as it may still prove to be successful (possibly coming in less than a millisecond later) and we'd want that result still.  It just turns off additional error processing to get the result back quicker.

### DIFF
--- a/src/lib/ares_private.h
+++ b/src/lib/ares_private.h
@@ -366,12 +366,20 @@ extern void (*ares_free)(void *ptr);
 int ares__timedout(struct timeval *now,
                    struct timeval *check);
 
-void ares__send_query(ares_channel channel, struct query *query,
-                      struct timeval *now);
-/* Identical to ares_query, but returns the query id */
-unsigned short ares_query_qid(ares_channel channel, const char *name,
+/* Returns one of the normal ares status codes like ARES_SUCCESS */
+int ares__send_query(ares_channel channel, struct query *query,
+                     struct timeval *now);
+
+/* Identical to ares_query, but returns a normal ares return code like
+ * ARES_SUCCESS, and can be passed the qid by reference which will be
+ * filled in on ARES_SUCCESS */
+int ares_query_qid(ares_channel channel, const char *name,
                               int dnsclass, int type, ares_callback callback,
-                              void *arg);
+                              void *arg, unsigned short *qid);
+/* Identical to ares_send() except returns normal ares return codes like
+ * ARES_SUCCESS */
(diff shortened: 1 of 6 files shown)


src/lib/ares_private.h
```diff
@@ -248,6 +248,8 @@ struct query {
   int using_tcp;
   int error_status;
   int timeouts; /* number of timeouts we saw for this request */
+  int no_retries; /* do not perform any additional retries, this is set when
```
bagder, line 251: Maybe make it `no_retries:1;` to signal that it actually is a boolean? Alternatively use a bool typedef.

src/lib/ares_private.h
```diff
@@ -364,8 +366,20 @@ extern void (*ares_free)(void *ptr);
...
+/* Identical to ares_query, but returns a normal ares return code like
+ * ARES_SUCCESS, and can be passed the qid by reference which will be
+ * filled in on ARES_SUCCESS */
+int ares_query_qid(ares_channel channel, const char *name,
+                              int dnsclass, int type, ares_callback callback,
+                              void *arg, unsigned short *qid);
+/* Identical to ares_send() except returns normal ares return codes like
+ * ARES_SUCCESS */
+int ares_send_ex(ares_channel channel, const unsigned char *qbuf, int qlen,
+                 ares_callback callback, void *arg);
```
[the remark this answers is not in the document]
bradh352: They're in the private header.  Not sure if they need to be public at all. I just needed to know if there was an error code to know if the query pointer was invalidated by the call or not as otherwise it would cause access to a free'd memory area.  In the future we could make them public and add docs for them, but for now they're private.

src/lib/ares_private.h
bradh352: This is another one where "convention" within c-ares gets in the way.  There is no `bool` type defined or used anywhere within c-ares, and I can't find any other instances of using the bitfield width modifiers.  You can see the "using_tcp" member above, it too is a boolean, however it too uses no modifiers.  I'd rather keep consistency at this point and maybe do a bulk "cleanup" in the future for things like this.

PR #564, opened 2023-10-08
ares-test: add typecast to silence warning

bagder:
warning: comparison of integer expressions of different signedness

### DIFF
--- a/test/ares-test.cc
+++ b/test/ares-test.cc
@@ -384,7 +384,7 @@ void MockServer::ProcessFD(int fd) {
     /* TCP might aggregate the various requests into a single packet, so we
      * need to split */
     while (tcp_data_len_ > 2) {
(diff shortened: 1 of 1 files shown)


test/ares-test.cc
```diff
@@ -385,7 +385,7 @@ void MockServer::ProcessFD(int fd) {
      * need to split */
     while (tcp_data_len_ > 2) {
       int tcplen = (tcp_data_[0] << 8) + tcp_data_[1];
-      if (tcp_data_len_ - 2 < tcplen)
+      if (tcp_data_len_ - 2 < (size_t)tcplen)
         break;
```
bradh352, line 389: maybe we should make tcplen size_t in the first place on line up?

bagder: yes, that's indeed nicer

PR #565, opened 2023-10-09
docs: provide better man page references

bagder:
When referring to another c-ares function use \fI function(3) \fP to let the webpage rendering find and cross-link them appropriately.

SEE ALSO references should be ".BR name (3),", with a space before the open parenthesis. This helps the manpage to HTML renderer.

### DIFF
--- a/docs/ares_parse_txt_reply.3
+++ b/docs/ares_parse_txt_reply.3
@@ -29,10 +29,8 @@ int ares_parse_txt_reply_ext(const unsigned char* \fIabuf\fP, int \fIalen\fP,
                              struct ares_txt_ext **\fItxt_out\fP);
 .fi
 .SH DESCRIPTION
-The
-.BR "ares_parse_txt_reply" " (" "ares_parse_txt_reply_ext" ")"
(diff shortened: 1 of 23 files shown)


docs/ares_parse_txt_reply.3
```diff
@@ -29,10 +29,8 @@ int ares_parse_txt_reply_ext(const unsigned char* \fIabuf\fP, int \fIalen\fP,
                              struct ares_txt_ext **\fItxt_out\fP);
 .fi
 .SH DESCRIPTION
-The
-.BR "ares_parse_txt_reply" " (" "ares_parse_txt_reply_ext" ")"
-function parses the response to a query of type TXT into a
-linked list (one element per sub-string) of
+The \fIares_parse_txt_reply(3)\fP function parses the response to a query of
+type TXT into a linked list (one element per sub-string) of
```
bradh352, line 33: looks like you removed reference to ares_parse_txt_reply_ext

bagder: oops

PR #567, opened 2023-10-10
ares_status_t enum for status codes

bradh352:
The list of possible error codes in c-ares was a #define list.  This not only doesn't provide for any sort of type safety but it also lacks clarification on what a function may return or what it takes, as an `int` could be an ares status, a boolean, or possibly even a length in the current code.

We are not changing any public APIs as though the C standard states the underlying size and type of an enum is `int`, there are compiler attributes to override this as well as compiler flags like `-fshort-enums`.  GCC in particular is known to expand an enum's width based on the data values (e.g., it can emit a 64bit integer enum).

All internal usages should be changed by this PR, but of course, there may be some I missed.

### DIFF
--- a/src/lib/ares__addrinfo_localhost.c
+++ b/src/lib/ares__addrinfo_localhost.c
@@ -215,7 +215,7 @@ ares_status_t ares__addrinfo_localhost(const char *name,
                                        struct ares_addrinfo *ai)
 {
   struct ares_addrinfo_node *nodes = NULL;
-  ares_status_t result;
+  ares_status_t status;
 
   /* Validate family */
   switch (hints->ai_family) {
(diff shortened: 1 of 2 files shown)


src/lib/ares__addrinfo_localhost.c
```diff
@@ -209,13 +209,13 @@ static int ares__system_loopback_addrs(int aftype,
...
-                             const struct ares_addrinfo_hints *hints,
-                             struct ares_addrinfo *ai)
+ares_status_t ares__addrinfo_localhost(const char *name,
+                                       unsigned short port,
+                                       const struct ares_addrinfo_hints *hints,
+                                       struct ares_addrinfo *ai)
 {
   struct ares_addrinfo_node *nodes = NULL;
-  int result;
+  ares_status_t result;
```
bagder, line 218: I think it would increase readability if we stick to the same name for the general `ares_status_t` return code variable and we use `status` already in most places. So maybe rename here?

src/lib/ares_process.c
```diff
@@ -783,7 +783,7 @@ int ares__send_query(ares_channel channel, struct query *query,
      * a send request.
      */
     if (server->tcp_conn == NULL) {
-      int err = open_socket(channel, server, 1);
+      ares_status_t err = open_socket(channel, server, 1);
       switch (err) {
```
bagder, line 787:
```suggestion
      status = open_socket(channel, server, 1);
      switch (status) {
```

src/lib/ares_process.c
```diff
@@ -836,7 +836,7 @@ int ares__send_query(ares_channel channel, struct query *query,
     }
 
     if (node == NULL) {
-      int err = open_socket(channel, server, 0);
+      ares_status_t err = open_socket(channel, server, 0);
       switch (err) {
```
bagder, line 840:
```suggestion
      status = open_socket(channel, server, 0);
      switch (status) {
```

bagder (approved):

gvanem:
What about `ares_set_sortlist()`? It still returns an `int`.

bradh352:
> What about `ares_set_sortlist()`etc.? Still many hits for `int ares_`.

I have intentionally not changed any public API interfaces as it could break ABI compatibility.  Compilers aren't guaranteed to internally use `int` for an `enum`, especially if someone is compiling their code using GCC and are specifying `-fshort-enums`.

PR #572, opened 2023-10-13
feat: use CMake to control symbol visibility

gjasny:
One of the shortcomings of the CMake build system is the inability to hide private symbols in shared libraries. That hinders adoption of the CMake build system in Debian or Ubuntu packages.

This PR uses the CMake-built-in [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) functionality to create an `ares_export.h` header containing the necessary compiler attributes. I tried to make this as less-intrusive as possible.

The other option would be to add a pretty minimal solution that only works with GCC and compatible compilers.


bradh352:
At first glance this seems reasonable.  Obviously adding another public header isn't optimal, especially one that will exist only based on the build system, but I don't think its a huge deal.  Any other caveats to this?

bradh352:
went with the other PR

PR #573, opened 2023-10-13
Fix internal datatype usage and warnings

bradh352:
PR #568 increased the warning levels and c-ares code emitted a bunch of warnings.  This PR fixes those warnings and starts transitioning internal data types into more proper forms (e.g. data lengths should be size_t not int).  It does, however, have to manually cast back to what the public API needs due to API and ABI compliance (we aren't looking to break integrations, just clean up internals).


bradh352:
i think i've got a vast majority of the data type usages now

PR #574, opened 2023-10-13
feat: use CMake to control symbol visibility (w/o extra header)

gjasny:
In contrast to #572 this solution does not need any extra headers. But it is also limited to GCC-like compilers.


bradh352:
Lets go this direction I think.  Should work on at least GCC and Clang. Afaik, Windows symbols are already hidden unless explicitly exported with MSVC.  There's just not *that* many other compiler solutions these days.

PR #578, opened 2023-10-15
clang-format

bradh352:
Create a clang-format style and run it through all code.


src/lib/ares_cancel.c
```diff
@@ -37,17 +37,16 @@
...
     ares__llist_node_t *next = NULL;
 
     /* Swap list heads, so that only those queries which were present on entry
      * into this function are cancelled. New queries added by callbacks of
      * queries being cancelled will not be cancelled themselves.
      */
-    ares__llist_t *list_copy = channel->all_queries;
-    channel->all_queries = ares__llist_create(NULL);
+    ares__llist_t      *list_copy = channel->all_queries;
+    channel->all_queries          = ares__llist_create(NULL);
```
bagder, line 49: Wait, why is this padding the code with spaces here? Is this just because it makes an assign below the variable declaration with an assign?

src/lib/ares_cancel.c
```diff
@@ -64,20 +63,22 @@ void ares_cancel(ares_channel channel)
       /* Cache next since this node is being deleted */
       next = ares__llist_node_next(node);
 
-      query = ares__llist_node_claim(node);
+      query                   = ares__llist_node_claim(node);
```
bagder, line 66: and here, I suppose it wants to ... maintain the alignment from above? I find it a peculiar way to format code.

src/lib/ares_expand_name.c
```diff
@@ -325,12 +321,13 @@ static ares_ssize_t name_length(const unsigned char *encoded,
...
                                              size_t alen, char **s,
-                                             size_t *enclen,
+                                             size_t     *enclen,
                                              ares_bool_t is_hostname)
 {
-  ares_status_t status = ares__expand_name_validated(encoded, abuf, alen, s,
-                                                     enclen, is_hostname);
-  if (status == ARES_EBADNAME)
+  ares_status_t status =
+      ares__expand_name_validated(encoded, abuf, alen, s, enclen, is_hostname);
```
bagder, line 328: This indents *four* spaces all of a sudden. Shouldn't this be indented by two spaces?

src/lib/ares_init.c
```diff
@@ -193,66 +195,76 @@ int ares_init_options(ares_channel *channelptr, struct ares_options *options,
...
+    }
+    if (channel->lookups) {
+      ares_free(channel->lookups);
+    }
+    if (channel->resolvconf_path) {
+      ares_free(channel->resolvconf_path);
+    }
+    if (channel->hosts_path) {
+      ares_free(channel->hosts_path);
+    }
```
bagder, line 255: ares_free() handles NULL by itself, maybe it is time to stop doing the superfluous NULL checks?

src/lib/ares_init.c
```diff
@@ -1659,32 +1706,31 @@ static ares_status_t init_by_defaults(ares_channel channel)
...
     }
 
     do {
       res = gethostname(hostname, lenv);
 
-      if(toolong(res)) {
+      if (toolong(res)) {
         char *p;
-        len *= 2;
+        len  *= 2;
```
bagder, line 1724: really?

bagder (commented):
I can live with the format choices even if I don't love it all. But unless there are also tools added to enforce the style, it seems hard to maintain. Especially the tricky alignments.

src/lib/ares_cancel.c
bradh352: yeah, looks like it. I think this is because there isn't a newline between the variable declaration and the assignment on the next line, as it won't do that across newlines.  

There are definitely some areas where its not flexible enough.

src/lib/ares_cancel.c
bradh352: actually, its aligning with the line below it.  But not with the line above it because I have it set not to span across lines.

src/lib/ares_init.c
bradh352: yep, aligning with the line below.

PR #590, opened 2023-10-27
ares_getnameinfo: Add missing netioapi.h include

jonaski:
Fixes mingw-w64 build error

ares_getnameinfo.c:357:9: error: implicit declaration of function 'if_indextoname' [-Werror=implicit-function-declaration]


bradh352:
interesting, docs says to include `Iphlpapi.h`:
https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-if_indextoname

Interesting that our CI/CD system didn't catch this though, doesn't seem to be a regression, so I'm assuming this is something about a newer MinGW ... we probably need to update our CI/CD to include a new one I'm guessing.

jonaski:
I'm using mingw-w64 11.0.1

jonaski:
[cares_x86_64-w64-mingw32.shared.txt](https://github.com/c-ares/c-ares/files/13194020/cares_x86_64-w64-mingw32.shared.txt)

gvanem:
MSDN also says: *The Ntddndis.h, Ifdef.h, and Netioapi.h header files should never be used directly.*

Wouldn't it be better to define `_WIN32_WINNT >= 0x0602` since `iphlpapi.h` has:
```c
#if (NTDDI_VERSION >= NTDDI_VISTA)

#include <netioapi.h>

#endif // (NTDDI_VERSION >= NTDDI_VISTA)
```

@jonaski I fail to see from your build-log what value (if any) `_WIN32_WINNT` has.

bradh352:
Just change the include to `Iphlpapi.h` since that's what the docs say and I'll merge it.

I'll also need to update our CI/CD system to use a newer mingw version so I can tackle some of those warnings.

jonaski:
The include needs to be lowercase to allow cross-compilation on Linux as the header provided by mingw-w64 is.
As for _WIN32_WINNT, I was using the default in mingw-w64 (0x0600), I bumped to 0x0A00, but got the same error.

bradh352:
Looks like there's a reason we're not seeing it elsewhere, the CMAKE build system wasn't properly detecting if_indextoname().  Look at PR #593 where I'm redoing the CI/CD system with appveyor to update the compilers, it incorporates your fix and fixes the Cmake build system.

Issue #550, opened 2023-09-25
Bad servers should be de-prioritized


bradh352:
Right now, if one query hits a bad server (such as one that times out), each future query will also hit that same bad server and have to decide to skip it too which will cause really bad system performance.  We should track consecutive failures on a server and sort the server list by this metric.  Its possible we may need to also timeout such a state as a sysadmin might configure servers further away (and thus have higher latency) later in the list so every once in a while assuming the server is still down, a query or two might re-hit that bad server even if there's a known good server, just to check the state, but its not clear what time window might make the most sense for that (5mins?  1hr?).

To implement this, the server list would need to be implemented using an `ares__slist_t`, and any tracking based on server indexes would need to be nixed.  However doing that would make it possible to implement #301 which is otherwise not possible right now due to the way servers are tracked.

bradh352:
Fixed in PR#594

Issue #440, opened 2021-11-10
Are open DNS queries still valid after changing the DNS servers?


ChristianAmmer-zz:
Assume that a communication channel for name service lookups was initialized with `ares_init` and name lookups were performed by `ares_getaddrinfo` or `ares_gethostbyname`, but these lookups are still open (were not answered or timed out).

Is it then possible to change the name servers for the same communication channel (using `ares_set_servers`) and still get a successful name lookup for the open queries? Or will the name lookups with the previous name servers be canceled?

bradh352:
I'm pretty sure that would be undefined behavior.  c-ares doesn't have any way to update the configuration at this time.  I know in my own private use of c-ares I keep a list of ares_channel pointers along with an active query count per pointer.  All new queries go to the most recent ares_channel in the list, and when the query count hits zero on any pointer that isn't the most recent, I destroy the channel.  So when I need to rescan for a config changes, I just create a new channel and make it the most recent in the list.

ChristianAmmer:
Thank you for the description of how to achieve this goal.
Should the effects (when calling `ares_set_servers`) on the open name lookups be documented?

bradh352:
Probably.  I'm not very much in love with the c-ares API in general.  Too many public structs, no way to cancel a single query, no way to reload system config on network changes...  I'm half-way tempted to create a whole new API and make the old API wrappers around the new.

MyLibh:
> no way to reload system config on network changes.

Does this problem still exist? If the list of DNS servers changes, you need to re-create the channel?

bradh352:
it does still exist, but this might be resolved in the next release as it is on my project list.  A lot of the framework to facilitate this was added in the last release.

MyLibh:
> it does still exist, but this might be resolved in the next release as it is on my project list. A lot of the framework to facilitate this was added in the last release.

So, speaking about older versions, when changing the NS list or network type, the channel should be recreated?

bradh352:
correct

dimbleby:
#191 disallowed this some time ago

bradh352:
The docs did previously mention it would return ARES_ENOTIMP in such a case.  But as of PR #594, this is fully supported.

PR #594, opened 2023-11-02
Dynamic Server List

bradh352:
This PR makes the server list a dynamic sorted list of servers.  The sort order is [ consecutive failures, system config index ].  The server list can be updated via `ares_set_servers_*()`.  Any queries currently directed to servers that are no longer in the list will be automatically re-queued to a different server.

Also, any time a failure occurs on the server, the sort order of the servers will be updated so that the one with the fewest consecutive failures is chosen for the next query that goes on the wire, this way bad or non-responsive servers are automatically isolated. 

Since the server list is now dynamic, the tracking of query failures per server has been removed and instead is relying on the server sort order as previously described.  This simplifies the logic while also reducing the amount of memory required per query.  However, because of this dynamic nature, it may not be easy to determine the server attempt order for enqueued queries if there have been any failures.

If using the `ARES_OPT_ROTATE`, this is now implemented to be a random selection of the configured servers.  Since the server list is dynamic, its not possible to go to the next server as configuration could have changed between queries or attempts for the same query.

Finally, this PR moved some existing functions into new files to logically separate them.

This should address issues #550 and #440, while also setting the framework to implement #301.  #301 needs a little more effort since it configures things other than the servers themselves (domains, search, sortlist, lookups), which need to make sure they can be safely updated.


bradh352:
alright, after that last commit, think things should build fully

PR #601, opened 2023-11-08
Update public prototypes with `const` where appropriate

bradh352:
Some existing public prototypes do not modify the provided parameters and therefore should mark such parameters as `const`.  This is considered a best practice as per MISRA-C, and will not impact API nor ABI compatibility.  No existing code should emit warnings due to this.

There are some other changes that _should_ be done, and would maintain ABI compatibility, but would cause warnings to be emitted (such as the callback structures for things like `ares_gethostbyname`, where `hostent` should be marked as `const`), this has not been done.

### DIFF
--- a/docs/ares_dup.3
+++ b/docs/ares_dup.3
@@ -22,7 +22,7 @@ ares_dup \- Duplicate a resolver channel
 .nf
 #include <ares.h>
 
-int ares_dup(ares_channel_t **\fIdest\fP, ares_channel_t *\fIsource\fP)
+int ares_dup(ares_channel_t **\fIdest\fP, const ares_channel_t *\fIsource\fP)
 .fi
 .SH DESCRIPTION
 The \fBares_dup(3)\fP function duplicates an existing communications channel
--- a/docs/ares_fds.3
+++ b/docs/ares_fds.3
@@ -22,9 +22,9 @@ ares_fds \- return file descriptors to select on
 .nf
 #include <ares.h>
 
-int ares_fds(ares_channel_t *\fIchannel\fP,
+int ares_fds(const ares_channel_t *\fIchannel\fP,
              fd_set *\fIread_fds\fP,
-	     fd_set *\fIwrite_fds\fP)
+             fd_set *\fIwrite_fds\fP)
 .fi
 .SH DESCRIPTION
 The \fBares_fds(3)\fP function retrieves the set of file descriptors which the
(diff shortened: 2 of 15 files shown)


bagder:
Should the docs maybe mention that this is changed in version X.Y? I'm sure there will be applications who want to be able to build with a mix of c-ares versions and knowing that arguments were not `const` until a particular version could then be useful input.

bradh352:
> Should the docs maybe mention that this is changed in version X.Y? I'm sure there will be applications who want to be able to build with a mix of c-ares versions and knowing that arguments were not `const` until a particular version could then be useful input.

The only case I can think of is wrappers, like the Rust guys needing to know a minimum version they can use the official c-ares instead of the fork.  If you think its important enough to throw into the docs for such a minimal use-case, we can definitely do so, but honestly its probably just clutter.

bradh352:
@dimbleby @bagder ...

I just thought of a reason to NOT do this, at least for `ares_channel_t`.  I was considering adding threading to c-ares, which would mean there would be a mutex within the ares channel, so even a read operation would need to lock the mutex to ensure thread safety.

You might ask "why threading"?  Well:

1. People seem to misuse c-ares already, having a mutex protect the channel would prevent subtle issues in the c-ares use.
2. I'm going to add `ares_reinit()` here shortly, likely people will register that as part of their callback within the OS when they are notified of network changes, and its very possible this may be in a different thread.
3. I was considering adding a new ares option for explicitly spawning a thread for handling the communication loop so the integrator doesn't have to manage this themselves.

dimbleby:
putting a mutex inside the c-ares channel would be a big change!  Does it even work?  What if one thread is calling ares_destroy(channel) while another thread is calling ares_do_something(channel), and the destroy thread wins the race?

"The Rust guys" is just me!  From my point of view it's really only pedantry and perfectionism that wants to tell the difference between "functions that require single-threaded access" and "functions that don't".  (But the Rust advantage is that so long as I get that right, the compiler will prevent users from making the mistake that you talk about).  I could mark everything as requires-single-threaded-access without losing very much: or if the API magically became thread-safe then I expect I could handle that too.

Which is to say that much as in the other places I've commented on this: what you have here looks good to me right now, but if you're afraid of it for some reason then I'm not going to lose much sleep over it not being done.

bradh352:
There's always going to be the "init" and "destroy" issue even when adding thread safety, that's true for any library in C. Its everything else in-between that you're protecting.  And the fact that with an `ares_reinit()` implementation, there will likely be exactly 1 `ares_channel_t` for the life of the application, it shouldn't be hard to ensure.  Previously people would have to destroy the entire channel and reinitialize it based on server state changes (and possibly lose queries due to that).

Anyhow, I don't want to prevent the threading change I just mentioned, so I think i'm going to close out the PR.

PR #606, opened 2023-11-10
Support jitter and maxtimeout for retry penalties

Kontakter:
We use c-ares for DNS resolving at the project https://github.com/YTsaurus/YTsaurus 

In our installations we face a problem with synchronized request to the DNS from hundreds of thousands of processes and I want to add jitter for retry backoffs to smooth these requests over time.

### DIFF
--- a/docs/ares_init_options.3
+++ b/docs/ares_init_options.3
@@ -28,6 +28,8 @@ struct ares_options {
   int timeout; /* in seconds or milliseconds, depending on options */
   int tries;
   int ndots;
+  int maxtimeout; /* in milliseconds */
+  int jitter; /* in 0.001 */
   unsigned short udp_port;
   unsigned short tcp_port;
   int socket_send_buffer_size;
@@ -101,6 +103,19 @@ The number of dots which must be present in a domain name for it to be
 queried for "as is" prior to querying for it with the default domain
 extensions appended.  The default value is 1 unless set otherwise by
 resolv.conf or the RES_OPTIONS environment variable.
+.TP 18
+.B ARES_OPT_MAXTIMEOUTMS
+.B int \fImaxtimeout\fP;
+.br
+The upper bound for timeout between sequential retry attempts.
+.TP 18
+.B ARES_OPT_JITTER
+.B int \fIjitter\fP;
+.br
+This option allows to add random noise for timeout between sequential retry attempts.
+It is used to smooth over time DNS requests from multiple processes. The value added
+to the retry timeout is choosen randomly from the range [-timeout * jitter / 1000.0,
+timeout * jitter / 1000.0]. The default value is 0.
 .TP 18
 .B ARES_OPT_UDP_PORT
 .B unsigned short \fIudp_port\fP;
--- a/docs/ares_init_options.3
+++ b/docs/ares_init_options.3
@@ -29,7 +29,6 @@ struct ares_options {
   int tries;
   int ndots;
   int maxtimeout; /* in milliseconds */
-  int jitter; /* in 0.001 */
   unsigned short udp_port;
   unsigned short tcp_port;
   int socket_send_buffer_size;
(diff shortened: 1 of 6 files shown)


bradh352:
A few things here:

1. on FreeBSD you'll need to include stdint.h to get SIZE_MAX (just make sure you have it in `#ifdef HAVE_STDINT_H` guards)
2. The OptionsChannelInit test is failing, looks like in ares_save_options() you used the wrong variable name in your if() statement, should use channel->optmask
3. Need to update docs/ares_init_options.3 for the new params
4. Can you provide more insight into the algorithm used here and how it helps?  I'm not familiar at all with jitter type algorithms, and the code really doesn't discuss anything.

Kontakter:
Thank you for fast reply!

I have fixed 1, 2 and 3. 

About 4 – new code just add some noise to the retry backoff. Value of the jitter regulates amplitude of this noise. The code could look tricky since I am trying to avoid overflows of timeplus after adding this noise.

bradh352:
Whats the purpose of the backoff "noise"? I'm just trying to understand the issue its resolving.  It is just adjusting timeouts for how long it waits for a response when the first round across all servers, since it didn't get any responses from any servers in the configured timeout interval, so it assumes that might mean we aren't waiting long enough.

Kontakter:
The issue is following, I start large cluster (or start large distributed computation) with 10'000+ processes, all of these processes at the start want to resolve addresses of some control hosts (or of some source hosts) and goes to the DNS server. In such case this server can start throttle resolve requests and we want to avoid any synchronisation of retries (that really happens in such kind of situations)

bradh352:
Interesting.  Is there a true need for this to be configurable?  Or any reason it shouldn't be always-on?  If we could come up with parameters that should be _acceptable_ for all use cases, my preference would be to not add more config options (one because its more to maintain, but two because other people could benefit from this but may not have realized there's a config to fix their usecase).

I'd think as long as `ares__retry_penalty()` returns something in the range of `channel->timeout` to `channel->timeout * (ntries / servercnt)`, I'd be fine with it.  Even better if it is more likely to return a larger number as `(ntries / servercnt)` increases.

Kontakter:
"Is there a true need for this to be configurable?"  – it is a good question!

In our project we actually use fixed value of this jitter for years. But my initial will was to remain unchanged the default behaviour of such a low-level library as c-ares, because my experience tells that engineers often rely on very concrete aspects of behaviour.

If we ok with changes of default behaviour, then I can suggest following:
  * I saves current behaviour with exponential backoffs that library has right now
  * I add `maxtimeout` parameters (it is a separate problem, but we actually want to have some upper bound)
  * I drop jitter parament and add randomness to pick the value from range `[channel->timeout * (ntries / servercnt) * 0.5, channel->timeout * (ntries / servercnt)]`

Are you ok with that?

bradh352:
I think that sounds reasonable to me :)

bradh352:
Looking at this, I think `ares__retry_penalty()` is misnamed as it takes the value directly from this function to set the overall query timeout (not just adding a penalty on retries). It should actually be `ares__calc_query_timeout()` I think, so it needs to return a _minimum_ of `channel->timeout`, and a maximum of `channel->timeoutmax` (if set), as this function is also called on even the *first* attempt.   Infact, this jitter should only be run if `shift != 0` as 0 is the first pass.

Kontakter:
> Looking at this, I think `ares__retry_penalty()` is misnamed as it takes the value directly from this function to set the overall query timeout (not just adding a penalty on retries). It should actually be `ares__calc_query_timeout()` I think, so it needs to return a _minimum_ of `channel->timeout`, and a maximum of `channel->timeoutmax` (if set), as this function is also called on even the _first_ attempt. Infact, this jitter should only be run if `shift != 0` as 0 is the first pass.

I have fixed it, thank you for the remark

bradh352:
not seeing another commit after my last comment, forget to push it ?

Kontakter:
> not seeing another commit after my last comment, forget to push it ?

Oops, fixed

bradh352:
can you look at https://github.com/c-ares/c-ares/commit/4acd5759e99e611bbf41afe2fa4303d3c893cdde to make sure I didn't negatively impact your logic? Just some fixes I noticed.

PR #615, opened 2023-11-13
Add DNS record manpages

bradh352:
The new DNS record parser and writer needs manpages.  This PR implements those.


bradh352:
can you look at 4acd575 to make sure I didn't negatively impact your logic?  Just some fixes I noticed.

PR #619, opened 2023-11-14
Fix typos and man page whatis entry

gjasny:
Those issues were detected by lintian.

### DIFF
--- a/docs/ares_dns_mapping.3
+++ b/docs/ares_dns_mapping.3
@@ -3,6 +3,11 @@
 .\"
 .TH ARES_DNS_MAPPINGS 3 "12 November 2023"
 .SH NAME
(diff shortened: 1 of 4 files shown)


gjasny:
Did you really mean "Assertation" (not Assertion) here?
https://github.com/c-ares/c-ares/blob/756b493e7678b74ff5f168271c88019fd21f6b3d/docs/ares_dns_rr.3#L289

bradh352:
> Did you really mean "Assertation" (not Assertion) here?
> 
> https://github.com/c-ares/c-ares/blob/756b493e7678b74ff5f168271c88019fd21f6b3d/docs/ares_dns_rr.3#L289

You are right, its supposed to be Assertion

gjasny:
> You are right, its supposed to be Assertion

Thanks. I updated the PR

PR #625, opened 2023-11-18
Query Cache support

### LINKED ISSUE #608: Feature: Add optional caching mechanism based on TTL
We should be able to optionally cache any request/response pairs up until the maximum TTL advertised in any returned record by the server and return the cached results if the minimum TTL has not yet expired.  The request cache check needs to validate all parameters, including flags (such as if one request had recursion desired and the other didn't, they could lead to different results).

Having a cache could significantly improve performance in some applications, especially if the DNS servers used are slow or far away.

This would be enabled via an ares_init_options() flag, likely we should allow specification of a maximum TTL to honor to prevent DNS propagation issues (3600 is probably a safe choice, but since its only a cache, even 300 would help). 

Anything read from files wouldn't necessarily need to be cached since there is already a built-in cache for that.  However, if we did want to extend for file lookups, it may make sense to use a TTL of 1 to prevent back to back stat()'s on the hosts file ... but really, not sure if this is worthwhile at all.

We should also cache fatal errors returned from a server, such as NXDOMAIN, if an SOA record is returned that we can use to obtain the negative response cache timeframe.

Finally, any ares_reinit() call should clear the cache automatically as if there was a configuration issue, we don't want bad results to remain in cache.

bradh352:
This PR implements a query cache at the lowest possible level, the actual dns request and response messages.  Only successful and `NXDOMAIN` responses are cached. The lowest TTL in the response message determines the cache validity period for the response, and is capped at the configuration value for `qcache_max_ttl`.  For `NXDOMAIN` responses, the SOA record is evaluated.

For a query to match the cache, the opcode, flags, and each question's class, type, and name are all evaluated.  This is to prevent matching a cached entry for a subtly different query (such as if the RD flag is set on one request and not another).

For things like ares_getaddrinfo() or ares_search() that may spawn multiple queries, each individual message received is cached rather than the overarching response.  This makes it possible for one query in the sequence to be purged from the cache while others still return cached results which means there is no chance of ever returning stale data.

We have had a lot of user requests to return TTLs on all the various parsers like `ares_parse_caa_reply()`, and likely this is because they want to implement caching mechanisms of their own, thus this PR should solve those issues as well.

Due to the internal data structures we have these days, this PR is less than 500 lines of new code.

Fixes #608 


bradh352:
need to write a test case that better validates this, but otherwise i think its good.

PR #626, opened 2023-11-18
Fix winuwp build

halx99:


bradh352:
This broke openwatcom due to including netioapi.h, what functions are missing that made you include that?

On the hosts file, we can probably kill the WIN9X statement all together as we removed other areas still supporting Win9x a long while ago.

bradh352:
I'm trying to add UWP building support to our CI/CD system in #627 ... No idea if this is a valid way to do it or not.  I can say the tools won't run, so I'm guessing they're not linked properly so there may be some other issue in the way its building.  I'm not seeing any errors like this PR is trying to address.

Perhaps I need to run a different cmake generator, like "Visual Studio 17 2022" and then build the project with msbuild?  I also see references to maybe needing to create manifests that get linked in, not sure if that's true.  Finally, I also see something about `VS_WINRT_COMPONENT TRUE` as a target property maybe that needs to be set.

There references saying only C++ apps can be console as well, can't tell if that's just an oversight an C apps can't be console or what.

Any insight into how you're building c-ares for UWP would be helpful.

halx99:
> I'm trying to add UWP building support to our CI/CD system in #627 ... No idea if this is a valid way to do it or not. I can say the tools won't run, so I'm guessing they're not linked properly so there may be some other issue in the way its building. I'm not seeing any errors like this PR is trying to address.
> 
> Perhaps I need to run a different cmake generator, like "Visual Studio 17 2022" and then build the project with msbuild? I also see references to maybe needing to create manifests that get linked in, not sure if that's true. Finally, I also see something about `VS_WINRT_COMPONENT TRUE` as a target property maybe that needs to be set.
> 
> There references saying only C++ apps can be console as well, can't tell if that's just an oversight an C apps can't be console or what.
> 
> Any insight into how you're building c-ares for UWP would be helpful.

refer to: https://github.com/halx99/c-ares/actions/runs/6917687282/job/18819143871

bradh352:
Does 'adig' or 'ahost' run for you on this build?  It fails for me.

bradh352:
I did this a slightly different way in PR #627 and gave you attribution.

PR #627, opened 2023-11-18
Fix Windows UWP (Store) building and add to CI/CD

bradh352:
When building for UWP (WindowsStore), additional headers are needed and some functions are not available.  This also adds AppVeyor CI/CD support to catch these issues in the future.

Fix By: Deal (@halx99) and Brad House (@bradh352)


lazka:
random note: this also fixes the cygwin build, which was broken by fb361064eb23474

PR #628, opened 2023-11-19
Try enable winuwp build by github action

halx99:


bradh352:
I did this a slightly different way in PR #627 and gave you attribution.  

That said, the tests and tools won't work at all.  I know nothing about UWP, and online docs seem to indicate only C++ apps will run?  If you have any insight into that aspect, that would be appreciated.

PR #632, opened 2023-11-21
Support attempts and timeout options from resolv.conf

Kontakter:
c-ares parses only antique version of options for timeout and number of retries from resolv.conf (`retrans` and `retry` are missing in modern documentation https://man7.org/linux/man-pages/man5/resolv.conf.5.html).

I add support of `attempts` and `timeout` options


Kontakter:
I cannot realise reasons of failures on Darwin, I failed to reproduce it on my MacBook

bradh352:
I think you may have found a bug in c-ares here as to why the macOS tests are failing.  The environment reading should come *after* sysconfig reading here: https://github.com/c-ares/c-ares/blob/main/src/lib/ares_sysconfig.c#L998 as the environment is supposed to act as overrides.

PR #635, opened 2023-11-24
CI: codespell

bagder:
Inspired by #634


cclauss:
% `codespell --ignore-words-list="aas,aci,acter,atleast,contentss,firey,fo,sais,seh,statics" ` # plus ignore gmock and m4

bagder:
As you can see, this runs on src and include, not the tests nor m4.

Also, it uses a file for words to ignore.

I want to let 634 merge first, then rebase this and make sure it runs green.

Issue #610, opened 2023-11-11
Feature: add basic thread safety


bradh352:
A lot of users get confused as to if c-ares functions can be called from different threads.  In general, that's a hard no.  However just adding a simple recursive mutex to `ares_channel_t` and making any public functions lock on entry and unlock on exit would easily eliminate many of the multi-threaded concerns.

One non-obvious aspect, however, is that if someone were to call `ares_getaddrinfo()` from another thread while their event loop is waiting on filedescriptors returned by `ares_fds()` or similar, they may hang forever since that other event loop won't be woken up to add a new file descriptor.  Users will need to know about this and know they need to wake their event loop up if they plan on using it this way.

Issue #611, opened 2023-11-11
Feature: add eventloop thread


bradh352:
This sort of expands on (and definitely depends on) #610. 

It probably makes sense to have an ares_init_options() flag that will automatically spawn a thread specifically for handling any events on file descriptors used by c-ares.  A lot of developers may want an asynchronous DNS resolver but not want to worry about the low-level behind-the-scenes things needed to use c-ares.

This thread would need to be able to be woken up any time a new query is enqueued (or at least any time a new file descriptor is opened that needs to be monitored).  On Linux we can use eventfd() and add it to our poll() loop, other Unix systems can use pipe() for this (and maybe we should on Linux too just so we don't have more codepaths).  Windows is a bit trickier, as we have to use undocumented internals ... reference https://github.com/piscisaureus/wepoll

Obviously not all systems may have threading so this may cause ARES_ENOTIMP to be returned to ares_init_options() if the system doesn't support threads and this was requested.

PR #636, opened 2023-11-26
Basic Thread Safety

bradh352:
c-ares does not have any concept of thread-safety.  It has always been 100% up to the implementor to ensure they never call c-ares from more than one thread at a time.  This patch adds basic thread-safety support, which can be disabled at compile time if not desired.  It uses a single recursive mutex per channel, which should be extremely quick when uncontested so overhead should be minimal.

Fixes Bug: #610 

Also sets the stage to implement #611


bradh352:
marked as draft as I still need to update docs and add an ares_threadsafe() function to query if the library was built with thread safety.

docs/ares_threadsafety.3
```diff
@@ -0,0 +1,37 @@
...
+.SH SYNOPSIS
+.nf
+#include <ares.h>
+
+ares_bool_t ares_threadsafety(void);
+.fi
+.SH DESCRIPTION
+The \fBares_threadsafety(3)\fP function returns if the library was built with
+thread safety enabled or not.
+
```
bagder, line 16: Maybe this description could also clarify somewhat what it means and what you can do with the library when it is thread-safe vs not thread-safe?

bagder (approved):

Issue #637, opened 2023-11-27
ARES_FLAG_NOSEARCH is ignored when using ares_getaddrinfo()


petrvh:
c-ares init options defines a flag ARES_FLAG_NOSEARCH that is supposed to prevent search using configured domain suffixes.

however when using ares_getaddrinfo() the flag was ignored and domain suffixes were used anyway.

Unfortunately explicitly configuring zero domains to search also does not work (if ndomains == 0 default domain search list is loaded regardless of the flag ARES_OPT_DOMAINS being set).

It seems that a simple way to fix this is to check for the flag in the function (Tested that this works)

static ares_bool_t as_is_only(const struct host_query *hquery)

PR #638, opened 2023-11-27
ares_getaddrinfo(): do not use search domains if ARES_FLAG_NOSEARCH is set

petrvh:
c-ares init options defines a flag ARES_FLAG_NOSEARCH that is supposed to prevent search using configured domain suffixes.

however when using ares_getaddrinfo() the flag was ignored and domain suffixes were used anyway.

Unfortunately explicitly configuring zero domains to search also does not work (if ndomains == 0 default domain search list is loaded regardless of the flag ARES_OPT_DOMAINS being set).

this change adds a check for the ARES_FLAG_NOSEARCH in as_is_only() function that is used by ares_getaddrinfo() to decide if to try to query next possible name ( next_dns_lookup() )

### DIFF
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -743,7 +743,7 @@ static ares_bool_t as_is_first(const struct host_query *hquery)
       ndots++;
     }
   }
-  if (hquery->name != NULL && nname && hquery->name[nname - 1] == '.') {
+  if (as_is_only(hquery)) {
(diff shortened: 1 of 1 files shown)


bradh352:
At first glance this seems to make sense.  My guess is this probably broke in the 1.16 timeframe when getaddrinfo was introduced and gethostbyname was made as a wrapper around getaddrinfo, so I think getaddrinfo has never supported this.

petrvh:
https://github.com/c-ares/c-ares/issues/637

PR #646, opened 2023-12-05
Support ipv6 link-local servers and %iface syntax

bradh352:
Some environments may send router advertisements on a link setting their link-local (fe80::/10) address as a valid DNS server to the remote system.  This will cause a DNS entry to be created like `fe80::1%iface`, since all link-local network interfaces are technically part of the same /10 subnet, it must be told what interface to send packets through explicitly if there are multiple physical interfaces.

This PR adds support for the %iface modifier when setting DNS servers via `/etc/resolv.conf` as well as via `ares_set_servers_csv()`. 

For MacOS and iOS it is assumed that libresolve will set the `sin6_scope_id` and should be supported, but my test systems don't seem to read the Router Advertisement for RDNSS link-local.  Specifying the link-local dns server on MacOS via adig has been tested and confirmed working.

For Windows, this is similar to MacOS in that the system doesn't seem to honor the RDNSS RA, but specifying manually has been tested to work.

At this point, Android support does not exist.

Fixes Bug #462
Supersedes PR #463

Fix By: Brad House (@bradh352) and Serhii Purik (@sergvpurik)


bradh352:
Tested on Ubuntu 22.04 after disabling systemd-resolved with another ubuntu host using radvd to advertise its link local address as a DNS server.  Works as intended.

src/lib/ares__socket.c
```diff
@@ -213,16 +213,10 @@ static int configure_socket(ares_socket_t s, struct server_state *server)
...
-    local.sa6.sin6_family = AF_INET6;
-#ifdef HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID
-    local.sa6.sin6_scope_id = server->ll_scope;
-#endif
-    bindlen = sizeof(local.sa6);
-  } else if (server->addr.family == AF_INET6 &&
+  } else if (server->addr.family == AF_INET6 && server->ll_scope == 0 &&
              memcmp(channel->local_ip6, ares_in6addr_any._S6_un._S6_u8,
                     sizeof(channel->local_ip6)) != 0) {
+    /* Only if not link-local */
```
sergvpurik, line 219: Is this a correct comment?

Looks like 
``` C
memcmp(channel->local_ip6, ares_in6addr_any._S6_un._S6_u8,
                    sizeof(channel->local_ip6)) != 0
```
means "channel->local_ip6 != 0::0"

PR #658, opened 2023-12-13
Add more ServerNoResponseFailover tests

apolukhin:


bradh352:
its not clear to me what this accomplishes over the tests I added previously.  The tests I added previously had strict checks on number of times a request would be called, so there's no need to set the tries = 1.  Also, ares_gethostbyname internally just calls ares_getaddrinfo so we shouldn't need any additional coverage in the mock-ai routines, most of those were technically written before that merge took place.

apolukhin:
This is my poor attempt to locate some issues. It failed - all the tests pass well. Feel free to close the PR.

Looks like https://github.com/c-ares/c-ares/commit/05181a6a798a649d6cc697c27233c03be7a24dea was one of the issues that affects us. Could you please add some tests for it?

Our failing tests still have the same shape: do some requests, shut down the old server and start a new one on different port, make sure that it resolves. We use 1 network attempt with big timeout as we have a cache over DNS requests, that deduplicates requests and allows them to finish in background. Changing network attempts to 3 or 5 does not help to pass the tests.

Could there be more issues with not updating statistics on server errors? If a DNS server stopped responding after a hundred of successful requests, how fast the c-ares would stop sending new requests to that server?

bradh352:
c-ares would have to timeout on a server, or receive a hard error from a server, before it would stop sending requests to that server.  You really should have a *short* timeout.  Unless you're going over something like a satellite link, 500ms would even be a *long* time, the default is 2000ms.  This is also why we have retries set, as the next attempt on the same server will automatically double the attempt time in case that original timeout really was set too low.  So really, something like 500ms x 3 retries would probably be where I'd go.

PR #659, opened 2023-12-13
Use SOCK_DNS extension on socket on OpenBSD

marty1885:
This patch added the SOCK_DNS flag when running on OpenBSD. Allowing a reduced set of pledge(2) promises. Before this patch. The "stdio rpath inet" promises must be used in order to resolve any records. After the patch. `inet` can be replaced with `dns` which only allows communication on destination port 53, instead of on all ports.

Side note: I checked the OpenBSD kernel source code. Even though the socket document says `the DNS port (typically 53).`, The OpenBSD 7.4 kernel only allows 53. Not sure why they say typically.

### DIFF
--- a/src/lib/ares__socket.c
+++ b/src/lib/ares__socket.c
@@ -253,6 +253,11 @@ ares_status_t ares__open_connection(ares_channel_t      *channel,
   struct server_connection *conn;
   ares__llist_node_t       *node;
   int                       type = is_tcp ? SOCK_STREAM : SOCK_DGRAM;
+#ifdef __OpenBSD__
+  if(!is_tcp && server->udp_port == 53) {
+    type |= SOCK_DNS;
+  }
+#endif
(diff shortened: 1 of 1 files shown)


bradh352:
Interesting.  That said, the manpage, https://man.openbsd.org/socket.2 doesn't say SOCK_DNS is limited to UDP.  DNS will upgrade from UDP to TCP if it gets back a DNS response with the 'TC' bit set indicating the response was truncated due to UDP packet size limitations.  Are you sure SOCK_DNS is really limited to UDP?  It wouldn't make sense that it would be.

marty1885:
You are right. I forgot that DNS can run over TCP. Description edited. I also force pushed to update my commit message

bradh352:
Seems reasonable.  I'll merge after the 1.24.0 release which has already been staged.

PR #679, opened 2024-01-05
optimize inet_ntop method

0xdddddddd:
It can reduce the amount of external calling code.
when dealing with IPv4 and IPv6 externally, it can effectively reduce their business code volume.


bradh352:
You can't change the meaning of the function.  This function is not documented to take a struct sockaddr like you are treating it, it takes only the pointer to the binary network byte order address itself.

PR #685, opened 2024-01-12
Do not sanity check RR Name vs Question

bradh352:
It appears as though we should never sanity check the RR name vs the question name as some DNS servers may return results for alias records.

Fixes Bug: #683

### DIFF
--- a/src/lib/ares__parse_into_addrinfo.c
+++ b/src/lib/ares__parse_into_addrinfo.c
@@ -81,7 +81,6 @@ ares_status_t ares__parse_into_addrinfo(const unsigned char *abuf, size_t alen,
   }
 
   for (i = 0; i < ancount; i++) {
-    const char          *rname = NULL;
     ares_dns_rec_type_t  rtype;
     const ares_dns_rr_t *rr =
       ares_dns_record_rr_get(dnsrec, ARES_SECTION_ANSWER, i);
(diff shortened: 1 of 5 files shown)


MaxXor:
Please release a small bugfix version 1.25.1 with this fix. It's rather important...

bradh352:
I see in the comment by @FingerlessGlov3s that removing the search option fixes this, that seems odd, can that be elaborated on?  This issue has existed basically forever, seems odd we'd just be seeing it now.

FingerlessGlov3s:
> I see in the comment by @FingerlessGlov3s that removing the search option fixes this, that seems odd, can that be elaborated on? This issue has existed basically forever, seems odd we'd just be seeing it now.

I've had that search option set for many years, only started happening after the move to Alpine 3.19 in the mentioned mailcow project above for the containers. When using the previous Alpine 3.17 version, the issue can't be reproduced, while the search option is set on the host which gets trickled down to the container. Using Alpine 3.19 the issue happens.

bradh352:
well, alpine is now building curl with c-ares whereas before they were using the system resolver, so that doesn't really help the c-ares project narrow anything down to ensure this issue really fixes it.  We really need to know that this fixes it.

Regarding a release date, once #696 and #693 are merged, we'll be doing a 1.26.0 release.  Likely that will be the end of the week.

Issue #687, opened 2024-01-14
`ares_fds` can cause an abort or write out-of-bounds


sfan5:
`select` has implementation-defined limits on the highest fd it can accept, e.g. 1024 on Linux.
When it is is exceeded and the API user calls `ares_fds` bad things happen.

## Test code

The test code provokes this issue by filling the fd table, but note that this can happen naturally if you e.g. use c-ares in a big server application.

```c
#include <stdio.h>
#include <string.h>
#include <unistd.h>
#include <sys/select.h>
#include <sys/resource.h>
#include <ares.h>

void test(ares_channel_t *channel) {
	fd_set dummy;
	ares_fds(channel, &dummy, &dummy);
}

static void empty(void *arg, int status, int timeouts, struct hostent *host)
{}

int main() {
	struct rlimit lim = { .rlim_cur = 3000, .rlim_max = 3000 };
	if (setrlimit(RLIMIT_NOFILE, &lim) == -1)
	{ perror("setrlimit"); return 1; }

	for (int i = 2; i < FD_SETSIZE + 42; i++) {
		int ret = dup(2);
		if (ret == -1) { perror("dup"); return 1; }
	}

	ares_library_init(ARES_LIB_INIT_ALL);
	ares_channel_t *channel;
	ares_init(&channel);

	ares_gethostbyname(channel, "google.com", AF_INET6, empty, NULL);

	test(channel);
}
```

## Expected result

Not sure actually. Maybe `ares_fds` could return an error?
`select` is just fundamentally broken in this situation so I think the documentation should also discourage using it.

## Actual result

```
% ./test
*** bit out of range 0 - FD_SETSIZE on fd_set ***: terminated
zsh: IOT instruction (core dumped)  ./test
```

Or if you have c-ares compiled without `_FORTIFY_SOURCE` you can actually see it corrupting stack data.
Replace `test()` with this:
```c
	char one[10] = {0};
	fd_set dummy;
	char two[10] = {0};

	ares_fds(channel, &dummy, &dummy);

	for (int i = 0; i < 10; i++)
		printf("%02x ", (int)one[i]);
	for (int i = 0; i < 10; i++)
		printf("%02x ", (int)two[i]);
	puts("");
```
results in
```
% ./test                                                
00 08 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
```

bradh352:
Honestly not sure if there is anything we can/should do other than to annotate in the docs.

[...]

PR #690, opened 2024-01-16
man ares_fds(3): mark as deprecated and add explanation

bradh352:
ares_fds(3) is not safe to use, mark as deprecated.

Fixes Issue: #687


bradh352:
committed to wrong branch

Issue #686, opened 2024-01-13
Feature: Disable parsing of known RR type


eriklax:
If it's not possible (and I was really digging deep). It would be useful if you could force ares_dns_parse() or ares_dns_rr_get_XX to NOT parse a rrtype response and known RR type and allow it to end up as ARES_RR_RAW_RR_DATA (as it were unknown). I want to do custom parsing of the RR data that is known to c-ares using the same API.

bradh352:
That should be trivial to add a flag for the parser to do that.  The real issue, however, is for name expansion for the RRs that RFC1035 allows the use of name compression for.  You wouldn't have the ability to just extract the RR itself and parse it independently without the rest of the data due to back references.

Out of curiosity, what is the exact use case?  The new parser is fairly robust, just want to know where it may be lacking.

eriklax:
I'm migrating a resolver function in a (DSL) scripting language from libresolv (with custom parsers) to c-ares. It currently have a more limited RR type support than c-ares, so for types such as "TLSA", the user would expect to receive raw data and unpack it himself. This makes it hard to implement, since c-ares parses this data and I the user expects it unparsed (raw bytes).

While I could undo the parsing, and pack the data again I don't later versions of c-ares to add support for more DNS-types than I've implemeneted (hence breaking the raw fallback in the future).

I'm not conserned about name compression and back references.

bradh352:
sure, should be pretty trivial to get into the next release

eriklax:
Thanks, that would be so appreciated. It was so great that the parser and writer API was made public!

eriklax:
As per the PR, for my use-case only `ARES_DNS_PARSE_ANSWER_RR_RAW` is needed. But I think it could make sense to allow disabling parsing of all three types for completeness (either invidual) or all together implementing only a `ARES_DNS_PARSE_RR_RAW` flag.

PR #693, opened 2024-01-17
Added flags to are_dns_parse to force RAW packet parsing

eriklax:
This pull request adds three flags to control parsing behaviour of RR types in various sections. This addresses the feature request in https://github.com/c-ares/c-ares/issues/686.

### DIFF
--- a/src/lib/ares_dns_parse.c
+++ b/src/lib/ares_dns_parse.c
@@ -1004,8 +1004,6 @@ static ares_status_t ares_dns_parse_rr(ares__buf_t *buf, unsigned int flags,
   size_t              remaining_len = 0;
   size_t              processed_len = 0;
 
-  (void)flags; /* currently unused */
-
   /* All RRs have the same top level format shown below:
    *                                 1  1  1  1  1  1
    *   0  1  2  3  4  5  6  7  8  9  0  1  2  3  4  5
@@ -1064,6 +1062,16 @@ static ares_status_t ares_dns_parse_rr(ares__buf_t *buf, unsigned int flags,
   rdlength = u16;
 
   if (!ares_dns_rec_type_isvalid(type, ARES_FALSE)) {
+    type = ARES_REC_TYPE_RAW_RR;
+  }
+
+  if (sect == ARES_SECTION_ANSWER && (flags & ARES_DNS_PARSE_ANSWER_RR_RAW)) {
+    type = ARES_REC_TYPE_RAW_RR;
+  }
+  if (sect == ARES_SECTION_AUTHORITY && (flags & ARES_DNS_PARSE_AUTHORITY_RR_RAW)) {
+    type = ARES_REC_TYPE_RAW_RR;
+  }
+  if (sect == ARES_SECTION_ADDITIONAL && (flags & ARES_DNS_PARSE_ADDITIONAL_RR_RAW)) {
     type = ARES_REC_TYPE_RAW_RR;
   }
 
--- a/src/lib/ares_dns_parse.c
+++ b/src/lib/ares_dns_parse.c
@@ -1003,6 +1003,7 @@ static ares_status_t ares_dns_parse_rr(ares__buf_t *buf, unsigned int flags,
   ares_dns_rr_t      *rr            = NULL;
   size_t              remaining_len = 0;
   size_t              processed_len = 0;
+  ares_bool_t         namecomp;
 
   /* All RRs have the same top level format shown below:
    *                                 1  1  1  1  1  1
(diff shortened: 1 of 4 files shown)


bradh352:
We probably need to think a little more about the flags.  Like maybe someone would want to go ahead and parse any records from RFC1035 as those may be using name compression, but any other records to leave raw.

eriklax:
If I reduce it to a single flag to leave everything raw, regardless of section that would reduce the footprint but still cover my use-case.

The use-case I'm trying to solve doesn't care if the RR data isn't useful for records using mangling, just want to have it without as-is without a custom parser on top of c-area :)

test/ares-test-internal.cc
```diff
@@ -816,6 +816,89 @@ TEST_F(LibraryTest, DNSRecord) {
...
+    ares_dns_rr_set_str(rr, ARES_RR_PTR_DNAME, "b.example.com"));
+
+  qdcount = ares_dns_record_query_cnt(dnsrec);
+  ancount = ares_dns_record_rr_cnt(dnsrec, ARES_SECTION_ANSWER);
+  nscount = ares_dns_record_rr_cnt(dnsrec, ARES_SECTION_AUTHORITY);
+  arcount = ares_dns_record_rr_cnt(dnsrec, ARES_SECTION_ADDITIONAL);
+
+  /* Write */
+  EXPECT_EQ(ARES_SUCCESS, ares_dns_write(dnsrec, &msg, &msglen));
+
```
bradh352, line 873: need
```
```
here to get rid of the memory leak in the test cases

src/lib/ares_dns_parse.c
```diff
@@ -1067,6 +1065,16 @@ static ares_status_t ares_dns_parse_rr(ares__buf_t *buf, unsigned int flags,
...
+  if (sect == ARES_SECTION_ANSWER && (flags & ARES_DNS_PARSE_ANSWER_RR_RAW)) {
+    type = ARES_REC_TYPE_RAW_RR;
+  }
+  if (sect == ARES_SECTION_AUTHORITY && (flags & ARES_DNS_PARSE_AUTHORITY_RR_RAW)) {
+    type = ARES_REC_TYPE_RAW_RR;
+  }
+  if (sect == ARES_SECTION_ADDITIONAL && (flags & ARES_DNS_PARSE_ADDITIONAL_RR_RAW)) {
+    type = ARES_REC_TYPE_RAW_RR;
+  }
+
```
bradh352, line 1077: I'd prefer to see this split into 6 total flags, or reduced to 2.  I know for your use case this is sufficient but I don't want to be locked in.  I'd like to see a split based on `ares_dns_rec_type_allow_name_compression(type)`, so either:

```
ARES_DNS_PARSE_AN_BASE_RAW /* Answers from RFC 1035 that allow name compression */
ARES_DNS_PARSE_NS_BASE_RAW /* Authority from RFC 1035 that allow name compression */
ARES_DNS_PARSE_AR_BASE_RAW /* Additional from RFC 1035 that allow name compression */
ARES_DNS_PARSE_AN_EXT_RAW /* Answers from later RFCs (no name compression) */
ARES_DNS_PARSE_NS_EXT_RAW /* Authority from later RFCs (no name compression) */
ARES_DNS_PARSE_AR_BASE_RAW /* Additional from later RFCs (no name compression) */
```
or 
```
ARES_DNS_PARSE_BASE_RAW /* Records from RFC 1035 that allow name compression */
ARES_DNS_PARSE_EXT_RAW /* Records from later RFCs (no name compression) */
```

bradh352 (changes requested):

eriklax:
Thanks for the feedback! I updated the patch accordingly!

src/lib/ares_dns_parse.c
```diff
@@ -1067,6 +1065,17 @@ static ares_status_t ares_dns_parse_rr(ares__buf_t *buf, unsigned int flags,
     type = ARES_REC_TYPE_RAW_RR;
   }
 
+  ares_bool_t namecomp = ares_dns_rec_type_allow_name_compression(type);
```
bradh352, line 1068: c89 doesn't allow variable declaration inline with code.

test/ares-test-internal.cc
```diff
@@ -816,6 +816,119 @@ TEST_F(LibraryTest, DNSRecord) {
...
+  EXPECT_EQ(qdcount, ares_dns_record_query_cnt(dnsrec));
+  EXPECT_EQ(2, ares_dns_record_rr_cnt(dnsrec, ARES_SECTION_ANSWER));
+  EXPECT_EQ(1, ares_dns_record_rr_cnt(dnsrec, ARES_SECTION_AUTHORITY));
+  EXPECT_EQ(1, ares_dns_record_rr_cnt(dnsrec, ARES_SECTION_ADDITIONAL));
+
+  rr = ares_dns_record_rr_get(dnsrec, ARES_SECTION_ANSWER, 0);
+  EXPECT_EQ(ARES_REC_TYPE_RAW_RR, ares_dns_rr_get_type(rr));
+
+  rr = ares_dns_record_rr_get(dnsrec, ARES_SECTION_ANSWER, 1);
+  EXPECT_EQ(ARES_REC_TYPE_TLSA_RR, ares_dns_rr_get_type(rr));
```
bradh352, line 894: Think you mean just `ARES_REC_TYPE_TLSA`, there's no `RR` suffix

eriklax: How did I miss this 🤦‍♂️

PR #705, opened 2024-01-31
lower EDNSPACKETSZ to 1232

crrodriguez:
In 2020, it was agreed this is optimal maximum size and all major server software was updated to reflect this.

see https://www.dnsflagday.net/2020/#faq


bradh352:
interesting, our current default is 1280 at least so based on https://indico.dns-oarc.net/event/36/contributions/776/attachments/754/1277/DefragDNS-Axel_Koolhaas-Tjeerd_Slokker.pdf at least it should be imperceptible.  But yeah, we should use the current consensus recommendation.

PR #706, opened 2024-02-01
mark deprecated functions as such

crrodriguez:
Multiple functions have been deprecated over the years, annotate them with __attribute__ deprecated.

When possible show a message about their replacements.

### DIFF
--- a/include/ares.h
+++ b/include/ares.h
@@ -119,6 +119,37 @@ extern "C" {
 #  endif
 #endif
 
+#ifdef __GNUC__
+#define CARES_GCC_VERSION (__GNUC__ * 10000 \
+                     + __GNUC_MINOR__ * 100 \
+                     + __GNUC_PATCHLEVEL__)
+#else
+#define CARES_GCC_VERSION 0
+#endif
+
+#ifndef __has_attribute
+    #define __has_attribute(x) 0
+#endif
+
+#ifdef CARES_ALLOW_DEPRECATED
+#define CARES_DEPRECATED
+#define CARES_DEPRECATED_FOR(f)
+#else
+#if CARES_GCC_VERSION >= 30200 || __has_attribute (__deprecated__)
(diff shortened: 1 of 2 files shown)


bradh352:
Looks pretty good, a couple minor things:

1. `ares_process()` should point to `ares_process_fd()` instead of pointing to using the event thread since not all systems may have threads (or c-ares might be built without thread support).  Since `ares_process()` assumes the use of `select()`, it should still be deprecated.
2. We probably need a way to set a build flag to disable deprecation warnings, like `CARES_NO_DEPRECATION` or something and use this in the test cases to not emit the warnings just because we're going to still need to have test cases for these functions that are deprecated and don't want to warn on their use.
3. Probably the `ares_parse_{a,aaaa,srv,ptr....}_reply()` functions should point to `ares_dns_parse()`
4. Probably `ares_gethostbyname()` should point to `ares_getaddrinfo()`

bradh352:
were you going to finish this off, or did you want me to run with it from here and make those changes I suggested?

crrodriguez:
> were you going to finish this off, or did you want me to run with it from here and make those changes I suggested?

I will.  don't worry.

bradh352:
FYI, we're looking to release c-ares 1.27.0 in about a week.

crrodriguez:
I have updated the PR with your suggestions/additions now. sorry for the delay.

crrodriguez:
> any update on this PR with my latest feedback ?

Not yet, but regardless point 2.. so you want it before the prototype so it works on msvc..I choose the other way around because it works everywhere else and looks less ugly. 
Other points are ok with me.

bradh352:
on point 2, if it works on msvc *and* everything else to put the attribute up front, then it seems like what we'll need.  Otherwise we'll have to figure out what to do.

bradh352:
moved to PR #732

PR #721, opened 2024-03-09
simple implementation for isascii where stdlib isascii is not available

fffaraz:
I had a similar patch for wolfssl as well.
https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/types.h#L934


bradh352:
I'd rather just always use your definition as its always correct.  That said, we should cast `c` to `unsigned char` when checked so we don't get warnings and we can omit the check for >= 0.  If you don't do both of those, depending on where its checked, you might get a warning emitted from the compiler about one condition always being true.

PR #728, opened 2024-03-19
Include netinet6/in6.h

farazrbx:
On some platforms, "netinet6/in6.h" is not included by "netinet/in.h" and needs to be included separately.


bradh352:
We probably need to add this header check to configure.ac and CMakeLists.txt too to set that header definition.  What platform is this?

farazrbx:
> We probably need to add this header check to configure.ac and CMakeLists.txt too to set that header definition.
Added for CMake. Not sure how to add it for configure.ac.

> What platform is this?
Same platform that doesn't have isascii. PlayStation.

bradh352:
For configure.ac you'd just add it to the list here: https://github.com/c-ares/c-ares/blob/main/configure.ac#L419

farazrbx:
@bradh352 Could you review, please?

Issue #717, opened 2024-02-24
Server list order reset


nUl1:
Previously #594 introduced a dynamic server list ordering based on the number of consecutive failures. 
It works great in cases where all servers in the list are interchangeable but issues might emerge when this list was originally ordered with server priorities in mind (e.g. primary server followed by a backup one).

Currently there seems to be no clear way to bring a server back up in the list. Options I see are: 
 * Fail enough queries at all other servers to make a query go to this one;
 * Do a server list update that removes this server then reintroduce it with a second update;
 * Create a new channel and retire the old one.

Maybe a flag can be added to the CSV representation that will force failure count reset in `servers_update`? Or a method that replaces the whole list? Or did I miss some obvious solution to this problem?

bradh352:
Out of curiosity, is there a particular purpose to defining a primary vs backup?  I've never found a *legitimate* purpose for such a thing where there wasn't a better solution for the actual problem.  But if the ordering really does matter, then we'd probably want to set a timer (not sure what the default should be, 5mins?) to invalidate the failure count and resort the server list upon expiration of this timer to possibly restore the original ordering.  It wouldn't be *that* difficult to do if there is a real reason to do so.

nUl1:
> Out of curiosity, is there a particular purpose to defining a primary vs backup?
Most of the cases I've seen were various caching servers. Basically distibuting the server load with smaller instances and reducing latency/network traffic.

> we'd probably want to set a timer [...] to invalidate the failure count
That's exactly what I implemented application-side before finding out that `set_servers` does not affect the order for existing entries.

bradh352:
ok, so really just for load balancing reasons.  I can look at implementing this internally after I finish rewriting the last config parser using the new memory-safe parsing routines.  There's no good way to do it externally without killing in-flight queries.

That said, for load reduction, you should look at enabling the query cache in c-ares.  Most apps tend to look up the same few domains over and over and the built-in query cache in c-ares uses the advertised TTLs to properly cache just like the upstream DNS server would but locally (so lowest latency possible and no load to the upstream servers).

bradh352:
You can change the server list no problem without affecting inflight queries, they'll just be requeued.  The problem is if you clear the server list, then repopulate the server list, all queries will be killed upon clearing the list.  And as you've noticed, if you repopulate the server list with the same list of existing servers, it will not re-order them to the order you provided due to failure counts.

bradh352:
fixed in #731

PR #731, opened 2024-03-26
Add server failover retry behavior, where failed servers are retried with small probability after a minimum delay

oliverwelsh:
**Summary**

By default c-ares will select the server with the least number of consecutive failures when sending a query. However, this means that if a server temporarily goes down and hits failures (e.g. a transient network issue), then that server will never be retried until all other servers hit the same number of failures.

This is an issue if the failed server is preferred to other servers in the list. For example if a primary server and a backup server are configured.

This PR adds new server failover retry behavior, where failed servers are retried with small probability after a minimum delay has passed. The probability and minimum delay are configurable via the `ARES_OPT_SERVER_FAILOVER` option. By default c-ares will use a probability of 10% and a minimum delay of 5 seconds.

This will solve https://github.com/c-ares/c-ares/issues/717.

In addition, this PR includes a small change to always close out connections to servers which have hit failures, even with `ARES_FLAG_STAYOPEN`. It's possible that resetting the connection can resolve some server issues (e.g. by resetting the source port).

**Testing**

A new set of regression tests have been added to test the new server failover retry behavior.


src/lib/ares_process.c
```diff
@@ -817,6 +848,76 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+    /* The best server is not the highest priority server. To avoid never
+     * retrying higher priority servers, with some probability we select a
+     * higher priority server at random instead.
+     */
+    ares__rand_bytes(channel->rand_state, &c, 1);
+    if ((size_t)c % channel->priority_server_chance == 0) {
+      /* Select a random higher priority server. */
+      ares__rand_bytes(channel->rand_state, &c, 1);
+      idx = (size_t)c;
+      idx = idx % server->idx;
```
bradh352, line 906: Should this be `idx = idx % ares__slist_len(channel->servers);` ?

include/ares.h
```diff
@@ -224,6 +224,9 @@ typedef enum {
 #define ARES_OPT_MAXTIMEOUTMS    (1 << 20)
 #define ARES_OPT_QUERY_CACHE     (1 << 21)
 #define ARES_OPT_EVENT_THREAD    (1 << 22)
+#define ARES_OPT_SERVER_FAIL     (1 << 23)
+#define ARES_OPT_SERVER_RECOVER  (1 << 24)
+#define ARES_OPT_PRIORITY_SERVER (1 << 25)
```
bradh352, line 230: Not sure I'm comfortable adding 3 new options since we're running out. We only have 31 bits to work with.   Can we instead reduce this to 1 option that requires you to set all 3 settings?  Basically make a struct a member of ares_options.  Maybe just `ARES_OPT_SERVER_FAILOVER` to control failover settings.

src/lib/ares_process.c
```diff
@@ -71,35 +71,66 @@ static void          end_query(ares_channel_t *channel, struct query *query,
...
                                const ares_dns_record_t *dnsrec);
 
-static void          server_increment_failures(struct server_state *server)
+static void server_increment_failures(struct server_state *server)
 {
-  ares__slist_node_t   *node;
-  const ares_channel_t *channel = server->channel;
+  ares__slist_node_t *node;
+  ares_channel_t     *channel = server->channel;
+  char                server_ip[INET6_ADDRSTRLEN];
```
bradh352, line 79: Move `server_ip` under the `if()` statement where it is used.  c89 allows variable declarations at the beginning of any code block.

src/lib/ares_process.c
```diff
@@ -71,35 +71,66 @@ static void          end_query(ares_channel_t *channel, struct query *query,
...
+
+  server->consec_successes = 0;
   server->consec_failures++;
   ares__slist_node_reinsert(node);
+
+  /* If the server has just become unhealthy, then invoke the server state
+   * callback.
+   */
+  if ((server->consec_failures >= channel->server_failure_threshold) &&
+      (server->is_healthy == ARES_TRUE)) {
```
bradh352, line 93: Avoid extra parens unless it really adds clarity due to order of operations.

src/lib/ares_process.c
```diff
@@ -71,35 +71,66 @@ static void          end_query(ares_channel_t *channel, struct query *query,
...
+    server->is_healthy = ARES_FALSE;
+    if (channel->server_state_cb) {
+      ares_inet_ntop(server->addr.family, &server->addr.addr, server_ip,
+                     sizeof(server_ip));
+      channel->server_state_cb(server_ip, server->is_healthy,
+                               channel->server_state_cb_data);
+    }
+  }
```
bradh352, line 101: so `channel->server_failure_threshold` is only used to control the `is_healthy` flag, the server list still gets sorted regardless?  Seems like it may take quite a while to trigger the callback for unhealthy since you're relying on a good number of queries with `priority_server_chance` to trigger this.  Maybe that's ok, but would definitely need to be documented.

src/lib/ares_process.c
```diff
@@ -71,35 +71,66 @@ static void          end_query(ares_channel_t *channel, struct query *query,
...
 {
-  ares__slist_node_t   *node;
-  const ares_channel_t *channel = server->channel;
-
-  if (!server->consec_failures) {
-    return;
-  }
+  ares__slist_node_t *node;
+  ares_channel_t     *channel = server->channel;
+  char                server_ip[INET6_ADDRSTRLEN];
```
bradh352, line 108: Move server_ip under the if() statement where it is used. c89 allows variable declarations at the beginning of any code block.

bradh352 (commented):
Overall I think this concept would be appreciated by sysadmins since eventually when all dns servers are healthy, queries would be directed towards the appropriate servers in the configured order after a failure.

I'm not entirely convinced of the usefulness of the server state callback as it is sort of random when it might be triggered.  I wonder if it would be more useful to simply have the server state callback be triggered on *every* state change.  Meaning if a server fails (either on initial attempt, or after a prior successful attempt), trigger the callback.  Likewise if a server that last had a failure now has a success, trigger the callback.  The callback should really be "cheap", and only triggering on changes (not every single query) would be fine.  Then app developers could add their own logic for their own notifications, etc.

That would also get rid of needing to have configurations for the thresholds while also reducing complexity overall.

src/lib/ares_options.c
```diff
@@ -474,6 +487,32 @@ ares_status_t ares__init_by_options(ares_channel_t            *channel,
...
+        (size_t)options->server_recovery_threshold;
+    }
+  }
+  if (optmask & ARES_OPT_PRIORITY_SERVER) {
+    if (options->priority_server_chance <= 0) {
+      optmask &= ~(ARES_OPT_PRIORITY_SERVER);
+    } else {
+      channel->priority_server_chance =
+        (size_t)options->priority_server_chance;
+    }
```
bradh352, line 513: Is there maybe a maximum value we should enforce here?  due to grabbing a random value in the range of 0-255, and using a % to check for probability, it seems like there should be some range restrictions for it to be usable.  That said, I guess you have a 1 in 256 chance of the random value itself being 0 so anything above that range would still every once in a while "hit".

src/lib/ares_process.c
```diff
@@ -817,6 +848,76 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+    return ares__random_server(channel);
+  }
+
+  if (server->idx > 0) {
+    /* The best server is not the highest priority server. To avoid never
+     * retrying higher priority servers, with some probability we select a
+     * higher priority server at random instead.
+     */
+    ares__rand_bytes(channel->rand_state, &c, 1);
+    if ((size_t)c % channel->priority_server_chance == 0) {
```
bradh352, line 902: if `channel->priority_server_chance` is never initialized, this is a divide by zero.

src/lib/ares_process.c
oliverwelsh: `server->idx` is correct. Because we are wanting to pick a random higher priority server, which will be one with index from `0, ..., server->idx - 1`.

src/lib/ares_process.c
oliverwelsh: I will move the variable declaration (I'm used to subtly different coding patterns and need to break the habit!).

On writing out the server, I notice that `ares_get_servers_csv()` has logic for writing out a server to a string (including port and interface). Perhaps we could pull this out into a function, and re-use it for the server state callback?

src/lib/ares_process.c
oliverwelsh: Thanks, I will get these all updated.

src/lib/ares_process.c
oliverwelsh: Makes sense, I will add a check that `channel->priority_server_chance > 0` (otherwise we can just go for the first server in the list).

src/lib/ares_process.c
bradh352: Hmm, because of the introduction of the callback for server state, we very well may want to rethink this a little.  Especially in the case where multiple servers go into a bad state simultaneously (like a network outage on the dns client side), they might get notifications of all these servers going down, but then only get a notification of some smaller portion of servers coming back online due to the logic in this area which could lead to improper conclusions and investigations based on this misinformation.

src/lib/ares_process.c
bradh352: yeah, that would make sense.  It might also be good to have an unused 'flags' field for additional information about the server we can use in the future if we want (like indicating mDNS, DNS over HTTPS, etc).

src/lib/ares_process.c
oliverwelsh: To give a bit of context, this change was originally written prior to the sorted server list, when:
- Servers were stored in an array on the channel.
- The server to send the next request to was stored in the `query` structure (`query->server`).
- If a request needed to be retried, we moved onto the next server with `query->server = (query->server + 1) % channel->nservers`.

This change still tracked healthy/unhealthy servers in the same way (based on consecutive failures/successes). But unhealthy servers were skipped at the point when moving onto the next server. E.g. when we moved to the next server, if it was unhealthy then we'd skip it with 90% probability.

This doesn't quite work now, because we always use the first server in the sorted list on retries. For example if we sorted by `(is_healthy, idx)`, then retries would always go to the first server until it hits 3 failures and gets marked as unhealthy. But I think we want retries to cycle through the servers in the list.

oliverwelsh: Perhaps the following would work:

- Introduce a new field on servers called `requests_for_current_query` that tracks how many times the server has been used on the current query.
- This field could be incremented in `server_increment_failures()` and `server_increment_successes()`. And it would get reset to 0 on all servers before the first call to `ares__send_query()`
- Servers would then be sorted by `(is_healthy, requests_for_current_query, idx)`.

That way:
- Unhealthy servers would be deprioritised. But only after they have hit the threshold for the number of consecutive failures.
- We still move to new servers when retrying requests within the same query.
- The highest priority healthy server will be used on the first attempt in a query. (We could still have the same logic to retry higher priority unhealthy servers with a small probability.)

bradh352: sounds a little complex for what is really needed, and we can't sort the server list by requests_for_current_query since there can be multiple outstanding queries to a single server server at one time.

I do think it makes sense to have logic to retry unhealthy servers with a small probability, however it probably makes sense to have a minimum time for such a try as otherwise for applications which perform a large number of DNS queries (thinking here, something doing a lot of reverse lookups in a webserver log to dereference an ip address to a name), that would end up stalling a lot if that server was still down.  So we'd want to track both down time and if its actively being attempted to see if its still live by another query incase it 'hits' again.

And I really do think we need to remove the is_healthy concept along with number of tries for up or down, as we can leave the actual alert logic to the application.

src/lib/ares_process.c
oliverwelsh: I moved out the logic to write a sever to a buffer into a new function:
`ares_status_t ares_write_server(const struct server_state *server, ares__buf_t *buf)`

In terms of future-proofing for things like mDNS and DNS over HTTPS, I suppose that info would likely be stored on the `server_state` struct? In which case `ares_write_server()` would have the info available.

include/ares.h
oliverwelsh: Thank you, I have reworked the server healthy concept (will comment in detail on another thread). I still needed 3 options for server failover behavior, but combined then into a single opt `ARES_OPT_SERVER_FAILOVER` with struct `ares_server_failover_options`.

src/lib/ares_process.c
oliverwelsh: I have reworked this so that retries are sent to all failed servers, rather than just "higher priority" failed servers.

I think this makes sense; it means that if any lower priority server fails, we will eventually retry it and update the server state callback (although once it recovers, it will not be re-selected until all higher priority servers have failures).

src/lib/ares_process.c
oliverwelsh: Hi Brad, thank you for the feedback and apologies for the delay in getting the markups. I have pushed some further changes based on the feedback. A summary:

Server state callback:
- I removed the concept of a healthy/unhealthy server (application code can handle decisions like this).
- The callback now always fires on success or failure (with the `ares_bool_t` argument indicating this).
- The server string argument now includes extra details like the port and interface.

Server failover retry behaviour:
- This is now controlled by 3 fields:
  - `retry_chance` - The probability that we retry a failed server.
  - `retry_delay` - The minimum delay between retries (channel-wide delay, rather than per-server).
  - `serious_fail_limit` - The number of consecutive failures beyond which a server is considered to have a serious failure.
    - If all servers exceed this, then servers are always selected at random.
    - Moreover, connections to such servers are always closed out.
- When sending queries, if the `retry_delay` has passed and the `retry_chance` is hit, then we select the first failed server in the server list.
- We no longer consider "higher priority" failed servers. This means that all failed servers will eventually be retried.

I still need to test this and there are some optimisations I can still make, but the high level behaviour should be OK for re-review. Please let me know if you have further thoughts. Thanks!

src/lib/ares_process.c
```diff
@@ -816,6 +854,77 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+          /* Check that the next retry time for failed servers has passed. If
+           * not then break out and return the best server.
+           */
+          struct timeval now = ares__tvnow();
+          if (ares__timedout(&now, &channel->server_next_retry_time)) {
+            timeadd(&now, channel->server_retry_delay);
+            channel->server_next_retry_time = now;
+            return node_val;
+          }
+          break;
```
nUl1, line 918: Wouldn't it be better to check the timeout outside of the loop and skip it altogether?

Also we can probably check the last server in the list to see if we have any failed servers to retry.

oliverwelsh: I put the timeout check inside the loop because I wanted to avoid the call to `ares_tvnow()` if there were no failed servers to retry (optimising the mainline "all servers work" case and avoiding system calls).

But yep, a simple check of the last server (can use `ares__slist_node_last()`) would improve this. I will fix this. Thanks!

src/lib/ares__close_sockets.c
```diff
@@ -97,6 +97,16 @@ void ares__check_cleanup_conn(const ares_channel_t     *channel,
...
+  /* If the associated server has a serious failure, close it out. Resetting
+   * the connection (and specifically the source port number) can help resolve
+   * situations where packets are being dropped.
+   */
+  if (channel->optmask & ARES_OPT_SERVER_FAILOVER &&
+      channel->server_serious_fail_limit > 0 &&
+      conn->server->consec_failures >= channel->server_serious_fail_limit) {
+    do_cleanup = ARES_TRUE;
+  }
+
```
bradh352, line 109: This really only has an impact if someone is using ARES_FLAG_STAYOPEN. Honestly, I think this should just be:
```
if (conn->server->consec_failures) {
```
and call it a day.  I don't see a need to hit some limit to kill the connection in this edge case, if there's a failure, kill it.

src/lib/ares_private.h
```diff
@@ -497,6 +523,8 @@ ares_status_t ares__sconfig_append_fromstr(ares__llist_t **sconfig,
 ares_status_t ares_in_addr_to_server_config_llist(const struct in_addr *servers,
                                                   size_t          nservers,
                                                   ares__llist_t **llist);
+ares_status_t ares_write_server(const struct server_state *server,
+                                ares__buf_t *buf);
```
bradh352, line 528: I'm not a fan of this function name.  Maybe `ares_get_server_addr()`.

src/lib/ares_process.c
bradh352: at least on linux ares_tvnow() should be very fast due to it using a vdso call so I wouldn't really worry about optimizing out that call.  But otherwise, yeah, checking the last member to see if any servers have failed makes sense due to sorting.

src/lib/ares_process.c
```diff
@@ -816,6 +854,77 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+  if (server == NULL) {
+    return NULL;
+  }
+
+  if (channel->server_serious_fail_limit > 0 &&
+      server->consec_failures >= channel->server_serious_fail_limit) {
+    /* All servers have a serious failure so fall back to a random server. */
+    return ares__random_server(channel);
+  }
+
```
bradh352, line 897: lets get rid of the `server_serious_fail_limit`.  We still want to check servers with failures that haven't hit some random limit, remember, they've been sorted to the end so won't be tried unless other servers have also failed.

src/lib/ares_process.c
```diff
@@ -79,26 +114,29 @@ static void          server_increment_failures(struct server_state *server)
   if (node == NULL) {
     return;
   }
+
   server->consec_failures++;
   ares__slist_node_reinsert(node);
+
+  invoke_server_state_cb(server, ARES_FALSE);
```
bradh352, line 121: I'd think this should only be called on the first failure, not every.

src/lib/ares_process.c
```diff
@@ -79,26 +114,29 @@ static void          server_increment_failures(struct server_state *server)
...
   }
 
-  server->consec_failures = 0;
-  ares__slist_node_reinsert(node);
+  if (server->consec_failures > 0) {
+    server->consec_failures = 0;
+    ares__slist_node_reinsert(node);
+  }
+
+  invoke_server_state_cb(server, ARES_TRUE);
```
bradh352, line 139: Shouldn't this be inside the above `if()` statement.  I don't see a reason to notify of success on every query.

src/lib/ares_process.c
```diff
@@ -890,7 +999,11 @@ ares_status_t ares__send_query(struct query *query, struct timeval *now)
 
   /* Choose the server to send the query to */
   if (channel->rotate) {
+    /* Pull random server */
     server = ares__random_server(channel);
+  } else if (channel->optmask & ARES_OPT_SERVER_FAILOVER) {
+    /* Pull server with failover behavior */
+    server = ares__failover_server(channel);
```
bradh352, line 1006: I think we should probably make this always-on with some reasonable default settings that can be overwritten which would resolve #717

src/lib/ares_process.c
oliverwelsh: To give an example of the issue this was trying to solve:

- Suppose we have 2 servers, A and B.
- Initially, both servers are failed. We will alternate query attempts between the two.
  - After 2000 queries A and B will both have 1000 failures.
- Suppose that A recovers.
  - After the next query A will have 0 failures and B will have 1000 failures.
- Suppose that A fails again and B recovers shortly after.
  - It will take 1000 queries sent to A before B is ever retried, which may prolong recovery and cause longer outages.

The serious failure limit was added to give an upper bound on the number of failures before we retry servers randomly.

As an example. setting to `3` would mean if all servers hit `>=3` failures then they are all treated as "equally failed" and chosen from randomly.

Maybe this is too rare of an edge case to care about, especially with the new retry mechanism?

src/lib/ares__close_sockets.c
oliverwelsh: Ok, makes sense and seems sensible. I'll make this change.

src/lib/ares_process.c
oliverwelsh: So I think there are use-cases for calling on every success/failure. It would give extra control to applications, although I can understand that calling for every query (possibly multiple times) is a lot of overhead.

E.g. if an application wants to track metrics for how many success/failed requests are sent to each server.

Also, in my case knowing if a server is failing repeatedly rather than just a one-off is useful, to detect a more serious error rather than a transient network issue.

We should make a decision between:
- Calling on all failures and successes.
- Calling on just the first failure and the first success (that changes the state).

The former gives more control, whilst the latter is a simpler interface to use. Do you have a feel for which would be useful more generally?

oliverwelsh: Also, if we go for the former then I think we should cache the string representation of the server in the `server_state` object; having a pre-built string that we can return a `const` pointer to would save on performance.

src/lib/ares_process.c
oliverwelsh: See comment above on the corresponding case for failures.

src/lib/ares_process.c
bradh352: I'd think with the new random server failure retry, that alone would resolve the issue without needing a serious failure count concept, or am I misunderstanding something?

src/lib/ares_process.c
bradh352: You're the one with the actual usecase for this, are you taking some 'action' immediately when notified of a failure or anything like that?  I sort of wonder if it would make sense to have a server status query, where it would populate a data structure with all servers and statistics that could be iterated across so you don't have to actually track each server's state.

src/lib/ares_private.h
```diff
@@ -315,6 +325,22 @@ struct ares_channeldata {
...
+   * The retry chance is the probability (1/N) by which we will retry a failed
+   * server instead of the best server when selecting a server to send queries
+   * to.
+   * The retry delay is the minimum time in milliseconds to wait between doing
+   * such retries. The next retry time tracks the timestamp when this wait
+   * expires.
+   */
+  unsigned char  server_retry_chance;
+  size_t         server_retry_delay;
+  struct timeval server_next_retry_time;
```
bradh352, line 343: `server_next_retry_time` should really be a per-server value, not per-channel as each server's retry should be timed independently.

src/lib/ares_process.c
```diff
@@ -816,6 +854,83 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+
+  /* If no servers have failures, then prefer the first server in the list. */
+  if (last_server != NULL && last_server->consec_failures == 0) {
+    return first_server;
+  }
+
+  /* If all servers have failures then there is not much reason to prefer one
+   * over another, so select randomly.
+   */
+  if (first_server->consec_failures >= SERVER_ALL_FAILED_THRESHOLD) {
```
bradh352, line 901: Think this `if` statement doesn't do anything useful.  I'd remove this code block and let the rest of the code do its thing.  It will auto re-sort based on number of consecutive failures, makes the sequence more predictable as well which makes test cases easier to write and validate.

src/lib/ares_process.c
```diff
@@ -816,6 +854,83 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+
+  /* If we are not configured with a server retry chance or the server retry
+   * delay has not yet passed then return the first server in the list.
+   */
+  now = ares__tvnow();
+  if (channel->server_retry_chance == 0 ||
+      !ares__timedout(&now, &channel->server_next_retry_time)) {
+    return first_server;
+  }
+
```
bradh352, line 913: since `server_next_retry_time` should be per-server, this timeout check would need to be in the below loop instead.

src/lib/ares_process.c
```diff
@@ -816,6 +854,83 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+  /* Generate a random byte to decide whether to retry a failed server. */
+  ares__rand_bytes(channel->rand_state, &c, 1);
+  if (c % channel->server_retry_chance == 0) {
+    /* Select a failed server to retry. */
+    ares__slist_node_t *node;
+    for (node = ares__slist_node_first(channel->servers); node != NULL;
```
bradh352, line 919: probably need to make a helper to get the first failed server (since its a sorted list), then cycle forward from there based on the per-server retry timeout.

src/lib/ares_private.h
oliverwelsh: Sounds good. I have re-implemented the next retry time to be a field on the `server_state`.

src/lib/ares_process.c
oliverwelsh: Happy to make this default. I have done this with a 10% retry probability and a minimum delay of 5 seconds between retries. (The default timeout is 2 seconds and I think we want to be more than this.)

Please let me know if you'd prefer different values.

src/lib/ares_process.c
oliverwelsh: Removed, thanks.

src/lib/ares_private.h
```diff
@@ -315,6 +323,20 @@ struct ares_channeldata {
 
   /* Query Cache */
   ares__qcache_t                     *qcache;
+
+  /* Callback triggered when a server has a successful or failed response. */
+  ares_server_state_callback server_state_cb;
+  void                      *server_state_cb_data;
+
+  /* Fields controlling server failover behavior.
+   * The retry chance is the probability (1/N) by which we will retry a failed
```
bradh352, line 332: pretty sure the probability isn't 1/N, as values 128-255 have the same probability of hitting, 2 in 256 (0 and itself). 86-127 have the same probability as well, 3 in 256. 64-85 have the same probability, 4 in 256.

Seems logarithmic, but haven't thought hard enough to say what the algorithm really is.

oliverwelsh: Ah, good spot! The probability will be ceil(256/N)/256, which is 1/N rounded up to the nearest 1/256. I think this is best we can do with a single random byte, as we only have 8 bits of precision.

Perhaps we can use 4 random bytes, which would give us probabilities to the nearest 1/2^32. That feels much better in terms of precision and we could always document the specific behaviour.

oliverwelsh: I have moved the retry chance to use `size_t`s, which should give increased precision. I think the precision will be small enough that external documentation is not hugely important, although happy to add a note when the man pages are written.

bradh352: I'm maybe slightly worried about draining the RNG since it does require a syscall to fetch more bytes and it is cryptographically secure which means its much slower than say something like xorshift (which would be sufficient for something like this).  At least we do pull from the OS in 256byte chunks these days.

Maybe grabbing 2 bytes (16bits) would be a good compromise?

oliverwelsh: Ok thanks, I think 1/2^16 will be more than enough precision. I have switch this to use an `unsigned short`.

src/lib/ares_process.c
```diff
@@ -816,6 +856,68 @@ static struct server_state *ares__random_server(ares_channel_t *channel)
...
+    /* Select a suitable failed server to retry. */
+    struct timeval      now = ares__tvnow();
+    ares__slist_node_t *node;
+    for (node = ares__slist_node_first(channel->servers); node != NULL;
+         node = ares__slist_node_next(node)) {
+      struct server_state *node_val = ares__slist_node_val(node);
+      if (node_val != NULL && node_val->consec_failures > 0 &&
+          ares__timedout(&now, &node_val->next_retry_time)) {
+        timeadd(&now, channel->server_retry_delay);
+        node_val->next_retry_time = now;
```
bradh352, line 911: This should be set in `server_increment_failures()` not here. Each time the server is tried and failed, it should push out the retry delay, regardless of if it was selected due to random selection or not (e.g. the highest priority server failed and its next in the priority order). 

This also mitigates the need to modify the test cases as it appears the current code path is allowing an *immediate* retry of the server without honoring the timeout.

It does make sense to add test cases for coverage for this scenario however, and that can be done using an ultra-low retry timeout and high probability, but that should be a separate test case and not modification of the existing one.

bradh352:
I'm thinking we need to back out the server state callback from this PR and open a different PR with just that logic.  As I'm comfortable with everything except that at this point, and honestly the failed server retry logic is a fairly significant change that would make sense to be stand alone.

src/lib/ares_process.c
oliverwelsh: I think we can do all of this cleanly in a single if statement:
```
      if (node_val != NULL && node_val->consec_failures > 0 &&
          ares__timedout(&now, &node_val->next_retry_time))
```
which skips through all healthy servers, and then skips past ones which haven't passed the retry delay.

Propose that we don't pull this out into a separate helper function.

src/lib/ares_process.c
oliverwelsh: Thanks, I definitely agree with that change and it makes the FV behaviour a lot more reliable.

I have added another set of test cases (and left the existing tests as-is), which cover 4 scenarios for server retries.
- No retries if all servers are healthy.
- Simple retry after the retry delay has passed.
- Failed servers are retried in sorted order (i.e. number of consec failures and then index).
- Failed servers which have not yet met the retry delay are skipped and not retried.

I set the retry to delay to 100ms in these tests, and call `usleep` to increment time. There is a trade off between reliability and speed of the tests, but I felt that 100ms was appropriate.

test/ares-test.cc
```diff
@@ -745,6 +745,14 @@ MockChannelOptsTest::MockChannelOptsTest(int count,
...
     optmask |= ARES_OPT_TRIES;
   }
+  // If not already overridden, set server failover options. For reliable
+  // testing we set the retry chance to 100%, and the retry delay to a high
+  // value (1 hour) so that retries are never performed on a test.
+  if (!(optmask & ARES_OPT_SERVER_FAILOVER)) {
+    opts.server_failover_opts.retry_chance = 1;
+    opts.server_failover_opts.retry_delay = 60 * 60 * 1000;
+    optmask |= ARES_OPT_SERVER_FAILOVER;
+  }
```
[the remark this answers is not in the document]
oliverwelsh: So the default retry delay is 5000ms, and some tests currently take ~1500ms to run.

I was thinking to leave this in, just in case longer-running tests are added in the future that hit the 5000ms default. Although it is probably safe to remove!

src/lib/ares_process.c
bradh352: yeah, i guess the likelihood of having an extensive server list is pretty small, likely 2-3 servers.  I was thinking scanning from the end where we know there's a bad server to find the first bad one was more optimal, but that would only be a win with a large number of servers.  I'm fine with this.

oliverwelsh:
> I'm thinking we need to back out the server state callback from this PR and open a different PR with just that logic. As I'm comfortable with everything except that at this point, and honestly the failed server retry logic is a fairly significant change that would make sense to be stand alone.

Thank you for the ongoing review so far. With the current iteration, the server state callback is completely separate to the server retry behaviour so I agree it makes sense to split off.

I will split this PR into two, and make this one just implement the server retry behaviour. I will also add man page docs now that we have settled on the changes.

test/ares-test.cc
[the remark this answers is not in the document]
bradh352: i'd go ahead and remove it since all tests pass at a much lower timeout.  besides retry_chance of 1 just seems wrong.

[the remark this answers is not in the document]
oliverwelsh: This is reverted now.

include/ares.h
```diff
@@ -326,6 +327,18 @@ typedef void (*ares_sock_state_cb)(void *data, ares_socket_t socket_fd,
...
+/* Options controlling server failover behavior.
+ * The retry chance is the probability (1/N) by which we will retry a failed
+ * server instead of the best server when selecting a server to send queries
+ * to.
+ * The retry delay is the minimum time in milliseconds to wait between doing
+ * such retries (applied per-server).
+ */
+struct ares_server_failover_options {
+  unsigned short retry_chance;
+  int            retry_delay;
```
bradh352, line 339: this member can be `size_t`, not reason to allow `int`

src/lib/ares_options.c
```diff
@@ -474,6 +482,19 @@ ares_status_t ares__init_by_options(ares_channel_t            *channel,
     }
   }
 
+  /* Set fields for server failover behavior */
+  if (optmask & ARES_OPT_SERVER_FAILOVER) {
+    if (options->server_failover_opts.retry_chance == 0 ||
```
bradh352, line 487: maybe we should allow `retry_chance` of 0 to disable this logic?

include/ares.h
oliverwelsh: Ok thanks, I have changed the option to use `size_t` to match the channel setting.

src/lib/ares_options.c
oliverwelsh: I think that is sensible, and the code in `ares__failover_server` is already resilient to the retry chance being 0. I have updated the options validation to allow this (and updated the man page docs to describe it). Thanks.

bradh352:
@oliverwelsh thanks for all your work on this, I know there were a lot of iterations here.  Looks good to me, was there anything left or do you feel this is good to merge?

oliverwelsh:
> @oliverwelsh thanks for all your work on this, I know there were a lot of iterations here. Looks good to me, was there anything left or do you feel this is good to merge?

Thanks for the ongoing support and helping to get this ready! I’m happy with the code changes now and don’t have anything more to add (tests and man page docs are done, and I updated the PR description and title). I think this is good to merge.

I will have a think about the server state callback and raise another PR shortly for the observability changes. Thanks.

Issue #642, opened 2023-11-30
Feature: domain-specific DNS servers


bradh352:
Its not uncommon for VPNs to advertise a domain-specific DNS server to use via the VPN connection, where queries for that domain (or domains) will go to the server through the VPN, but then any other queries will go to the normal DNS servers.  Often these servers may be configured as non-recursive so sending other queries to them will not work.

c-ares has no means to handle that at this point in time.

OS-specific information:
 - MacOS, can be extracted via SC (System Configuration) ... details to follow
 - Windows. ???
 - Linux (not in resolv.conf, but Network Manager does have it).  Need to determine if we send dbus requests to retrieve info or read NetworkManager configs directly ... or ... ?

bradh352:
@oliverwelsh hate to pull you in on this, but noticed you work for Microsoft and have contributed a significant amount lately to c-ares.  Do you happen to know the proper way to pull domain-specific server lists from Windows?  Or do you have a contact that you could bother to confirm if GetInterfaceDnsSetting() might be on the right path?  

I think this is pretty important and common for VPNs to do this sort of thing, so I think its a major lacking area for c-ares vs the system resolver and I should probably tackle this soon-ish.

oliverwelsh:
> @oliverwelsh hate to pull you in on this, but noticed you work for Microsoft and have contributed a significant amount lately to c-ares. Do you happen to know the proper way to pull domain-specific server lists from Windows? Or do you have a contact that you could bother to confirm if GetInterfaceDnsSetting() might be on the right path?
> 
> I think this is pretty important and common for VPNs to do this sort of thing, so I think its a major lacking area for c-ares vs the system resolver and I should probably tackle this soon-ish.

@bradh352 - Apologies but I'm not familiar with domain-specific server lists on Windows and don't have any contacts within Microsoft who I think would be; the area I work in is almost all Linux-based.

I searched for `GetInterfaceDnsSetting` to see if I could find any additional information, but only the public-facing documentation showed up (https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-getinterfacednssettings).

bradh352:
> @bradh352 - Apologies but I'm not familiar with domain-specific server lists on Windows and don't have any contacts within Microsoft who I think would be; the area I work in is almost all Linux-based.

Alright, thanks, worth a try :)

PR #745, opened 2024-04-16
Fix support of link-local IPv6 DNS server in /etc/resolv.conf on macOS.

shpala:
Fix support of link-local IPv6 DNS server in /etc/resolv.conf on macOS.
The issue only happens when using system default link-local IPv6 DNS. Specifying link-local IPv6 DNS address manually works.
Disabling libresolv and using the manual parsing fixes the issue after the change made in #646

On macOS c-ares uses libresolv and assumes that its API res_getservers() fills sin6_scope_id in case of a link-local IPv6 DNS addresses in /etc/resolv.conf.

Unfortunately this is not the case.
See res_getservers() here: https://opensource.apple.com/source/libresolv/libresolv-60/res_init.c.auto.html


bradh352:
as far as I know, iOS does not have `/etc/resolv.conf` at all, or if it does, its not readable by applications as at one point we had to use libresolv.  Also, libresolv first tries calling `dns_configuration_copy()` then falls back to reading `/etc/resolv.conf`, so likely it never reads from `/etc/resolv.conf` at all, but it is true it looks like it has a bug not reading `if_index` which is indeed populated by `dns_configuration_copy()` which can be found here:

https://opensource.apple.com/source/configd/configd-596.15/dnsinfo/dnsinfo_copy.c.auto.html

The proper solution would be to instead use the functions as found in the above implementation to read the system configuration.   What's interesting is its using a completely different method to fetch DNS config than what I had previously found and sort of put in #642 for reading DNS servers.  I wonder which is better...

shpala:
@bradh352 You're right. I missed that part in iOS.
For now modified the PR to only affect macOS.
I'll look at the real solution you proposed.
Thanks!

bradh352:
I played around with the System Configuration calls `SCDynamicStoreCreate()` and friends and can successfully retrieve all the appropriate DNS information that c-ares currently uses.  However, there's a major caveat, its not able to see the domain-specific resolvers that are used by some VPNs or if a user places a custom config in `/etc/resolver/` which is what #642 is all about which is a shame as the API is much easier to use than the stuff found in the configd source code.

I'll need to play around with the code that utils like `scutil` and `libresolv` use to retrieve the info to see how slim I can get it.

bradh352:
oh wow, this is even more interesting, looks like there is some `macosx.internal` SDK that is required in order to build `libresolv`.

It looks like google chrome works around this by cloning their own `dnsinfo.h`: https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/apple_apsl/dnsinfo.h

And then using `dlopen()` to manually get the symbol for `dns_configuration_copy()`: https://chromium.googlesource.com/chromium/src/+/HEAD/net/dns/dns_config_watcher_mac.cc#32

So we'll probably need to take that same path.

bradh352:
I have a working prototype now, I'll try to get up a proper PR for this tomorrow.

bradh352:
closing in preference of PR #750

PR #747, opened 2024-04-17
Improve reliability in the server retry delay regression tests

oliverwelsh:
Improve reliability in the server retry delay regression tests by increasing the retry delay and sleeping for a little more than the retry delay when attempting to force retries.

This helps to account for unreliable timing (e.g. NTP slew) intermittently breaking pipelines.

### DIFF
--- a/test/ares-test-mock-ai.cc
+++ b/test/ares-test-mock-ai.cc
@@ -731,7 +731,7 @@ class ServerFailoverOptsMockTestAI : public MockMultiServerChannelTestAI {
   static struct ares_options* FillOptions(struct ares_options *opts) {
     memset(opts, 0, sizeof(struct ares_options));
     opts->server_failover_opts.retry_chance = 1;
-    opts->server_failover_opts.retry_delay = 100;
+    opts->server_failover_opts.retry_delay = 250;
     return opts;
   }
  private:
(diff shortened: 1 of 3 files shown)


oliverwelsh:
@bradh352 - This is my proposed fix for the unreliable retry delay regression tests, as discussed in https://github.com/c-ares/c-ares/pull/744.

I'm hoping this will help reliability. If issues still occur then please ping me and I will be more than happy to investigate further.

bradh352:
What about doubling or quadrupling the retry delay so the timing doesn't have to be quite so precise.  I think on these CI/CD systems, they are horribly oversubscribed so you may not get a time slice when you might expect.

bradh352:
i drastically reduced some other timings for tests, but only to a level that made sense for various CI/CD systems, i'd rather the timeout be longer than some distribution hitting a test case failure and not accepting a package.  Timing related issue are a real pain for test  cases, kind of have to be more conservative than you think

oliverwelsh:
> i drastically reduced some other timings for tests, but only to a level that made sense for various CI/CD systems, i'd rather the timeout be longer than some distribution hitting a test case failure and not accepting a package. Timing related issue are a real pain for test cases, kind of have to be more conservative than you think

Ok thanks, yep I agree that having intermittent failures is a lot more expensive than having slightly longer tests. I have increased the retry delay from 100ms to 250ms (matching the default timeout set in https://github.com/c-ares/c-ares/blob/main/test/ares-test.cc#L738-L742).

Issue #753, opened 2024-05-08
Use of SOCK_DNS extension on OpenBSD breaks apps


brad0:
With the introduction of the following PR https://github.com/c-ares/c-ares/pull/659 it has broken at least Node and maybe some other apps utilizing c-ares.

We only ran into this recently when updating our port / package from 1.19.1 to 1.27.

https://marc.info/?t=171512363600002&r=1&w=2

Theo said the usage of this extension is not as intended and inappropriate.

PR #754, opened 2024-05-08
Revert PR #659

VlkrS:
As described in https://github.com/c-ares/c-ares/issues/753 the change merged with https://github.com/c-ares/c-ares/pull/659 is ill-advised and leads to breakage in applications using c-ares on OpenBSD. 

See also https://github.com/nodejs/node/issues/52439

### DIFF
--- a/src/lib/ares__socket.c
+++ b/src/lib/ares__socket.c
@@ -253,12 +253,6 @@ ares_status_t ares__open_connection(ares_channel_t      *channel,
   struct server_connection *conn;
   ares__llist_node_t       *node;
   int                       type = is_tcp ? SOCK_STREAM : SOCK_DGRAM;
-#ifdef __OpenBSD__
-  if ((is_tcp && server->tcp_port == 53) ||
-      (!is_tcp && server->udp_port == 53)) {
-    type |= SOCK_DNS;
-  }
-#endif
 
   switch (server->addr.family) {
     case AF_INET:


bradh352:
Very interesting, I didn't evaluate SOCK_DNS when I accepted the original PR other than to check the relevant manpage that was linked in the PR which made no mention about its real use case or restrictions on its use.

It sounds like its an extension that is only allowed to be used by libc and not external 3rd parties?  Odd.

Its unfortunate Theo replied in such a manner, however.  As a maintainer looking in from the outside, I'm not sure how I could have caught this not being an OpenBSD user myself when the documentation it woefully incomplete.  Also, the fact that no CI system seems to natively support OpenBSD means testing of this can't be automated.

VlkrS:
> Its unfortunate Theo replied in such a manner, however. As a maintainer looking in from the outside, I'm not sure how I could have caught this not being an OpenBSD user myself 

Thanks for merging this so quickly.
I can't speak for anyone else, but as far as I am concerned, I see both your point and how this appeared to be  a good idea to the person making the PR.

bradh352:
i actually found an alternative with cross-platform-actions and added minimal CI for OpenBSD and NetBSD

catap:
> i actually found an alternative with cross-platform-actions and added minimal CI for OpenBSD and NetBSD

I'm quite interesting in alternative solution. May you share the link? :)

bradh352:
https://github.com/c-ares/c-ares/blob/main/.github/workflows/openbsd.yml

https://github.com/c-ares/c-ares/blob/main/.github/workflows/netbsd.yml

catap:
On Wed, 08 May 2024 22:45:59 +0100,
Brad House ***@***.***> wrote:
> https://github.com/c-ares/c-ares/blob/main/.github/workflows/openbsd.yml
> 
> https://github.com/c-ares/c-ares/blob/main/.github/workflows/netbsd.yml
>

Thanks!

-- 
wbr, Kirill

PR #758, opened 2024-05-18
const is fine on ares__channel_[un]lock

dimbleby:
at https://github.com/c-ares/c-ares/pull/601#issuecomment-1801935063 you chose not to scatter `const` on the public interface because of the plan - now realised - to add threading to c-ares, and in the expectation that even read operations would need to lock the mutex.

But the threading implementation has a _pointer_ to a mutex inside the ares channel and as I understand it, that means that it is just fine to mark `ares__channel_lock` (and `ares__channel_unlock`) as taking a `const` channel.  It is the pointed-to mutex that is not constant, but C does not propagate `const`-ness through pointers.

Anyway here is a pull request limited to just that change, in the expectation that the pipelines will say so if I have misunderstood this.  

This is not necessarily worth merging for itself, rather it is intended as a demonstration that - if you are still interested in it - #601 can be back on the table.

### DIFF
--- a/docs/ares_dup.3
+++ b/docs/ares_dup.3
@@ -9,7 +9,7 @@ ares_dup \- Duplicate a resolver channel
 .nf
 #include <ares.h>
 
-int ares_dup(ares_channel_t **\fIdest\fP, ares_channel_t *\fIsource\fP)
+int ares_dup(ares_channel_t **\fIdest\fP, const ares_channel_t *\fIsource\fP)
 .fi
 .SH DESCRIPTION
 The \fBares_dup(3)\fP function duplicates an existing communications channel
@@ -23,4 +23,3 @@ handle when the channel is no longer needed.
 .BR ares_library_init (3)
 .SH AVAILABILITY
 \fIares_dup(3)\fP was added in c-ares 1.6.0
-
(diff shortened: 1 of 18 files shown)


bradh352:
you're right, the thread implementation ended up mallocing a memory area for the OS-specific implementations rather than trying to create a system-specific typedef, such as to pthread_mutex_t (which would have made the original decision correct).

bradh352:
did you want to go ahead and sprinkle const on the relevant public functions with this PR too?

dimbleby:
sure, I can do that - might take me a day or two to get to it

dfandrich:
IIRC adding const changes the ABI on certain (probably obscure) platforms, but
it's possible my memory is failing me on that point.

bradh352:
@dfandrich are you sure?  I mean on C++ it might change name mangling, but that's not a concern with C.

dfandrich:
I have a vague recollection of problems with this years ago on some obscure
system and I can't for certain any more. It will be a few days before I can get
a chance to verify it, so go ahead if your information says otherwise.

dimbleby:
updates pushed, pipeline green

looks as though I am using a more recent version of clang-format than was previously used, but since the pipeline doesn't mind I have just let it do its thing.

PR #759, opened 2024-05-19
Auto reload config on changes (requires EventThread)

bradh352:
Automatically detect configuration changes and reload.  On systems which provide notification mechanisms, use those, otherwise fallback to polling.


bradh352:
@bagder any comments/questions/concerns with this?   I'm just enabling it always if EventThread is enabled.

PR #765, opened 2024-05-24
SIG(0) record class

nodech:
I am using c-ares with a resolver that supports: SIG(0) record class, according to https://datatracker.ietf.org/doc/html/rfc2931#section-3

> For all SIG(0) RRs, the owner name, class, TTL, and original TTL, are
meaningless. The TTL fields SHOULD be zero and the CLASS field
SHOULD be ANY.

--

This PR enables `ANY` records. Alternatively we could have different method that accepts type + rclass to do class validation and that could narrow ANY + SIG.

### DIFF
--- a/src/lib/ares_dns_record.c
+++ b/src/lib/ares_dns_record.c
@@ -412,7 +412,7 @@ ares_status_t ares_dns_record_rr_add(ares_dns_rr_t    **rr_out,
   if (dnsrec == NULL || name == NULL || rr_out == NULL ||
       !ares_dns_section_isvalid(sect) ||
       !ares_dns_rec_type_isvalid(type, ARES_FALSE) ||
-      !ares_dns_class_isvalid(rclass, ARES_FALSE)) {
+      !ares_dns_class_isvalid(rclass, ARES_TRUE)) {
     return ARES_EFORMERR;
   }
 


bradh352:
yeah, i think we'd want to narrow it to catch malformations at the parser level.  I'll dig into the SIG(0) spec a bit as I'm not familiar with it...

bradh352:
@nodech can you tell me if #773 fixes your issue.  I don't actually have a means to test this myself against a real server at this time.

bradh352:
@nodech can you also tell me what DNS server supports SIG(0) records for signing responses and a link to the docs on configuring it.  I've tried looking at both unbound and bind, but the only mentions of it appear to be for dynamic updates, not for general replies ...

nodech:
Thanks for the fix ! #773 seem to work fine.

The project uses https://github.com/chjj/bns. That is part of the https://github.com/handshake-org/hsd.

I have not configured bns separately, but for testing it's simple to run `hsd`:
  - `./bin/hsd --network=regtest --memory` - for regtest dns ports are these:
    - `root server` - 25349
    - `recursive server` - 25350
I used `./build/adig` for simple tests (and tested this PR with node.js as well)

Using #773:
```
> ./build/bin/adig -s 127.0.0.1 -U 25349 -t SOA . 

; <<>> c-ares DiG 1.29.0 <<>> .
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 26484
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 3

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: 0; udp: 4096
;; QUESTION SECTION:
;.                      IN      SOA

;; ANSWER SECTION:
.               86400   IN      SOA     . . 2024060312 1800 900 604800 21600

;; ADDITIONAL SECTION:
_fs00008._synth.                518400  IN      A       127.0.0.1
.               0       ANY     SIG     0 253 0 0 1717440590 1717397390 64595 . ddda1a91926763b64914981acd89206fa929b3787b3eb3b258966771c3a723ea2124e2c0790db7b63e5be81c4a5d7df3363eae00b3cfec30516610a28786d9b5

;; AUTHORITY SECTION:
.               518400  IN      NS      _fs00008._synth.

;; MSG SIZE  rcvd: 199
```

bradh352:
Oh interesting, this is using some sort of private algorithm (algorithm = 253) implementation.  Gotcha.  I'll go ahead and get that PR #773 merged.

PR #775, opened 2024-06-06
Start annotating code for coverage exclusions

bradh352:
Defensive Coding and Out Of Memory are often lines of code that can't be covered.  Annotate the code to add exclusions for these.

### DIFF
--- a/src/lib/ares__buf.c
+++ b/src/lib/ares__buf.c
@@ -87,7 +87,7 @@ void ares__buf_destroy(ares__buf_t *buf)
 static ares_bool_t ares__buf_is_const(const ares__buf_t *buf)
 {
   if (buf == NULL) {
-    return ARES_FALSE;
+    return ARES_FALSE; /* LCOV_EXCL_LINE: defensive coding */
   }
 
   if (buf->data != NULL && buf->alloc_buf == NULL) {
@@ -107,7 +107,7 @@ void ares__buf_reclaim(ares__buf_t *buf)
   }
 
   if (ares__buf_is_const(buf)) {
-    return;
+    return; /* LCOV_EXCL_LINE: defensive coding */
   }
 
   /* Silence coverity.  All lengths are zero so would bail out later but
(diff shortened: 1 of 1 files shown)


dfandrich:
FWIW, many OOM conditions can be checked without a huge amount of difficulty. curl does it with its "torture" tests with help from some compile-time macros, and I've done it in other projects entirely at run-time using failmalloc (which seems to be dead upstream but which I haven't found a good replacement for).

bradh352:
c-ares has some of that memory fail logic built into its test suite.  Personally I'm against even checking for OOM since its meaningless on modern systems for a library that performs relatively small allocations since the system will likely optimistically approve the allocation even if there is really no memory available.  

But since c-ares has always done it, I've made sure any new code checks for return values.  I'm not committed enough to actually enhance the test system and test cases for testing these paths.

I have a blog post on my philosophy :) 
https://bradhouse.dev/2023/11/how-to-handle-out-of-memory-oom-in-c/

bradh352:
I should also mention, its another reason I'm annotating the exclusion lines.  Makes it quick and easy to just use `sed` to kill the exclusions of the same 'class', like OutOfMemory.  So if someone else wants to take it up, its easy :)

bagder:
> Personally I'm against even checking for OOM since its meaningless on modern systems

1. it is used on lots of systems not "modern" as well, where malloc failures are real. Do not do the mistake of thinking that the entire world runs like Linux.
2. users, such as me, would not build on top of libraries that cannot fail and return error correctly on OOM. We don't think rust-based libraries are for real use either until they can return error on OOM.

(I think this kind of annotation makes the code ugly and harder to read. If there was a vote, I would vote against it)

bradh352:
The code has had annotations ever since someone enabled coverage support, just with the major rework of the code I wasn't keeping up with re-adding the annotations.  I guess the only change is I'm actively tagging branches that are out of memory conditions as to be skipped in the coverage calculation.  My goal is to only leave branches as uncovered where we *need* test cases to validate so we can focus on better test case coverage.  If we focus on out of memory branches that are not the likely path, or possibly not even a path that *could* be taken on most systems, we'd be doing a vast majority of hte users a disservice. 

Like I said, someone with a particular eye for OutOfMemory test cases can easily sed out the annotations for that case and address those.  I definitely wouldn't turn away a PR adding test cases that cover those.

bagder:
My experience from having and doing extensive OOM testing in curl for decades, is that it helps for a lot more than just OOM errors: it also helps in generally testing the exit paths for errors in many more places than other tests cover (easily). It basically inserts a lot more ways for error returns, with the accompanying verification that everything can return errors fine from those places - without leaking memory or otherwise crashing. IOW: it has helped making a more reliable library.

(and I dislike the annotations no matter how they came to be)

PR #779, opened 2024-06-07
fix some obvious errors reported by the CLion Project Analyzer

gjasny:


bradh352:
looks reasonable, if not minor :)

Issue #780, opened 2024-06-09
WATT-32 support is broken


bradh352:
Moving this discussion from comments:

>>> @gvanem
>>> So all this code needs Winsock2, no? So why not check for
>>> #ifdef USE_WINSOCK or #if (USE_WINSOCK == 2).
>>> Currently this code breaks with USE_WATT32.

>> @bradh352
>> This wasn't an intentional thing, just something more out of ignorance. I honestly don't know anything about WATT32. If you could give me some information on basic building instructions I can get this stuff cleaned up and also add a CI/CD workflow to prevent issues like this from cropping up again.

> @gvanem
> But I've no idea about CI for C-ares, but I have Appveyour [CI jobs](https://ci.appveyor.com/project/gvanem/watt-32) for Watt-32.
> So I guess building C-ares with -DWATT32 and Windows would need to fork Watt-32 (as a submodule) or use some artifacts from the above CIs (but Watt-32 does not currently contain any artifacts).

@bradh352 
c-ares already uses AppVeyor, all of our Windows builds use it except for OpenWatcom.   So we can probably use some examples from your code, looks like your actual `appveyor.yml` is here: https://github.com/gvanem/Watt-32/blob/master/appveyor.yml

1. What is the target OS you're most concerned with?
 - Win32 or DOS?
2. Which build system or makefile do you use within c-ares to try to build with WATT32 support?
 - I do see `-DWATT32` as an available option for `Makefile.msvc`, that obviously targets Win32.
 - I see `Makefile.watcom` can take a `use_watt32` argument, that also targets Win32. 
 - There's also an old `Makefile.dj` for DJGPP that targets DOS but I'm pretty sure that makefile is broken after a code reorg a few years back.
3. Should we use the official WATT-32 packages in CI, or actually build WATT-32 from Git and link to that?

gvanem:
> What is the target OS you're most concerned with?
> Win32 or DOS?

Win32. I've not tested DOS (i.e. djgpp) and C-ares in some time. 

> 2. Which build system or makefile do you use within c-ares to try to build with WATT32 support?

I use a home-brewed [GNU-makefile](https://github.com/user-attachments/files/15753852/Makefile.Windows.txt) for all the combinations: Winsock2/Watt-32, MSVC/clang-cl, x86/x64 etc.

> I see Makefile.watcom can take a use_watt32 argument, that also targets Win32.

I seldom use Watcom. It's basically a dead product IMHO.

[...]

PR #781, opened 2024-06-10
Fix Watt-32 compilation support

bradh352:
Watt-32 (https://www.watt-32.net/) support has been broken for a long time.  Patch c-ares to fix Watt-32 support and also gets rid of the `WIN32` macro which adds confusion, only use `USE_WINSOCK` macro.

Add a CI/CD task to build c-ares on Windows using MSVC with Watt-32.

Fixes Issue: #780

### DIFF
--- a/src/lib/ares__buf.c
+++ b/src/lib/ares__buf.c
@@ -229,9 +229,9 @@ ares_status_t ares__buf_append(ares__buf_t *buf, const unsigned char *data,
   return ARES_SUCCESS;
 }
 
-ares_status_t ares__buf_append_byte(ares__buf_t *buf, unsigned char byte)
+ares_status_t ares__buf_append_byte(ares__buf_t *buf, unsigned char b)
 {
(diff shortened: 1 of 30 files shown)


src/lib/ares__buf.c
```diff
@@ -229,9 +229,9 @@ ares_status_t ares__buf_append(ares__buf_t *buf, const unsigned char *data,
   return ARES_SUCCESS;
 }
 
-ares_status_t ares__buf_append_byte(ares__buf_t *buf, unsigned char byte)
+ares_status_t ares__buf_append_byte(ares__buf_t *buf, unsigned char b)
```
gvanem, line 232: This could be fixed by adding `-DWATT32_NO_OLDIES` somewhere (in `Makefile.msvc`?).

bradh352: Eh, it was a simple enough change that it doesn't hurt anything.  Easier than needing to document that someone compiling for WATT32 needs to also define that.

Issue #761, opened 2024-05-22
Windows: `NotifyIpInterfaceChange()` does not notify on manual dns config changes


bradh352:
If a system administrator were to manually update the dns configuration, `NotifyIpInterfaceChange()` nor `NotifyUnicastIpAddressChange()` will trigger the event callback.  I would think `NotifyIpInterfaceChange()` would be called with `MibParameterNotification` since an interface parameter changed, but its not.

Perhaps this is a bug in Windows?  There is no means to get a notification otherwise, we'd literally have to just query `GetAdapterAddresses()` to enumerate all the DNS servers every few seconds (in a different thread as who knows how long that request might take), which seems wasteful.

See PR #759 for the implementation of change detection.

Opened a ticket with Microsoft:
https://learn.microsoft.com/en-us/answers/questions/1692011/notifyipinterfacechange-function-(netioapi-h)-does?comment=question#newest-question-comment
Someone replied saying to use RegNotifyChangeKeyValue() to monitor HKLM\SYSTEM\CurrentControlSet\Services\Tcpip6\Parameters\Interfaces and HKLM\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\Interfaces

sykhro:
Using `RegNotifyChangeKeyValue()` broke our builds. we are using c-ares on Xbox, which implements a subset of Win32 lacking this and other functions (the entirety of `winreg.h` I fear, see `WINAPI_PARTITION_GAMES`).

Would you accept a PR which makes this check conditional?

bradh352:
> Using `RegNotifyChangeKeyValue()` broke our builds. We are using c-ares on Xbox, which implements a subset of Win32 lacking this and other functions (the entirety of `winreg.h` I fear, see `WINAPI_PARTITION_GAMES`).
> 
> Would you accept a PR which makes this check conditional?
> 
> _(none of this is covered by NDA, such information is publicly available in the Windows docs)_

Sure. send a PR.  The goal of c-ares is to be as portable as possible.  Not sure how to build for xbox, but if possible we should also add a CI/CD target to auto-build to make sure we don't regress in the future.

sykhro:
> > Using `RegNotifyChangeKeyValue()` broke our builds. We are using c-ares on Xbox, which implements a subset of Win32 lacking this and other functions (the entirety of `winreg.h` I fear, see `WINAPI_PARTITION_GAMES`).
> > Would you accept a PR which makes this check conditional?
[...]

PR #785, opened 2024-06-14
Windows: Reload config on manual DNS changes

bradh352:
The `NotifyIpInterfaceChange()` method only notifies of automatic changes, such as via DHCP or interfaces going up or down.  However, someone editing the DNS settings manually would not trigger a notification. 

Use `RegNotifyChangeKeyValue()` to monitor `HKLM\SYSTEM\CurrentControlSet\Services\Tcpip6\Parameters\Interfaces` and `HKLM\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\Interfaces` to pick up these changes.

This new check also works when using the OpenWatcom compiler as the symbols are available, so they have been enabled (though `NotifyIpInterfaceChange()` method is disabled for OpenWatcom still due to the linker libraries missing this symbol).

NOTE: it does appear that about 5 events get triggered for some reason, it doesn't seem to hurt anything, but it will actually cause it to re-read the configuration multiple times.

Fixes Issue: #761 

### DIFF
--- a/src/lib/ares_event_configchg.c
+++ b/src/lib/ares_event_configchg.c
@@ -311,7 +311,7 @@ ares_status_t ares_event_configchg_init(ares_event_configchg_t **configchg,
    *       that didn't get triggered either.
    */
   if (NotifyIpInterfaceChange(
-        AF_UNSPEC, (PIPINTERFACE_CHANGE_CALLBACK)ares_event_ipchg_cb,
+        AF_UNSPEC, (PIPINTERFACE_CHANGE_CALLBACK)ares_event_configchg_ip_cb,
         *configchg, FALSE, &c->ifchg_hnd) != NO_ERROR) {
     status = ARES_ESERVFAIL;
     goto done;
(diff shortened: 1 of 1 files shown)


dg0yt:
These changes seem to have broken UWP builds. Is this intentional? 
Observed in vcpkg (https://github.com/microsoft/vcpkg/pull/40213) and triggering downstream changes there. (I think I also saw existin packages avoiding c-ares for UWP.)

bradh352:
definitely not intentional.  We had added an AppVeyor CI build for that, but then there was some later breakage due to an MSVC bug ( https://developercommunity.visualstudio.com/t/clexe-compiler-error-C1007-when-compili/10486219 ) so we had to disable it until AppVeyor upgraded, so UWP wasn't being tested at all.

I'll see if we can switch to GitHub Actions for the UWP build to see if we can fix things there.

bradh352:
fixed in 3fd5925

Issue #806, opened 2024-07-04
Feature: rework c-ares initialization functions


bradh352:
c-ares uses `ares_init_options()` and `ares_save_options()` which has significant ABI issues for setting and retrieving the various configuration values since it relies on a single structure which may change from release to release.  There also isn't a way to get all possible configuration set, due to a mix of system configuration and configuration values set by the integrator.

We need to drop the `ares_init_options()` and `ares_save_options()` concepts completely and switch to a new paradigm, though.  we need to ensure that we can retain `ares_init_options()` and `ares_save_options()` as lightweight wrappers around the new API so we don't break existing integrations.

Likely we need to make a new function like `ares_create()` to initialize the channel object, then various `ares_set_*()` and `ares_get_*()` functions to set and get the possible values.  The `ares_get_*()` routines should also be able to indicate if the value was set by the system configuration or by the integrator.

The setters should be allowed to be used at any time, even after queries have been processed so that configuration can be dynamically updated if needed.  Obviously some options could have a significant impact on processing, such as changing socket state callbacks.

PR #832, opened 2024-07-25
struct areas_options: reduce padding

mykaul:
Moved nservers parameter to reduce padding.

Before:
```
struct ares_options {
        int                        flags;                /*     0     4 */
        int                        timeout;              /*     4     4 */
        int                        tries;                /*     8     4 */
        int                        ndots;                /*    12     4 */
        short unsigned int         udp_port;             /*    16     2 */
        short unsigned int         tcp_port;             /*    18     2 */
        int                        socket_send_buffer_size; /*    20     4 */
        int                        socket_receive_buffer_size; /*    24     4 */

        /* XXX 4 bytes hole, try to pack */

        struct in_addr *           servers;              /*    32     8 */
        int                        nservers;             /*    40     4 */

        /* XXX 4 bytes hole, try to pack */

        char * *                   domains;              /*    48     8 */
        int                        ndomains;             /*    56     4 */

        /* XXX 4 bytes hole, try to pack */

        /* --- cacheline 1 boundary (64 bytes) --- */
        char *                     lookups;              /*    64     8 */
        ares_sock_state_cb         sock_state_cb;        /*    72     8 */
        void *                     sock_state_cb_data;   /*    80     8 */
        struct apattern *          sortlist;             /*    88     8 */
        int                        nsort;                /*    96     4 */
        int                        ednspsz;              /*   100     4 */
        char *                     resolvconf_path;      /*   104     8 */
        char *                     hosts_path;           /*   112     8 */
        int                        udp_max_queries;      /*   120     4 */
        int                        maxtimeout;           /*   124     4 */
        /* --- cacheline 2 boundary (128 bytes) --- */
        unsigned int               qcache_max_ttl;       /*   128     4 */
        ares_evsys_t               evsys;                /*   132     4 */

        /* size: 136, cachelines: 3, members: 24 */
        /* sum members: 124, holes: 3, sum holes: 12 */
        /* last cacheline: 8 bytes */
```
After:
```
struct ares_options {
        int                        flags;                /*     0     4 */
        int                        timeout;              /*     4     4 */
        int                        tries;                /*     8     4 */
        int                        ndots;                /*    12     4 */
        short unsigned int         udp_port;             /*    16     2 */
        short unsigned int         tcp_port;             /*    18     2 */
        int                        socket_send_buffer_size; /*    20     4 */
        int                        socket_receive_buffer_size; /*    24     4 */
        int                        nservers;             /*    28     4 */
        struct in_addr *           servers;              /*    32     8 */
        char * *                   domains;              /*    40     8 */
        int                        ndomains;             /*    48     4 */

        /* XXX 4 bytes hole, try to pack */

        char *                     lookups;              /*    56     8 */
        /* --- cacheline 1 boundary (64 bytes) --- */
        ares_sock_state_cb         sock_state_cb;        /*    64     8 */
        void *                     sock_state_cb_data;   /*    72     8 */
        struct apattern *          sortlist;             /*    80     8 */
        int                        nsort;                /*    88     4 */
        int                        ednspsz;              /*    92     4 */
        char *                     resolvconf_path;      /*    96     8 */
        char *                     hosts_path;           /*   104     8 */
        int                        udp_max_queries;      /*   112     4 */
        int                        maxtimeout;           /*   116     4 */
        unsigned int               qcache_max_ttl;       /*   120     4 */
        ares_evsys_t               evsys;                /*   124     4 */

        /* size: 128, cachelines: 2, members: 24 */
        /* sum members: 124, holes: 1, sum holes: 4 */
```

### DIFF
--- a/include/ares.h
+++ b/include/ares.h
@@ -377,8 +377,8 @@ struct ares_options {
   unsigned short tcp_port; /* host byte order */
   int            socket_send_buffer_size;
   int            socket_receive_buffer_size;
-  struct in_addr    *servers;
   int                nservers;
+  struct in_addr    *servers;
   char             **domains;
   int                ndomains;
   char              *lookups;


dimbleby:
presumably ABI compatibility will make this a no go

seems surprising that you would have so many `ares_options` structures that eight bytes could make a difference?

dimbleby:
> I'm not sure what ABI I'm breaking here.

code compiled against c-ares version N will be broken when distributions roll out c-ares version N+1

but I am just a drive-by interested party: I'm _fairly_ sure that I'm right but if you are keen on this then you should wait for maintainer comment!

bradh352:
This would definitely break ABI.  The problem is with dynamic linking, an application may be compiled against c-ares 1.17, but then the OS is upgraded to a later c-ares version. That structure is public, and gets passed from one c-ares version to a different one.  At the binary level the structure is used to determine the offset from the start of the memory address to where the data to be retrieved resides, if that offset is different from version to version, the wrong data is going to be read.

The only change we can make is to append members to the structure, and the only reason this works is there is an `optmask` parameter, we can reasonably assume any well-behaved application using c-ares wouldn't set bits in `optmask` that weren't defined in their version of c-ares, so c-ares won't read outside the memory bounds of the structure pointer the application provided since it won't try to read parameters outside the `optmask` provided.

`ares_init_options` is honestly one of those design flaws in c-ares though because if there are members that need memory allocation, `ares_destroy_options` can't clean them up without breaking ABI so there are things we can't do to expand the structure.

I've got #806 open to at some point work on.

Issue #837, opened 2024-07-29
AttachCurrentThread spoils current thread mame


Hsilgos:
This is known "Side effect" of the function. After call current thread is renamed to "Thread-xx".
The solution - get current thread name and pass it through `JavaVMAttachArgs` as 3rd parameter to `AttachCurrentThread`.
Fix proposal is upcoming.

PR #838, opened 2024-07-29
Preserve thread name after attaching it to JVM

Hsilgos:
When native thread is attached to JVM, then it's name is taken from JavaVMAttachArgs. When no JavaVMAttachArgs, or no JavaVMAttachArgs::name passed, then JVM on its own decides on taming thread. Those names are not descriptive. To preserve thread name, pass the currently set thread name in JavaVMAttachArgs::name. pthread_getname_np was introduced in API 26, hence use more generic approach.

Fixes: #837

### DIFF
--- a/src/lib/ares_android.c
+++ b/src/lib/ares_android.c
@@ -26,6 +26,7 @@
 #if defined(ANDROID) || defined(__ANDROID__)
 #  include "ares_private.h"
 #  include <jni.h>
+#  include <sys/prctl.h>
 
 static JavaVM   *android_jvm                  = NULL;
 static jobject   android_connectivity_manager = NULL;
@@ -80,6 +81,25 @@ static jmethodID jni_get_method_id(JNIEnv *env, jclass cls,
   return mid;
 }
 
+static const char *get_thread_name()
+{
+    static __thread char name[17] = {0};
+    if (prctl(PR_GET_NAME, name) == 0)
+    {
+        return name;
+    }
+    return NULL;
+}
+
+static JavaVMAttachArgs get_jvm_attach_args()
+{
+    JavaVMAttachArgs args;
+    args.version = JNI_VERSION_1_6;
+    args.name = get_thread_name();
+    args.group = NULL;
+    return args;
+}
+
 void ares_library_init_jvm(JavaVM *jvm)
 {
   android_jvm = jvm;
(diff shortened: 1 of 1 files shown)


src/lib/ares_android.c
```diff
@@ -80,6 +81,25 @@ static jmethodID jni_get_method_id(JNIEnv *env, jclass cls,
   return mid;
 }
 
+static const char *get_thread_name()
+{
+    static __thread char name[17] = {0};
```
bradh352, line 86: not a big fan of using ThreadLocalStorage like this unless its necessary.  I can't find in the docs if the name you pass is required to exist for the life of the thread or not.  If it does, then so be it, its necessary.

Otherwise, really, you should just create a function like: `int jvm_attach(JNIEnv *env);`
That does both the `get_thread_name()` and `get_jvm_attach_args()` internally as one function along with calling `return (*android_jvm)->AttachCurrentThread(android_jvm, env, &args);` or whatever.  Replacing all direct calls to AttachCurrentThread with that new function.

bradh352 (changes requested):

bradh352:
also, not sure if you'd be interested in maybe helping out with creating a github action for building the Android library.  Currently our CI system isn't building android to validate PRs like this. :)

src/lib/ares_android.c
Hsilgos: Refactored

Hsilgos:
> also, not sure if you'd be interested in maybe helping out with creating a github action for building the Android library. Currently our CI system isn't building android to validate PRs like this. :)

I may have a look at weekend

src/lib/ares_android.c
```diff
@@ -80,6 +81,30 @@ static jmethodID jni_get_method_id(JNIEnv *env, jclass cls,
...
+  return args;
+}
+
+static int jvm_attach(JNIEnv **env)
+{
+  char             name[17] = {0};
+  JavaVMAttachArgs args     = get_jvm_attach_args(name);
+  return (*android_jvm)->AttachCurrentThread(android_jvm, env, &args);
+}
+
```
bradh352, line 107: honestly, I'd just combine jvm_attach(), get_jvm_attach_args(), and get_thread_name() into a single function.  They're only used together and it doesn't really enhance readability.  But otherwise, yeah, this is good.

Hsilgos: Combined

bradh352:
looks good to me.  If you think you can add a GitHub Action to do a test build for android, I'll wait on that and accept it as one PR ...

bradh352 (approved):

bradh352:
I ended up adding a GitHub action, so I'll go ahead and merge this.  Thanks!

PR #840, opened 2024-07-29
Implement TCP FastOpen (TFO) RFC7413

bradh352:
In Process. Draft.


cd606:
Hello, this update broken my vcpkg build on CentOS 7, which has an old kernel version that does not support TCP_FASTOPEN_CONNECT. Is it possible to get this working for these kind of older kernels? Thanks.

bradh352:
> Hello, this update broken my vcpkg build on CentOS 7, which has an old kernel version that does not support TCP_FASTOPEN_CONNECT. Is it possible to get this working for these kind of older kernels? Thanks.

already fixed in 4abdf7d

PR #844, opened 2024-08-01
Adds android GitHub action and gradle scripts

Hsilgos:


Hsilgos:
Oh... I see you don't need it anymore :-)

bradh352:
Do you think there is any advantage to the gradle methodology for what we're attempting?  If there is, such as maybe the possibility for better integration to be able to run it in the emulator and run tests or something, that would be great.

What I added is a really basic set of builds, but building is only 10% of the problem, it would be great to be able to  verify it actually works.

bradh352:
I just looked at react native and see you're right, API level 23 is their minimum as well.  I just changed the current CI workflow to use that.

jekkyll76 (approved):

PR #856, opened 2024-08-12
adig: read arguments from adigrc

tacerus:
This allows for default arguments to be provided through a user configuration file, useful for avoiding repetitive passing of flags on the command line when using adig frequently. The syntax within $XDG_CONFIG_HOME/adigrc resembles the one from dig's ~/.digrc, meaning any command line arguments can be provided, separated either by line breaks or whitespaces.

### DIFF
--- a/src/tools/adig.c
+++ b/src/tools/adig.c
@@ -137,14 +137,57 @@ static void print_help(void)
     "              SOA, SRV, TXT, TLSA, URI, CAA, SVCB, HTTPS\n\n");
 }
 
+static int split_rcline(char *rcline, char **rcargv, int len) {
+  int   rcargc;
+  char *last = NULL;
+
+  for (rcargc = 1, rcargv[rcargc] = strtok_r(rcline, " \n", &last);
+    rcargc < len && rcargv[rcargc];
+    rcargv[++rcargc] = strtok_r(NULL, " \n", &last))
+  {
+    DEBUGF(fprintf(stderr, "split_rcline() argc: %d, argv: %s\n", rcargc, rcargv[rcargc]));
+  }
+
+  return (rcargc);
+}
+
 static ares_bool_t read_cmdline(int argc, const char * const *argv,
-                                adig_config_t *config)
+                                adig_config_t *config, ares_bool_t is_rcfile)
 {
-  ares_getopt_state_t state;
-  int                 c;
+  ares_getopt_state_t  state;
+  int                  c;
+  char                *homedir;
 
   ares_getopt_init(&state, argc, argv);
   state.opterr = 0;
+
+  if (!is_rcfile) {
+    homedir = getenv("HOME");
+    if (homedir != NULL) {
+      char            rcfile[256];
+      static FILE    *rcdata = NULL;
+      char            rcline[256];
+
+      unsigned int    rclen;
+      unsigned int    rcsize = sizeof(rcfile);
+      int             rcargc;
+      char           *rcargv[64];
+
+      rclen = (unsigned int) snprintf(rcfile, rcsize, "%s/.adigrc", homedir);
+      if (rclen < rcsize) {
+        rcdata = fopen(rcfile, "r");
+
+        if (rcdata != NULL) {
+          while (fgets(rcline, sizeof(rcline), rcdata) != 0) {
+            rcargc = split_rcline(rcline, rcargv, 62);
+            read_cmdline(rcargc, (const char * const *)rcargv, config, ARES_TRUE);
+          }
+
+          fclose(rcdata);
+        }
+      }
+    }
+  }
 
   while ((c = ares_getopt(&state, "dh?f:s:c:t:T:U:")) != -1) {
     int f;
@@ -933,7 +976,7 @@ int main(int argc, char **argv)
   memset(&config, 0, sizeof(config));
   config.qclass = ARES_CLASS_IN;
   config.qtype  = ARES_REC_TYPE_A;
-  if (!read_cmdline(argc, (const char * const *)argv, &config)) {
+  if (!read_cmdline(argc, (const char * const *)argv, &config, ARES_FALSE)) {
     printf("\n** ERROR: %s\n\n", config.error);
     print_help();
     rv = 1;
--- a/src/tools/adig.c
+++ b/src/tools/adig.c
@@ -149,70 +149,12 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
 {
   ares_getopt_state_t  state;
   int                  c;
-  char                 configdir[256];
-  char                *configdir_xdg;
-  unsigned int         cdlen = 0;
-  unsigned int         cdsize = sizeof(configdir);
-  char                *homedir;
 
   ares_getopt_init(&state, argc, argv);
   state.opterr = 0;
 
   if (is_rcfile) {
     state.optind = 0;
-  } else {
-
-#if defined(WIN32)
-    cdlen = (unsigned int) snprintf(configdir, cdsize, "%s/%s", getenv("APPDATA"), "c-ares");
-
-#elif defined(__APPLE__)
-    homedir = getenv("HOME");
-    if (homedir != NULL) {
-      cdlen = (unsigned int) snprintf(configdir, cdsize, "%s/%s/%s/%s", homedir, "Library", "Application Support", "c-ares");
-    }
-
-#else
-    configdir_xdg = getenv("XDG_CONFIG_HOME");
-
-    if (configdir_xdg == NULL) {
-      homedir = getenv("HOME");
-      if (homedir != NULL) {
-        cdlen = (unsigned int) snprintf(configdir, cdsize, "%s/%s", homedir, ".config");
-      }
-    }
-    else {
-      cdlen = (unsigned int) snprintf(configdir, cdsize, "%s", configdir_xdg);
-    }
-
-#endif
-
-    DEBUGF(fprintf(stderr, "read_cmdline() configdir: %s\n", configdir));
-
-    if (cdlen != 0 && cdlen < cdsize) {
-      char            rcfile[256];
-
-      unsigned int    rclen;
-      unsigned int    rcsize = sizeof(rcfile);
-
-      size_t          rcargc;
-      char          **rcargv;
-      ares__buf_t    *rcbuf;
-
-      rclen = (unsigned int) snprintf(rcfile, rcsize, "%s/adigrc", configdir);
-      if (rclen < rcsize) {
-        rcbuf = ares__buf_create();
-        if (ares__buf_load_file(rcfile, rcbuf) == ARES_SUCCESS) {
-          ares__buf_split_str(rcbuf, (const unsigned char *)"\n ", 2, ARES_BUF_SPLIT_TRIM, 0, &rcargv, &rcargc);
-          ares__buf_destroy(rcbuf);
-
-          read_cmdline((int)rcargc, (const char * const *)rcargv, config, ARES_TRUE);
-
-          ares_free_array(rcargv, rcargc, ares_free);
-        } else {
-          DEBUGF(fprintf(stderr, "read_cmdline() failed to load rcfile"));
-        }
-      }
-    }
   }
 
   while ((c = ares_getopt(&state, "dh?f:s:c:t:T:U:")) != -1) {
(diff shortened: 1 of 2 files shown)


tacerus:
I didn't find an existing test suite for adig/ahost to expand for this ..

bradh352:
windows has strtok_s instead of strtok_r

I'm not sure I'm thrilled about its use in the first place though, internally with c-ares we have a whole bunch of library functions for parsing that we've converted to.  The problem is they are only used internally and its not structured in a way to easily be accessible to the apps like ahost and adig.  I'll need to think about that a bit.

Also, likely would need changes for paths for Windows and MacOS since getenv("HOME") probably won't be the right thing.

dfandrich:
The modern place to put rc files is $XDG_CONFIG_HOME, or if XDG_CONFIG_HOME is
not set, ~/.config/. The file should then not start with a dot since it's
already hidden by being inside a dotted directory, so it should be loading
~/.config/adigrc

tacerus:
Thanks for the review.

I added support for strtok_s, but I'm not sure it's the right practice, and whether WIN64 needs to be specified as well (I only find it in one place in the existing code as opposed to WIN32). Any idea, or will the test suite tell?

@dfandrich I definitely agree that `~/.config/` is a better place. I tried to keep it aligned with `dig` ( which searches `$HOME`) but would be happy to diverge in this instance.

I think it would still require finding the right location on Windows/macOS, though.

bradh352:
MacOS would be something like `${HOME}/Library/Application Support/${name}/adig.rc` ... Windows would be in `%APPDATA%` somewhere.

tacerus:
I added a patch to facilitate this, and tested the following variations on Linux:

```
env XDG_CONFIG_HOME=/tmp src/tools/adig
env XDG_CONFIG_HOME="" src/tools/adig
env XDG_CONFIG_HOME=/home/georg/.config src/tools/adig
src/tools/adig
```

I can't test the macOS and Windows conditions.

src/tools/adig.c
```diff
@@ -151,15 +155,88 @@ static void print_help(void)
...
+  char                *homedir;
 
   ares_getopt_init(&state, argc, argv);
   state.opterr = 0;
 
+  if (!is_rcfile) {
+
+#if defined(WIN32)
+    cdlen = (unsigned int) snprintf(configdir, cdsize, "%s", getenv("%%APPDATA%"));
+
```
bradh352, line 190: should just be APPDATA here, no %, the % is just the syntax Windows uses for environment variables in the old cmd.exe shell.  Also need to specify a sub-directory here I'd think like you do for macos.

src/tools/adig.c
```diff
@@ -151,15 +155,88 @@ static void print_help(void)
...
+    if (configdir_xdg == NULL) {
+      homedir = getenv("HOME");
+      if (homedir != NULL) {
+        cdlen = (unsigned int) snprintf(configdir, cdsize, "%s/%s", homedir, ".config");
+      }
+    }
+    else {
+      cdlen = (unsigned int) snprintf(configdir, cdsize, "%s", configdir_xdg);
+    }
+
```
bradh352, line 209: Think we'd need a subdirectory in both these cases as I don't think the xdg config home nor the .config directory usually just have files directly in it, there's usually a per-app directory.

src/tools/adig.c
```diff
@@ -151,15 +155,88 @@ static void print_help(void)
...
+      int             rcargc;
+      char           *rcargv[64];
+
+      rclen = (unsigned int) snprintf(rcfile, rcsize, "%s/adigrc", configdir);
+      if (rclen < rcsize) {
+        rcdata = fopen(rcfile, "r");
+
+        if (rcdata != NULL) {
+          while (fgets(rcline, sizeof(rcline), rcdata) != 0) {
+            rcargc = split_rcline(rcline, rcargv, 62);
```
bradh352, line 230: What happens if the file has binary/non-printable data?

src/tools/adig.c
tacerus: Thanks, will adjust.

src/tools/adig.c
tacerus: In mine there are various files without a subdirectory:

```
$ find ~/.config -maxdepth 1 -type f|wc -l
106
```

The [specification](https://specifications.freedesktop.org/basedir-spec/0.6/) says

> directory relative to which user-specific configuration files should be written

making it sound like files are fine.

But no strong opinion from me.

src/tools/adig.c
tacerus: The help menu will be printed.

dfandrich:
> The specification says
> 
>     directory relative to which user-specific configuration files should be
>     written
> 
> making it sound like files are fine.

I think you're right. If there are going to be a dozen files it might make
sense to put them into a subdirectory, but a single one is fine in ~/.config/
IMHO.

bradh352:
alright, I know this is a pain, but I exposed some of c-ares's internal library functions for use by adig/ahost, and I'd much prefer the parser to be using the `ares__buf_*()` functions, which have inline documentation in the header: https://github.com/c-ares/c-ares/blob/main/src/lib/include/ares__buf.h  (there are also some other semi-public headers in that directory that you can use).

Something similar to this might work (error handling needed):
```c
ares__buf_t *buf;
ares__llist_t *llist;
ares__llist_node_t *node;
ares__array_t *arr;

if (ares__buf_load_file(rcfilename, &buf) != ARES_SUCCESS) {
  /* fail! */
}

/* Split on newlines and normal spaces */
ares__buf_split(buf, "\n ", 2, ARES_BUF_SPLIT_TRIM, 0, &llist);

/* Convert from a linked list of ares__buf_t * objects to a standard C array of strings
 * ... It may be worthwhile to turn this logic into a helper, not sure if there are other
 * usecases.... */
arr = ares__array_create(sizeof(char *), ares_free);
for (node = ares__llist_node_first(llist); node != NULL; node = ares__llist_node_next(node)) {
  char *str = NULL;
  char **ptr;
  ares__buf_t *lbuf = ares__llist_node_val(node);

  ares__buf_fetch_str_dup(lbuf, ares__buf_len(lbuf), &str);
  if (str == NULL)
    continue;
  ares__array_insert_last(&ptr, arr);
  *ptr = str; 
}
ares__llist_destroy(llist);
ares__buf_destroy(buf);
size_t argc;
char **argv = ares__array_finish(arr, &argc);

/* TODO: Free each element of argv when done */
```

bradh352:
FYI, i just made a new PR #867 to make this a bit easier.  Once that PR is merged, the code would be reduced to:

```c
ares__buf_t *buf;
char **argv;
size_t argc;

buf = ares__buf_create();
if (ares__buf_load_file(rcfilename, buf) != ARES_SUCCESS) {
  /* fail! */
}

/* Split on newlines and normal spaces */
ares__buf_split_str(buf, "\n ", 2, ARES_BUF_SPLIT_TRIM, 0, &argv, &argc);
ares__buf_destroy(buf);

....

ares_free_array(argv, argc, ares_free);
```

tacerus:
Thank you for the examples!
I'll try refactoring with your helpers.

tacerus:
I replaced the logic now, let me know what you think and if I should squash the previous attempts.

src/tools/adig.c
```diff
@@ -138,14 +145,77 @@ static void print_help(void)
...
+          ares__buf_split_str(rcbuf, (const unsigned char *)"\n ", 2, ARES_BUF_SPLIT_TRIM, 0, &rcargv, &rcargc);
+          ares__buf_destroy(rcbuf);
+
+          read_cmdline((int)rcargc, (const char * const *)rcargv, config, ARES_TRUE);
+
+        } else {
+          DEBUGF(fprintf(stderr, "read_cmdline() failed to load rcfile"));
+        }
+
+        ares_free_array(rcargv, rcargc, ares_free);
```
bradh352, line 214: You probably want to move this up within the if(), as otherwise rcargc and rcargv are uninitialized.  Either that, or initialize rcargc and rcargv.

src/tools/adig.c
```diff
@@ -138,14 +145,77 @@ static void print_help(void)
...
+
+      size_t          rcargc;
+      char          **rcargv;
+      ares__buf_t    *rcbuf;
+
+      rclen = (unsigned int) snprintf(rcfile, rcsize, "%s/adigrc", configdir);
+      if (rclen < rcsize) {
+        rcbuf = ares__buf_create();
+        if (ares__buf_load_file(rcfile, rcbuf) == ARES_SUCCESS) {
+          ares__buf_split_str(rcbuf, (const unsigned char *)"\n ", 2, ARES_BUF_SPLIT_TRIM, 0, &rcargv, &rcargc);
```
bradh352, line 205: Should probably check this result so you don't silently ignore errors.  This will fail on things like non-printable data.

src/tools/adig.c
```diff
@@ -138,14 +145,77 @@ static void print_help(void)
...
+  unsigned int         cdlen = 0;
+  unsigned int         cdsize = sizeof(configdir);
+  char                *homedir;
 
   ares_getopt_init(&state, argc, argv);
   state.opterr = 0;
 
+  if (is_rcfile) {
+    state.optind = 0;
+  } else {
```
bradh352, line 163: This whole code block should be pulled out into a helper function that is called from within the main() function.  So you'll process the RCfile first then the command line args.

src/tools/adig.c
tacerus: Oops. I moved it now, which resolves the fault, but Valgrind seems to still report loss on the cmake build if no ~/.config/adigrc is present.

src/tools/adig.c
tacerus: Good point, I have it pass the error message from ares__buf_split_str() now.

src/tools/adig.c
tacerus: Done!

src/tools/adig.c
```diff
@@ -43,9 +43,16 @@
 #endif
 
 #include "ares.h"
+#include "ares__array.h"
+#include "ares__buf.h"
 #include "ares_dns.h"
+#include "ares_private.h"
```
bradh352, line 49: don't include ares_private.h

src/tools/adig.c
```diff
@@ -43,9 +43,16 @@
...
 #include "ares_str.h"
 
+#if defined(_WIN32)
+#  define strtok_r strtok_s
```
bradh352, line 53: this shouldn't be needed anymore

src/tools/adig.c
```diff
@@ -251,6 +262,80 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
   return ARES_TRUE;
 }
 
+static ares_bool_t read_rcfile(adig_config_t *config)
```
bradh352, line 265: in read_cmdline() should probably skip config->args_processed assignment and sanity check if is_rcfile

src/tools/adig.c
```diff
@@ -251,6 +262,80 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
...
+  }
+  else {
+    cdlen = (unsigned int) snprintf(configdir, cdsize, "%s", configdir_xdg);
+  }
+
+#endif
+
+  DEBUGF(fprintf(stderr, "read_cmdline() configdir: %s\n", configdir));
+
+  if (cdlen != 0 && cdlen < cdsize) {
```
bradh352, line 299: invert this if statement and just return out of the function early would probably make this more readable so you don't have an indented code block here.

src/tools/adig.c
tacerus: How do I get "ares_free" otherwise without it?

src/tools/adig.c
bradh352: include ares_mem.h

src/tools/adig.c
bradh352: include ares_mem.h

src/tools/adig.c
tacerus: Added.

src/tools/adig.c
tacerus: Fair point, I adjusted it and reduced some indentations.

tacerus:
I still notice this when running the `cmake` through Valgrind:

```
==1480325== HEAP SUMMARY:
==1480325==     in use at exit: 48 bytes in 1 blocks
==1480325==   total heap usage: 278 allocs, 277 frees, 157,888 bytes allocated
==1480325==
==1480325== 48 bytes in 1 blocks are definitely lost in loss record 1 of 1
==1480325==    at 0x4846794: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1480325==    by 0x4870D58: ares_malloc_zero (in /home/georg/Work/git/c-ares/build/lib64/libcares.so.2.18.1)
==1480325==    by 0x488989D: ares__buf_create (in /home/georg/Work/git/c-ares/build/lib64/libcares.so.2.18.1)
==1480325==    by 0x4025DF: main (in /home/georg/Work/git/c-ares/build/bin/adig)
==1480325==
==1480325== LEAK SUMMARY:
==1480325==    definitely lost: 48 bytes in 1 blocks
==1480325==    indirectly lost: 0 bytes in 0 blocks
==1480325==      possibly lost: 0 bytes in 0 blocks
==1480325==    still reachable: 0 bytes in 0 blocks
==1480325==         suppressed: 0 bytes in 0 blocks
```

tacerus:
Ah I think I found the problem, `ares__buf_destroy()` was inside the conditional.

src/tools/adig.c
```diff
@@ -243,14 +256,108 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
     }
   }
 
-  config->args_processed = state.optind;
+  if (!is_rcfile) {
+    config->args_processed = state.optind;
+  }
```
bradh352, line 261: move the closing bracket below the sanity check that uses the setting

src/tools/adig.c
```diff
@@ -243,14 +256,108 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
...
     return ARES_FALSE;
   }
   return ARES_TRUE;
 }
 
+static ares_bool_t read_rcfile(adig_config_t *config)
+{
+  char         configdir[128];
+  unsigned int cdlen  = 0;
+  unsigned int cdsize = sizeof(configdir);
```
bradh352, line 273: lets just use sizeof(configdir) instead of defining a variable for this, its more consistent with other c-ares code

src/tools/adig.c
```diff
@@ -243,14 +256,108 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
...
+
+#if !defined(WIN32)
+#  if !defined(__APPLE__)
+  char *configdir_xdg;
+#  endif
+  char *homedir;
+#endif
+
+  char          rcfile[128];
```
bradh352, line 282: might be better to use PATH_MAX instead of 128 here, likely need to include limits.h to get that

src/tools/adig.c
```diff
@@ -243,14 +256,108 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
...
+  unsigned int  rclen;
+  unsigned int  rcsize = sizeof(rcfile);
```
bradh352, line 284: lets just use sizeof(rcfile) where needed rather than rcsize

src/tools/adig.c
tacerus: Sure, changed.

src/tools/adig.c
```diff
@@ -243,14 +256,108 @@ static ares_bool_t read_cmdline(int argc, const char * const *argv,
...
   if (config->args_processed >= argc) {
     snprintf(config->error, sizeof(config->error), "missing query name");
     return ARES_FALSE;
   }
   return ARES_TRUE;
 }
 
+static ares_bool_t read_rcfile(adig_config_t *config)
+{
+  char         configdir[128];
```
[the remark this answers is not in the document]
tacerus: Thanks, replaced - it worked without including limits.h (might already be through some other header).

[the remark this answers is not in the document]
tacerus: ok, seems it is needed for some platforms. ;)

src/tools/adig.c
tacerus: For rcfile I replaced it as well now, but that yields format-truncation warning.

src/tools/adig.c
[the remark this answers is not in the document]
tacerus: On WATT32 (Windows?) it does not seem to do:

```
error C2065: 'PATH_MAX': undeclared identifier
```

Shall I try to define it manually for Win32?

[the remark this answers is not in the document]
bradh352: yeah, or just an #ifndef PATH_MAX ... looks like Windows decided to use MAX_PATH, technically, but i'd just define it to 1024 or something to be safe.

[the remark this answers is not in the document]
tacerus: Sounds good, added, and now it builds on Windows again as well.

bradh352:
looks good to me.  you ready for this to be merged, or have anything else ?

tacerus:
It's good to ship by me as well. Just note I did not have any means to test behavior on platforms other than Linux.
I might revisit it for some feature enhancement in a future patch.

PR #867, opened 2024-08-25
new library functions

bradh352:
* ares__buf_split_str: Ability to split a buf object into a C array of C strings
* ares__array_insertdata_{at,last,first}: copy data provided into array member as a helper
* ares_free_array: Free a normal C array with custom callback to free each element

### DIFF
--- a/src/lib/str/ares__buf.c
+++ b/src/lib/str/ares__buf.c
@@ -906,6 +906,9 @@ ares_status_t ares__buf_split_str(ares__buf_t *buf, const unsigned char *delims,
     return ARES_EFORMERR;
   }
 
+  *strs  = NULL;
+  *nstrs = 0;
+
   status = ares__buf_split(buf, delims, delims_len, flags, max_sections, &list);
   if (status != ARES_SUCCESS) {
     goto done;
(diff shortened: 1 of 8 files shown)


src/lib/str/ares__buf.c
```diff
@@ -884,6 +892,61 @@ ares_status_t ares__buf_split(ares__buf_t *buf, const unsigned char *delims,
...
+      goto done;
+    }
+  }
+
+done:
+  ares__llist_destroy(list);
+  if (status == ARES_SUCCESS) {
+    *strs = ares__array_finish(arr, nstrs);
+  } else {
+    ares__array_destroy(arr);
```
tacerus, line 945: I'm not sure I'm missing something in my calling code, but trying to use `ares__buf_split_str()` like in the example at https://github.com/c-ares/c-ares/pull/856#issuecomment-2308956347 I get a `free(): double free detected in tcache 2` at this stage.

bradh352: I need to add some tests still before I merge this.  Infact, this hasn't been tested at all yet, so its not unlikely there's some issues hiding out in this PR.

tacerus: Ah alright, I'll stay tuned then. :)

bradh352: ok, fixed now :)

PR #870, opened 2024-08-27
MSVC: Force different PDB names for different targets

bradh352:
During compiling using MSVC and Ninja, its not uncommon to see errors such as:
```
C:\projects\c-ares\test\ares_queryloop.c: fatal error C1041: cannot open program database 'C:\projects\c-ares\build\bin\vc140.pdb'; if multiple CL.EXE write to the same .PDB file, please use /FS
```

We see this as build failures in our AppVeyor runs.

The suggestion of using `/FS` is misleading as it already exists in the compiler flags (and previous attempts to append them again were wrong).  The real issue I believe is that different output targets are being compiled in parallel, but trying to use the same pdb file, but clearly shouldn't be since they're different targets.

This CMake issue may be related:
https://gitlab.kitware.com/cmake/cmake/-/issues/20188

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -708,15 +708,6 @@ IF (CARES_COVERAGE)
 	)
 ENDIF()
 
-IF (MSVC)
-	# AppVeyor reports issues like "fatal error C1041: cannot open program database" randomly.
-	# The /FS flag is supposed to correct this on highly parallel builds like performed when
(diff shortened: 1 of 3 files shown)


bradh352:
issued an appveyor rebuild a few times.  Not definitive of course because of how rarely it occurs, but its at least no worse :/

PR #875, opened 2024-09-05
New function `ares_process_fds()`

bradh352:
Add a new public function of `ares_process_fds()` to work around the shortcomings of `ares_process_fd()` and `ares_process()`. 

This new function allows integrators to specify more than one file descriptor at a time in an array, with an event mask per file descriptor.  It also adds a flags parameter which can be used to skip extra expensive processing if they may be calling the function multiple times before waiting on more events.

This new function is used by the event thread, and all the other functions internally call this new function simplifying the code structure.

### DIFF
--- a/docs/Makefile.inc
+++ b/docs/Makefile.inc
@@ -115,6 +115,7 @@ MANPAGES = ares_cancel.3		\
   ares_parse_uri_reply.3		\
   ares_process.3			\
   ares_process_fd.3			\
+  ares_process_fd2.3			\
   ares_process_pending_write.3		\
   ares_query.3				\
   ares_query_dnsrec.3			\
--- a/docs/Makefile.inc
+++ b/docs/Makefile.inc
@@ -115,7 +115,7 @@ MANPAGES = ares_cancel.3		\
   ares_parse_uri_reply.3		\
   ares_process.3			\
   ares_process_fd.3			\
-  ares_process_fd2.3			\
+  ares_process_fds.3			\
   ares_process_pending_write.3		\
   ares_query.3				\
   ares_query_dnsrec.3			\
--- a/docs/ares_process.3
+++ b/docs/ares_process.3
@@ -23,10 +23,16 @@ typedef struct {
   unsigned int  events; /*!< Mask of ares_fd_event_t */
 } ares_fd_events_t;
 
+typedef enum {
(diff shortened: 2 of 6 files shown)


jmaggard10:
@bradh352 , this commit caused a regression, where now if I use the event thread and do two consecutive lookups with a dead server, the second lookup hangs forever. Please see the attached simple test program for a quick reproducer.
[test-pr-875.zip](https://github.com/user-attachments/files/21508877/test-pr-875.zip)

jmaggard10:
I was able to hack around it like this, but it's probably not the right way to approach it.
```
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -26,6 +26,7 @@
  */
 
 #include "ares_private.h"
+#include "event/ares_event.h"
 
 #ifdef HAVE_STRINGS_H
 #  include <strings.h>
@@ -1333,6 +1334,14 @@ ares_status_t ares_send_query(ares_server_t *requested_server,
     ares_probe_failed_server(channel, server, query);
   }
 
+  /* If using the event thread backend, wake the event thread */
+  if (channel->optmask & ARES_OPT_EVENT_THREAD) {
+    ares_event_thread_t *e = channel->sock_state_cb_data;
+    if (e) {
+      ares_event_update(NULL, e, 0, NULL, 0, NULL, NULL, NULL);
+    }
+  }
+
   return ARES_SUCCESS;
 }
 
```

bhouse-nexthop:
> @bradh352 , this commit caused a regression, where now if I use the event thread and do two consecutive lookups with a dead server, the second lookup hangs forever. Please see the attached simple test program for a quick reproducer. [test-pr-875.zip](https://github.com/user-attachments/files/21508877/test-pr-875.zip)

Do you think this is the same as https://github.com/c-ares/c-ares/issues/1000 ?  I really meant to have time to look into that.

jmaggard10:
> > @bradh352 , this commit caused a regression, where now if I use the event thread and do two consecutive lookups with a dead server, the second lookup hangs forever. Please see the attached simple test program for a quick reproducer. [test-pr-875.zip](https://github.com/user-attachments/files/21508877/test-pr-875.zip)
> 
> Do you think this is the same as #1000 ? I really meant to have time to look into that.

Oh, yes that's exactly the same issue.

PR #877, opened 2024-09-08
Probe for failed servers instead of redirecting query

bradh352:
The previous implementation would redirect a query to a failed server based on a timeout and random chance per query.  This could lead to issues of having to deal with server timeout scenarios when the server isn't back online yet causing latency issues.  Instead, we should continue to use the known good servers for the query itself, but spawn a second query with the same question to a different downed server.  That query will be able to be processed in the background and potentially bring the server back online.

Also, when using the `rotate` option, servers were previously chosen at random from the complete list.  This PR changes that to choose only from the servers that share the same highest priority.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -60,9 +60,9 @@ Administrators may customize these settings via `ARES_OPT_SERVER_FAILOVER`.
 In the future we may use independent queries to probe downed servers to not
 impact latency of any queries when a server is known to be down.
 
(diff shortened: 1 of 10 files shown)


bradh352:
@oliverwelsh I reworked your server retry logic you contributed to duplicate the query and use it as a probe rather than scheduling the actual query to the failed server to improve responsiveness.  See any issues with this strategy?

PR #887, opened 2024-09-24
ares_socket: set IP_BIND_ADDRESS_NO_PORT on ares_set_local_ip* tcp sockets

crrodriguez:
If you bind to a local address, you now only have approx 32k possible source ports to initiate connections.
In modern days that can quickly run out.
setting IP_BIND_ADDRESS_NO_PORT let's the kernel choose a port at connect time, increasing the limit of combinations to around a million.


bradh352 (changes requested):
Can you use the same pattern as SO_NOSIGPIPE?  `int opt = 1;` and pass that in instead of the form you're using?

crrodriguez:
ook. compound literals are valid c99 thou..

PR #890, opened 2024-09-25
adig: rework command line arguments to mimic dig from bind

### LINKED ISSUE #885: adig: Is there a way to inactivate OPT record
Currently, it uses EDNS by default:
https://github.com/c-ares/c-ares/blame/8b2f5e01ebc39f50438911f610dfd4edc2bc7b26/src/lib/ares_init.c#L132-L135

It results in unable to use `adig` for testing purpose without OPT record.
Flag `edns` for `-f` staled.

bradh352:
The c-ares dig utility outputs a similar format to the bind dig utility, we should also make it take arguments that mimic theirs for compatibility.

This PR starts down that path supporting some of the major options.

Fixes #885


bradh352:
still need to redo the manpage

PR #893, opened 2024-10-07
ares_socket: add virtual function for getsockname()

tchaikov:
in 4bedfd0d, we introduced DNS cookie support, which uses `getsockname()` to retrieve the client IP address. but this broke the use case where the parent project uses an abstraction layer on top of userspace TCP/UDP stack, and the fd is not necessarily a real file descriptor in system, instead, it could be an index in a map in application, and the map keeps track of the state of sockets implemented with user space IP stack. in that case, `ECONNREFUSED` is always used as the fd passed to `getsockname()` is not necessily a valid socket fd, not to mention the real fd under the hood.

in this change, we

* introduce another virtual function to `ares_socket_functions`. this allows a c-ares application to provide its own implementatino in the place of `getsockname()`.
* use the new virtual function if it is defined.

### DIFF
--- a/docs/ares_set_socket_functions.3
+++ b/docs/ares_set_socket_functions.3
@@ -14,6 +14,7 @@ struct ares_socket_functions {
     ares_ssize_t (*\fIarecvfrom\fP)(ares_socket_t, void *, size_t, int,
                               struct sockaddr *, ares_socklen_t *, void *);
     ares_ssize_t (*\fIasendv\fP)(ares_socket_t, const struct iovec *, int, void *);
+    int (*\fIagetsockname\fP)(ares_socket_t, struct sockaddr *, ares_socklen_t *);
 };
 
 void ares_set_socket_functions(ares_channel_t *\fIchannel\fP,
--- a/include/ares.h
+++ b/include/ares.h
@@ -565,6 +565,7 @@ struct ares_socket_functions {
   ares_ssize_t (*arecvfrom)(ares_socket_t, void *, size_t, int,
                             struct sockaddr *, ares_socklen_t *, void *);
   ares_ssize_t (*asendv)(ares_socket_t, const struct iovec *, int, void *);
+  int (*agetsockname)(ares_socket_t, struct sockaddr *, ares_socklen_t *);
 };
 
 CARES_EXTERN void
--- a/src/lib/ares_conn.c
+++ b/src/lib/ares_conn.c
@@ -129,8 +129,9 @@ static ares_status_t ares_conn_set_sockaddr(const ares_conn_t *conn,
 
 static ares_status_t ares_conn_set_self_ip(ares_conn_t *conn, ares_bool_t early)
 {
+  ares_channel_t         *channel = conn->server->channel;
   struct sockaddr_storage sa_storage;
-  int                     rv;
+  ares_conn_err_t         err = ARES_CONN_ERR_SUCCESS;
   ares_socklen_t          len = sizeof(sa_storage);
 
   /* We call this twice on TFO, if we already have the IP we can go ahead and
@@ -141,8 +142,8 @@ static ares_status_t ares_conn_set_self_ip(ares_conn_t *conn, ares_bool_t early)
 
   memset(&sa_storage, 0, sizeof(sa_storage));
 
-  rv = getsockname(conn->fd, (struct sockaddr *)(void *)&sa_storage, &len);
-  if (rv != 0) {
+  err = ares_socket_get_sockname(channel, conn->fd, (struct sockaddr *)(void *)&sa_storage, &len);
+  if (err != ARES_CONN_ERR_SUCCESS) {
     /* During TCP FastOpen, we can't get the IP this early since connect()
      * may not be called.  That's ok, we'll try again later */
     if (early && conn->flags & ARES_CONN_FLAG_TCP &&
--- a/src/lib/ares_socket.c
+++ b/src/lib/ares_socket.c
@@ -730,6 +730,25 @@ void ares_socket_close(ares_channel_t *channel, ares_socket_t s)
   }
 }
 
+ares_conn_err_t ares_socket_get_sockname(ares_channel_t *channel, ares_socket_t s,
+                                         struct sockaddr *sa,
+                                         ares_socklen_t *addrlen)
+{
+  int             rv;
+  ares_conn_err_t err = ARES_CONN_ERR_SUCCESS;
+
+  if (channel->sock_funcs && channel->sock_funcs->agetsockname) {
+    rv = channel->sock_funcs->agetsockname(s, sa, addrlen);
+  } else {
+    rv = getsockname(s, sa, addrlen);
+  }
+
+  if (rv < 0) {
+    err = ares_socket_deref_error(SOCKERRNO);
+  }
+  return err;
+}
+
 void ares_set_socket_callback(ares_channel_t           *channel,
                               ares_sock_create_callback cb, void *data)
 {
--- a/src/lib/ares_socket.h
+++ b/src/lib/ares_socket.h
@@ -77,4 +77,7 @@ ares_conn_err_t ares_socket_write_tfo(ares_channel_t *channel, ares_socket_t fd,
                                       size_t                *written,
                                       const struct sockaddr *sa,
                                       ares_socklen_t         salen);
+ares_conn_err_t ares_socket_get_sockname(ares_channel_t *channel, ares_socket_t fd,
+                                         struct sockaddr *sa,
+                                         ares_socklen_t *addrlen);
 #endif
--- a/test/ares-test-internal.cc
+++ b/test/ares-test-internal.cc
@@ -1790,6 +1790,7 @@ const struct ares_socket_functions VirtualizeIO::default_functions = {
   NULL,
   NULL,
   NULL,
+  NULL,
   NULL
 };
 


bradh352:
hmm, I think the problem with this PR is it is definitely an ABI break.  I was wondering who might be using the socket callbacks these days.  I just took a look at Seastar and see it implements its own userspace TCP/IP stack, so its definitely a valid usecase, especially since it would be a circular dependency (if it wasn't circular, I'd say it would make sense to add Seastar support to c-ares as a build-time option).  That said if you can figure out a way to make it a build-time dependency, I'd accept patches for that ... that would be the better route as you'd get to use our full test suite for testing as well as our tools like 'adig' and 'ahost' would work under seastar.  Likewise, you'd get to use the new event thread code paths which can greatly simplify integrations.

Regarding getsockname() itself, it is also called in ares_sortaddrinfo.c and has for years as part of the address sorting logic as required by getaddrinfo().

I think the appropriate thing to do is to create a `ares_set_socket_functions_ex()` and mark `ares_set_socket_functions()` as deprecated. The new functions structure should have, as its first member, an `unsigned int version;` and mandate that all known callbacks for that version are populated (current version allows you set some to NULL).  Internally to c-ares, we'd need to have a policy that any extension to the structure must be optional or conditional in some way, meaning it shouldn't fail unless a specific feature is requested that requires the new callback (implying that an integrator is trying to use a new feature but didn't implement the callbacks right).  We just don't want existing code to fail just because an external library is updated (like it appears it is now).

Also, we should go ahead and not call getsockname() if socket functions were overwritten in the old way.  We could just ignore ip address changes on the client side and still use cookies the same.  I don't think there'd be much impact other than slight RFC divergence.

Does this seem reasonable?  If so I can go ahead and implement that so it doesn't fall on you.

tchaikov:
Thank you Brad for your prompt reply and constructive suggestions and help!

> hmm, I think the problem with this PR is it is definitely an ABI break. 

yeah, the ABI breakage is indeed a show stopper. 

> I was wondering who might be using the socket callbacks these days. I just took a look at Seastar and see it implements its own userspace TCP/IP stack, so its definitely a valid usecase, especially since it would be a circular dependency (if it wasn't circular, I'd say it would make sense to add Seastar support to c-ares as a build-time option). 

i don't think it would be a circular dependency. because seastar's dns subsystem -> c-ares -> virtual functions -> seastar's networking stack -> DPDK. 

> That said if you can figure out a way to make it a build-time dependency, I'd accept patches for that ... that would be the better route as you'd get to use our full test suite for testing as well as our tools like 'adig' and 'ahost' would work under seastar. Likewise, you'd get to use the new event thread code paths which can greatly simplify integrations.
> 

yeah, we could have `adig` and `ahost` implemented using seastar as well. in other words, we can drive c-ares using seastar, and perform all async operations using C++20 coroutines. as seastar is a C++ library. but i am afraid this would be a very intrusive change.

> Regarding getsockname() itself, it is also called in ares_sortaddrinfo.c and has for years as part of the address sorting logic as required by getaddrinfo().

yeah, probably because we don't check the return value of `ares__sortaddrinfo()` and fail the query. so we survived this until 1.33.0 .

> 
> I think the appropriate thing to do is to create a `ares_set_socket_functions_ex()` and mark `ares_set_socket_functions()` as deprecated. The new functions structure should have, as its first member, an `unsigned int version;` and mandate that all known callbacks for that version are populated (current version allows you set some to NULL). Internally to c-ares, we'd need to have a policy that any extension to the structure must be optional or conditional in some way, meaning it shouldn't fail unless a specific feature is requested that requires the new callback (implying that an integrator is trying to use a new feature but didn't implement the callbacks right). We just don't want existing code to fail just because an external library is updated (like it appears it is now).
> 
> Also, we should go ahead and not call getsockname() if socket functions were overwritten in the old way. We could just ignore ip address changes on the client side and still use cookies the same. I don't think there'd be much impact other than slight RFC divergence.
> 
> Does this seem reasonable? If so I can go ahead and implement that so it doesn't fall on you.

it does make sense. just out of curiosity, i am wondering how we are going to enforce the policy on a certain mandatory / optional socket function. is it something like:
```c
void
ares_set_socket_functions_ex(ares_channel_t                     *channel,
                                       const struct ares_socket_functions_ex *funcs,
                                       void                               *user_data)
{
  if (channel->optmask & ARES_OPT_DNS_COOKIE) {
    if (!funcs->agetsockname) {
      // choke
    }
  }
}
```

bradh352:
> it does make sense. just out of curiosity, i am wondering how we are going to enforce the policy on a certain mandatory / optional socket function. is it something like:
> 
> ```c
> void
> ares_set_socket_functions_ex(ares_channel_t                        *channel,
>                              const struct ares_socket_functions_ex *funcs,
>                              void                                  *user_data)
> {
>   if (channel->optmask & ARES_OPT_DNS_COOKIE) {
>     if (!funcs->agetsockname) {
>       // choke
>     }
>   }
> }
> ```

Yep, pretty much...

I'm starting to work on a branch now ... https://github.com/bradh352/c-ares/tree/sock_functions  I'd say I'm about 50% done.  I just need to properly figure out how to handle things like TCP FastOpen.  Switching to always using callbacks and just hook in the native OS callbacks by default so we have a more straight forward code path for supporting pluggable networking libraries.

tchaikov:
>  I just need to properly figure out how to handle things like TCP FastOpen.

was reading your implementation, yes, it's convoluted to support TFO, what about

1. having a wrapper around the user-provided `data`. it's like a closure, so we can store both the `aconnect` and `data` provided by user.
2. having wrapper functions around the user-provided sock functions
3. add `is_tfo` parameter to `aconnect()`, as we are introducing a new API, so i think we have the luxury of having another function signature of `aconnect()`


 it's not perfect, and it hurts the maintainability. but it addresses the backward compatibility to `ares_set_socket_functions()`, like
```c
struct ares_legacy_funcs_data {
  int (*aconnect)(ares_socket_t sock, const struct sockaddr *address, ares_socklen_t address_len,
                  void *user_data);
  void* user_data;
};
static ares_legacy_aconnect(ares_socket_t sock, ares_bool_t is_tfo, const struct sockaddr *address, ares_socklen_t address_len,
                  void *user_data) {
  ares_socket_funcs_data* funcs_data = user_data;
  if (is_tfo) {
    // connectx()
  } else {
    return funcs_data->aconnect(sock, address, len, funcs_data->user_data);
}
// and adaptors to other sock functions.
void ares_set_socket_functions(ares_channel_t                     *channel,
                               const struct ares_socket_functions *funcs,
                               void                               *data)
{
  // free it in ares_destroy()?
  channel->ares_legacy_funcs_data = ares_malloc(sizeof(*ares_func_data));
  channel->ares_legacy_funcs_data->aconnect = funcs->aconnect;
  channel->ares_legacy_funcs_data->user_data = data;
  channel->sock_funcs.aconnect = ares_legacy_aconnect;
  // ...
  channel->sock_func_cb_data = channel->ares_legacy_funcs_data;
}
```

bradh352:
i think you were evaluating a commit from a couple hours ago.  My latest commit is closer and does pass flags onto the connect callback now for things like TFO.  I need to address some test case failures, and also need to make a wrapper for bind() for binding a source address.

You'll see quite a few new callbacks need to be registered already to try to allow registered callbacks to implement all functionality including TFO.

But I'm also making the legacy stuff at least work so existing programs should be fixed once I'm done even if they won't fully support newer features.

bradh352:
I should also mention I saw this comment in seastar: https://github.com/scylladb/seastar/blob/bddc7b27db27c0e21d0f08c032180e7c06ac9970/src/net/dns.cc#L1017

That is no longer true.

Also, ares_fds() really shouldn't ever be used, there's so many downsides.  The socket state callbacks are what really need to be used these days.  We have an example how to use those here: https://c-ares.org/docs.html#examples

bradh352:
@tchaikov There are a few other functions I'm wondering if are relevant when the IP stack is switched out or if maybe they should be the same as the underlying OS.  If you can comment on this, I'd appreciate it:
 * `gethostname()`
 * `if_nametoindex()`
 * `if_indextoname()`

tchaikov:
@bradh352 hi Brad,

> @tchaikov There are a few other functions I'm wondering if are relevant when the IP stack is switched out or if maybe they should be the same as the underlying OS. If you can comment on this, I'd appreciate it:
> 
>     * `gethostname()`

IIUC, hostname is backed by utsname which tracks the current system information in the uts namespace of the current process. so, i don't think this syscall is relevant when switching to another IP stack. it should be the same as the underlying OS.

> 
>     * `if_nametoindex()`

in order to simplify the discussion, let's assume that we are using DPDK as the hardware abstraction layer. but DPDK does not use the names like "eth0". instead, it enumerates the PCI devices attached to PCI bus based on their vendor ID and device ID.  since seastar's userspace IP stack uses DPDK to access the underlying devices,  i don't think this applies to DPDK. but if we are trying to figure out which physical interface to use when sending packets in an environment supporting link-local server. i think it is indeed relevant when switching to another networking stack, even if the networking stack does not support the feature implemented by d974c556bbd078e425419096056b9870068b43b3 .

> 
>     * `if_indextoname()`

the same applies, assuming this syscall is also used in the context of `ares_sconfig_linklocal()` implementation.

bradh352:
> @bradh352 hi Brad,
> 
> > @tchaikov There are a few other functions I'm wondering if are relevant when the IP stack is switched out or if maybe they should be the same as the underlying OS. If you can comment on this, I'd appreciate it:
> > ```
> > * `gethostname()`
> > ```
> 
> IIUC, hostname is backed by utsname which tracks the current system information in the uts namespace of the current process. so, i don't think this syscall is relevant when switching to another IP stack. it should be the same as the underlying OS.
> 
> > ```
> > * `if_nametoindex()`
> > ```
> 
> in order to simplify the discussion, let's assume that we are using DPDK as the hardware abstraction layer. but DPDK does not use the names like "eth0". instead, it enumerates the PCI devices attached to PCI bus based on their vendor ID and device ID. since seastar's userspace IP stack uses DPDK to access the underlying devices, i don't think this applies to DPDK. but if we are trying to figure out which physical interface to use when sending packets in an environment supporting link-local server. i think it is indeed relevant when switching to another networking stack, even if the networking stack does not support the feature implemented by [d974c55](https://github.com/c-ares/c-ares/commit/d974c556bbd078e425419096056b9870068b43b3) .
> 
> > ```
> > * `if_indextoname()`
> > ```
> 
> the same applies, assuming this syscall is also used in the context of `ares_sconfig_linklocal()` implementation.

Ok, so yeah, I'll need to add callbacks for if_indextoname/if_nametoindex for proper ipv6 link local scope support to my PR.

bradh352:
@tchaikov I guess that brings up another question, how are DNS servers configured in this environment?  Are we still reading the system configuration for that?  If so, link local scope might be an issue if the names/indexes are different for the system using native calls vs something using DPDK, no ?

bradh352:
> > If so, link local scope might be an issue if the names/indexes are different for the system using native calls vs something using DPDK, no ?
> 
> but, if a network device is not managed / indexed by the kernel's networking stack, how can it be "different"? different from what?

I'm not familiar with DPDK, ok, so you're saying that a NIC either uses the OS or DPDK, it can't be both ways.

So that brings up 2 questions.
1. If we have a server in resolv.conf like `nameserver fe80:abcd:1234::1%eth0`  which is a DNS server residing across a link-local interface configured by the native OS networking, what is supposed to happen with Seastar/DPDK in this sort of instance?  I'd assume it should be ignored, and I guess would be if `if_nametoindex()` failed.
2. What about Seastar/DPDK, wouldn't there be instances where you might receive an RDNSS advertisement containing a link-local DNS server?  Do you populate the system DNS configuration with that?  If not, how might c-ares learn about such a server?

tchaikov:
> > > If so, link local scope might be an issue if the names/indexes are different for the system using native calls vs something using DPDK, no ?
> > 
> > 
> > but, if a network device is not managed / indexed by the kernel's networking stack, how can it be "different"? different from what?
> 
> I'm not familiar with DPDK, ok, so you're saying that a NIC either uses the OS or DPDK, it can't be both ways.
> 
> So that brings up 2 questions.
> 
> 1. If we have a server in resolv.conf like `nameserver fe80:abcd:1234::1%eth0`  which is a DNS server residing across a link-local interface configured by the native OS networking, what is supposed to happen with Seastar/DPDK in this sort of instance?  I'd assume it should be ignored, and I guess would be if `if_nametoindex()` failed.

this is an interesting setting. Seastar either uses the userspace stack or the kernel stack, but not both. if a name server uses link-local interface, we would have to ignore it. as we don't have access to "eth0" with DPDK. so yes, i agree with you. unless we emulate one and implement the link-local-as-DNS-server machinery in Seastar. but AFAICT, we don't.

> 2. What about Seastar/DPDK, wouldn't there be instances where you might receive an RDNSS advertisement containing a link-local DNS server?  Do you populate the system DNS configuration with that?  If not, how might c-ares learn about such a server?

if you are talking about https://datatracker.ietf.org/doc/html/rfc8106. i am afraid that we don't support it at this moment. actually, we don't support IPv6 in our DNS subsystem yet.

bradh352:
correct, that's 

> > > > If so, link local scope might be an issue if the names/indexes are different for the system using native calls vs something using DPDK, no ?
> > > 
> > > 
> > > but, if a network device is not managed / indexed by the kernel's networking stack, how can it be "different"? different from what?
> > 
> > 
> > I'm not familiar with DPDK, ok, so you're saying that a NIC either uses the OS or DPDK, it can't be both ways.
> > So that brings up 2 questions.
> > 
> > 1. If we have a server in resolv.conf like `nameserver fe80:abcd:1234::1%eth0`  which is a DNS server residing across a link-local interface configured by the native OS networking, what is supposed to happen with Seastar/DPDK in this sort of instance?  I'd assume it should be ignored, and I guess would be if `if_nametoindex()` failed.
> 
> this is an interesting setting. Seastar either uses the userspace stack or the kernel stack, but not both. if a name server uses link-local interface, we would have to ignore it. as we don't have access to "eth0" with DPDK. so yes, i agree with you. unless we emulate one and implement the link-local-as-DNS-server machinery in Seastar. but AFAICT, we don't.
> 
> > 2. What about Seastar/DPDK, wouldn't there be instances where you might receive an RDNSS advertisement containing a link-local DNS server?  Do you populate the system DNS configuration with that?  If not, how might c-ares learn about such a server?
> 
> if you are talking about https://datatracker.ietf.org/doc/html/rfc8106. i am afraid that we don't support it in our TCP/IP stack at this moment. actually, we don't support IPv6 in our DNS subsystem yet. but, as you know, this is not inherently a limit of DPDK. it's just this spec is not implemented.

correct, that is what I'm talking about.  Ok, so we'll "ignore" this whole link-local DNS issue for now since its only relevant to ipv6.

tchaikov:
thank you again, Brad for your help!

closing this PR. as it is superseded by #894

PR #894, opened 2024-10-08
Improve socket function replacement subsystem

bradh352:
Some external users swap out the entire IP stack, such as [Seastar](https://github.com/scylladb/seastar) that uses DPDK.  So we need to allow them to override all network stack functions used by c-ares.  Since we don't know what network functions may be needed in the future, we need to make a versioned structure that can be expanded.  We'll need to disable any features in the future if the versioned interface is less than is required to support any new features.

We recently had a regression when we added DNS Cookie support as we relied on `getsockname()` to get the local source ip address, but we were calling the native version since there was no registered callback available.  We will mark this as optional and just skip this step in DNS Cookie invalidation if its not available for the legacy api users.

Supersedes #893 


bradh352:
still need to update manpages, and work with seastar to ensure there is nothing else that needs to be extended

bradh352:
@tchaikov Other than updating the manpages, I think the code itself is done.  I added if_nametoindex() and if_indextoname() as external network functions that can be specified in the new api.  

That said, even without changing anything in your existing integration, it should fix the regression from 1.33.0.

PR #895, opened 2024-10-09
ares_conn: use sock_funcs.agetsockname() instead of getsockname()

tchaikov:
this change addresses an issue with DNS cookie invalidation for users of the legacy API and custom network stacks (e.g., Seastar's DNS).

in 02745437, we added an optional `agetsockname` to `ares_socket_functions_ex` struct, and intended to skip calling this function in DNS cookie invalidation if this function is not available for legacy API users.

but we are still calling this system call directly, this fails Seastar's DNS which uses c-ares to resolve names, and it uses the socket function table to override the network stack to use its own userspace TCP/IP stack, in that case, the fd passed to c-ares is not a valid file descriptor, hence the call fails.

in this change, we

- replace direct getsockname() call with sock_funcs.agetsockname()
- skip the call if agetsockname() is not configured

Refs 02745437

### DIFF
--- a/src/lib/ares_conn.c
+++ b/src/lib/ares_conn.c
@@ -129,6 +129,7 @@ static ares_status_t ares_conn_set_sockaddr(const ares_conn_t *conn,
 
 static ares_status_t ares_conn_set_self_ip(ares_conn_t *conn, ares_bool_t early)
 {
+  ares_channel_t         *channel = conn->server->channel;
   struct sockaddr_storage sa_storage;
   int                     rv;
   ares_socklen_t          len = sizeof(sa_storage);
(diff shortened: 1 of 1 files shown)


tchaikov:
@bradh352 hi Brad, could you please help review this change? i tested the main HEAD with Seastar. after applying this change, the tests exercising seastar's dns subsystem  passes.

tchaikov:
see also the discussion at https://github.com/c-ares/c-ares/pull/893#issuecomment-2396825207

bradh352:
ugh, I apparently committed the changes to ares_conn.c in a different branch I was working on.  Sorry about that.  I'll go ahead and accept this PR as mine did basically the same thing.

PR #910, opened 2024-11-08
docs: add deprecation note for dns function

createyourpersonalaccount:


createyourpersonalaccount:
Hm, I've noticed that I get an `ARES_EBADRESP` even though I haven't configured a DNS server for the channel nor a query for the record. Is this a bug? The situation is:

```c
ares_init_options(&channel, &options, ARES_OPT_EVENT_THREAD);
ares_dns_record_create(&dnsrec, 0, 0, ARES_OPCODE_QUERY, 0);
ares_send_dnsrec(channel, dnsrec, callback, NULL, &query_id);
ares_destroy(channel);
ares_library_cleanup();
exit(0);
```

I know that I'm supposed to wait for the query, I'm just wondering if `ARES_EBADRESP` is an appropriate status in this situation for the callback.

bradh352:
Its possible that this is a bad error code, it maybe should be `ARES_EBADQUERY` since your query isn't properly formed, you're missing the actual question, use `ares_dns_record_query_add()`.  My guess is its this line that's generating that error code:
https://github.com/c-ares/c-ares/blob/a70637c67ba2657adb54b077f4f8434f00ecbf83/src/lib/ares_send.c#L154

The duplicate query writes out the query as binary, then parses it into a new data structure.  The parse likely failed due to missing the question hence `ARES_EBADRESP` since a parser is typically used on responses.  I probably need to look at the duplicate function and catch this to rewrite the error code.

bradh352:
the status code should be better reflected in 91519e76 if I'm correct

docs/ares_create_query.3
```diff
@@ -19,6 +19,8 @@ int ares_create_query(const char *\fIname\fP,
                       int \fImax_udp_size\fP)
 .fi
 .SH DESCRIPTION
+Deprecated function. See \fIares_dns_record_create(3)\fP instead!
+
```
bradh352, line 23: Maybe:

> This function is deprecated as of c-ares 1.22.0.  Please see \fIares_dns_record_create(3)\fP.

docs/ares_mkquery.3
```diff
@@ -14,7 +14,7 @@ int ares_mkquery(const char *\fIname\fP, int \fIdnsclass\fP, int \fItype\fP,
                  int *\fIbuflen\fP)
 .fi
 .SH DESCRIPTION
-Deprecated function. See \fIares_create_query(3)\fP instead!
+Deprecated function. See \fIares_dns_record_create(3)\fP instead!
```
bradh352, line 17: Maybe:

> This function is deprecated as of c-ares 1.22.0. Please see \fIares_dns_record_create(3)\fP.

docs/ares_send.3
```diff
@@ -113,6 +113,9 @@ is being destroyed; the query will not be completed.
 .B ARES_ENOSERVER
 The query will not be completed because no DNS servers were configured on the
 channel.
+.TP 19
+.B ARES_EBADRESP
+There was a bad response by the DNS server.
```
bradh352, line 118: After my referenced commit, this should probably be changed to `ARES_EBADQUERY`

bradh352 (changes requested):

docs/ares_mkquery.3
createyourpersonalaccount: Were both functions deprecated at the same version? Because one had a deprecation note for the other, it seems strange that they were both deprecated at the same version.

bradh352: hmm, no. But the function to use got changed.  I'm actually not sure the version the original deprecation occurred.  I guess we can omit that since it wasn't previously documented.

createyourpersonalaccount: I will work on this PR in a few days, I'll try to dig up the appearance of `ares_create_query` from git.

createyourpersonalaccount: Yes, I can't find the original deprecation either. That seems to be lost in time now. Moving forward, it would be good to document deprecation...

`ares_create_query` was introduced in 90364defb038d5d6b97acf9df9ed152d2393a863 which was first included in release v1.10. Of course, that does not mean that `ares_mkquery` was deprecated at that release.

The commands I used were:
```sh
# Find the first commit introducing ares_create_query
git log -S ares_create_query --reverse | head -n1
# Find the first release including the commit
git tag --sort=creatordate --contains 90364defb038d5d6b97acf9df9ed152d2393a863 --merged main | head -n1
```

createyourpersonalaccount: What I'll do is find the release that first introduced the previous deprecation and say that it's been deprecated at least since then.

createyourpersonalaccount:
@bradh352 does this look okay?

bradh352:
ARES_ENOSERVER is definitely a valid return code to ares_send() and friends, why did it get removed?

createyourpersonalaccount:
> ARES_ENOSERVER is definitely a valid return code to ares_send() and friends, why did it get removed?

Oops! I meant to change the status I introduced, `ARES_EBADRESP`, not `ARES_ENOSERVER`.

PR #920, opened 2024-11-12
chore: license text is actually NTP

gjasny:
While working on a DEP-5 machine readable copyright file for the Debian packages, I noticed that the SPDX tag differs from the actual license text.

The license text in question is identical to the [NTP](https://spdx.org/licenses/NTP.html) license, not [MIT](https://spdx.org/licenses/MIT.html).


dimbleby:
Surely it is the text that is mis-copied from somewhere, rather than the identifier?

So far as I can see it does not make sense for the copyright on these files to be assigned to MIT, or for the license to note that "MIT makes no representations..." - this code has never had anything to do with MIT.  

Much more likely that the intention was to use the MIT license, and there was some mix-up in the contribution such that what got pasted in was instead a license from some MIT software, no?

I Am Not A Lawyer and have no idea whether or not it is problematic to correct such things...

dimbleby:
ah no I see, it is bits that escaped the sweep in #556

bradh352:
yeah, seems like it just needs an update to the license text, i guess i missed a few files when we switched to the modern mit license

gjasny:
> yeah, seems like it just needs an update to the license text, i guess i missed a few files when we switched to the modern mit license

I updated the text in the three files.

PR #931, opened 2024-12-07
Add new function ares_strnlen (#929)

createyourpersonalaccount:
Add a new function `ares_strnlen()` to fix the logic in default_asetsockopt(), see
<https://github.com/c-ares/c-ares/pull/929#issuecomment-2522990742>.


bradh352:
I think we should check to see if the system support strnlen() natively first, as it likely is optimized if it does, then fallback to this code if not.

createyourpersonalaccount:
@bradh352
1. Needs to be done for both autotools & CMake (fun!)
2. Need to guarantee same semantics across all implementations (even more fun!) (e.g. [this](https://www.gnu.org/software/gnulib/manual/html_node/strnlen.html))

I prefer the easiest path which is this, and I won't be doing what you're suggesting to spare myself the time... it can always be implemented later too.

bradh352:
Eh, the current cmake and autotools code is pretty straight forward to add something like that at least.  It used to be the autotools code was almost impossible to update and get right but I fixed that earlier this year.

The link you provided is kind of an odd thing for them to even mention.  I'd expect most implementations don't handle pointer overflows without crashing, and that's likely ok as the caller will probably crash somewhere else due to such poor use.

bradh352:
> > Unlike autotools, it doesn't auto-add stuff to the header.
> 
> It should be possible to write a CMake function `MY_CHECK_SYMBOL_EXISTS()` to auto-add those defines. Do you want that?

I honestly can't say I've seen that done, but I'm definitely no CMake expert.  It would make it easier to maintain in the long run, obviously.

src/lib/config-win32.h
```diff
@@ -190,6 +190,9 @@
 /* Define if you have the sendto function. */
 #define HAVE_SENDTO 1
 
+/* Define if you have the strnlen function. */
+#define HAVE_STRNLEN 1
+
```
bradh352, line 195: Well, I guess we need a
```
if !defined(__WATCOMC__)
```
around this since their library doesn't have strnlen like MSVC and MinGW do...

createyourpersonalaccount: > Glad we have all these CI systems building c-ares to catch this stuff.

Yes it is good because I'm not very careful with this PR. I can add the `strnlen_s()` variant for Watcom tomorrrow. The code already does a NULL check and returns 0 so that aligns with the `strnlen_s()` behavior.

createyourpersonalaccount: It appears that in order to access `strnlen_s()` in Watcom you need to define `__STDC_WANT_LIB_EXT1__` before including `<string.h>`. Do you want that?

bradh352: I could go either way on that.  Either just use the fallback strnlen code as performance doesn't matter for something like openwatcom, or just turn that flag on at the top of this header.  honestly, not sure when this was introduced to openwatcom, if someones using an older version than 2.0 (like 1.9) it might not even exist there.

createyourpersonalaccount: I'm seeing that `strnlen_s()` is in C17 in (normative) Annex K.3.7.4.4. Any implementation that defines `__STDC_LIB_EXT1__` will have `strnlen_s()`, and they will only exist if `__STDC_WANT_LIB_EXT1__` is set to nonzero by the programmer. I guess that instead of the Watcom macro we can just check for these.

createyourpersonalaccount:
@bradh352 
Here's info on `strnlen_s`, <https://stackoverflow.com/a/50921096>. Let me draw your attention to the fact that `strnlen_s` is, according to him, not supported by Windows as defined in the C standard annex.

I'm wondering if, in absence of `strnlen`, we should just fall back to `memchr(s, 0, maxlen)`. Or perhaps indeed scrap the entire PR and just use `memchr`.

bradh352:
I'm guessing EPEL must have just moved CentOS 7 support to archive ... guess I need to look at fixing the repo paths for that.

Otherwise, this looks good to me, you ready for it to be merged?

createyourpersonalaccount:
@bradh352 

> I'm guessing EPEL must have just moved CentOS 7 support to archive ... guess I need to look at fixing the repo paths for that.
> 
> Otherwise, this looks good to me, you ready for it to be merged?

Yes I think it's fine.

PR #934, opened 2024-12-12
include/ares.h: fix the QNX #ifdef check

bagder:
Unfortunately I have no public way to "prove" this is correct.

### DIFF
--- a/include/ares.h
+++ b/include/ares.h
@@ -74,7 +74,7 @@
 #if defined(_AIX) || defined(__NOVELL_LIBC__) || defined(__NetBSD__) || \
   defined(__minix) || defined(__SYMBIAN32__) || defined(__INTEGRITY) || \
   defined(ANDROID) || defined(__ANDROID__) || defined(__OpenBSD__) ||   \
-  defined(__QNXNTO__) || defined(__MVS__) || defined(__HAIKU__)
+  defined(__QNX__) || defined(__MVS__) || defined(__HAIKU__)
 #  include <sys/select.h>
 #endif
 


bradh352:
wonder if the methodology we use for Solaris and OpenBSD build and tests could be effective for testing QNX ... it basically boots it in a VM hosted on linux ...

bagder:
Yeah, it probably could. But the development environment is not freely available, so it would require support from Blackberry or someone else.

bagder:
I also updated the [predef listing for QNX](https://github.com/cpredef/predef/blob/master/OperatingSystems.md#QNX).

bradh352:
Thought I saw an announcement that they made it free for non-commercial use.  Granted, its still behind a "login" to download the image, but so is Solaris....

bagder:
Ah, that's more than I know. It would be really cool if we could get that up; it could help us sort of the set of out-of-tree changes in here: https://github.com/qnx-ports/c-ares

bradh352:
looking through the revisions, I still new usages of your diff of __QNXNTO__ ... that seems odd, no ?  Or are those code sections really specific to QNX6-only which uses that define?

bagder:
6.x is an out-of-date QNX version as I understand it, so it might be that it isn't tested much these days. Presumably then those ifdefs are not needed for newer versions.

bradh352:
i'm going to at least pull in autoconf archive macro updates to make your diff smaller.  I assume those are needed to support QNX properly.

I'm definitely not comfortable with the config parsers in there, however.  Also the weird sprinkled config reload stuff in a couple of places wouldn't be the proper way to do that.

bradh352:
I just got my development license for QNX, I'll take a look at it.  Doesn't look like they support direct compilation on QNX, only cross compilation, so that does make running test suites a bit harder since we'd have to use a multi-stage pipeline to cross-build the tests then boot a vm to run them ... assuming we could make it work in the first place.

bagder:
Right, but just making it build fine could be a good first step.

bradh352:
basic QNX build CI workflow has been committed. At some point we should try to start a vm and then run the tests in the vm.

PR #938, opened 2024-12-13
QNX: enhance config reading

bradh352:
QNX prefers config read from confstr(_CS_RESOLVE, ...) and confstr(_CS_DOMAIN, ...) rather than those read from /etc/resolv.conf and friends which are only used as fallbacks.  Its documented that confstr() reads from /etc/net.cfg, but that file format is NOT documented so we can't read directly from there ... however, we can monitor that file for changes to know when the system config has been updated.

This has been validated to successfully build, but has not yet had the tests run on an actual QNX platform.  Build results here:
https://github.com/bradh352/c-ares/actions/runs/12324751945/job/34402863134

### DIFF
--- a/.github/workflows/qnx.yml
+++ b/.github/workflows/qnx.yml
@@ -68,19 +68,19 @@ jobs:
           source "${{ github.workspace }}/qnx800/qnxsdp-env.sh"
           echo "QNX_TARGET=${QNX_TARGET}" >> $GITHUB_ENV
           echo "CMAKE_FIND_ROOT_PATH=${QNX_TARGET};${QNX_TARGET}/${CPUVARDIR}" >> $GITHUB_ENV
+      - name: "CMake: build c-ares"
+        env:
+          BUILD_TYPE: CMAKE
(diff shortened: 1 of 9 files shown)


bradh352:
@bagder do you think this PR will finish off your needs for QNX modifications?

bagder:
Thanks! It certainly looks like this at least fixes the build issues. I'm not completely educated about the purposes of all  the QNX specific commits, but if there are additional details that "need" to be fixed we can work on them in follow-up PRs.

bradh352:
I just ran through the test cases on a QNX vm i spun up, the only thing that failed was service name lookup due to getservbyname() not returning resuls, likely due to lack of /etc/services ... I think its a seldom-used feature, not sure the proper workaround.

That said, there's no /etc/net.cfg and therefore confstr() I don't think is being used to read the config.  I need to dig deeper on that I guess.

bradh352:
I found one bug in reading confstr, I was able to use setconf on the command line to get a string to parse.  anyhow, I think this stuff is looking good now.

PR #939, opened 2024-12-14
ares_buf: turn off buffering when loading a file

createyourpersonalaccount:
The commit is a small optimization in how a file is loaded into memory.

`FILE` buffering is just a libc (or user) provided buffer that transparently reduces (typically) the number of system calls to I/O. `fseek()` is not affected (not because `fseek()` will not perform I/O but because it won't use that buffer, apart from potentially flushing it).

In this case we predetermine the number of bytes to read (the entire file) and then we load the entire file in one `fread()`, and in this case buffering can only hurt: it results in multiple system `read()` calls instead of a single `read()`, as well as copying each read byte from the I/O buffer to the `fread()` buffer.

Furthermore, the `setvbuf()` call eagerly (before `fclose()`) releases the buffer memory back to the system (at least `BUFSIZ` >= 256 bytes). Releasing memory eagerly before performing I/O may be advantageous on a low-memory system with slow I/O.

### DIFF
--- a/src/lib/str/ares_buf.c
+++ b/src/lib/str/ares_buf.c
@@ -1384,6 +1384,11 @@ ares_status_t ares_buf_load_file(const char *filename, ares_buf_t *buf)
     }
   }
 
+  if (setvbuf(fp, NULL, _IONBF, 0) != 0) {
+    status = ARES_EFILE; /* LCOV_EXCL_LINE: DefensiveCoding */
+    goto done;           /* LCOV_EXCL_LINE: DefensiveCoding */
+  }
+
   /* Get length portably, fstat() is POSIX, not C */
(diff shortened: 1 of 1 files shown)


bradh352:
what's the motivation behind this?  I'm honestly not sure if its completely portable to turn this off and still use fseek()

createyourpersonalaccount:
@bradh352 I edited the PR message & commit message to answer your question. It's just an optional optimization I noticed while browsing the source code, it's certainly not of grave importance if you don't like it.

bradh352:
did some research to make sure there wouldn't be any system-specific side effects.  looks fine as long as setvbuf() occurs before any other file operations so going to go ahead and approve this.

PR #942, opened 2024-12-15
CI: QNX run tests in a nested VM

bradh352:
The QNX SDK has a method to generate an image and start it in a nested VM.  Use this to upload the generated test cases and run them within the VM to validate the builds.


bradh352:
@bagder wasn't that bad to run the tests in a nested VM so I went ahead and did it

PR #950, opened 2024-12-28
Fix sortlist documentation

createyourpersonalaccount:


bradh352:
sortlist is an odd one, it originated in resolv.conf for gethostbyname() manual sorting, and predates getaddrinfo() which has some built-in sorting rules in the RFC. 

It might be worth stating something similar to the above in the docs, as ares_set_sortlist() will override the resolv.conf if one exists (unlikely).

i'd really consider it deprecated, and only impacts gethostbyname() and not getaddrinfo(). Again, maybe this should be noted too.

I also don't know why it supports semi-colon separated lists.  The old version did, so I maintained it when rewriting it with the memory-safe parsers in #653.  I highly doubt anyone is using semicolons since its not documented to be valid in either c-ares nor any resolv.conf docs, I don't think we should actually document it.  You can put a comment in the code to that effect.

PR #951, opened 2024-12-29
docs: remove twice mentioned status value from list

createyourpersonalaccount:
The value ARES_EBADQUERY is mentioned already above with a better description.


createyourpersonalaccount:
I made a mistake in https://github.com/c-ares/c-ares/pull/910#issuecomment-2523084388 and added twice...

I should get troffit perhaps to generate the HTML docs myself. My bad!

PR #954, opened 2025-01-03
tests: Convert some live tests to Mock tests

### LINKED ISSUE #953: test: failure on openSUSE TW in ares-test-live.cc
[   65s] /home/abuild/rpmbuild/BUILD/c-ares-1.34.4/test/ares-test-live.cc:721: Failure
[   65s] Expected: (0) != (bitmask), actual: 0 vs 0
[   65s] 
[   65s] /home/abuild/rpmbuild/BUILD/c-ares-1.34.4/test/ares-test-live.cc:731: Failure
[   65s] Expected: ((size_t)0) != (sock_cnt), actual: 0 vs 0
[   65s] 

https://github.com/c-ares/c-ares/blob/d10a714c01762a5ee291230b7531fcfade5121dc/test/ares-test-live.cc#L717-L731

This is built in a VM with no network except for localhost

bradh352:
In Issue #953 `GetTCPSock` and `VerifySocketFunctionCallback` tests might rely on the fact that `connect()` doesn't return an immediate failure.  If `connect()` returned `EWOULDBLOCK`/`EAGAIN` the test would succeed, but on systems that check on the call for available routes or listening servers on localhost the `connect()` function would return an immediate failure.

These functions were not actually tagged with a `Live` prefix so they would run even callers exclude Live tests.  These are testing functions that shouldn't be limited to live scenarios so we need to move them to the Mock frameworks and test appropriately.

Fixes #953


bradh352:
looks like linux is smarter than I thought in `connect()`, it actually does a check to see if there is a listening daemon on the port too, not just route to host. That means this test should be moved to a Mock test where we set up a DNS responder.

PR #955, opened 2025-01-04
chore(man): use exactly same copyright string as for sources

gjasny:
The (automatically) [generated Debian Copyright file](https://salsa.debian.org/debian/c-ares/-/blob/debian/1.34.4-2.1/debian/copyright) contains more sections than necessary due an inconsistent copyright string:

```
The c-ares project and its contributors[.]
```

While it's not a big deal due to the file is generated automatically, it would be nice to shrink the relatively complex file.

Maybe also the copyrights by "Brad House" and "Daniel Stenberg" could be replaced with "The c-ares project and its contributors".

Thanks for considering.
-Gregor

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -5,17 +5,17 @@ Source: https://c-ares.org/
 
 # Fuzzer data
 Files: test/fuzzinput/* test/fuzznames/*
-Copyright: The c-ares project and its contributors.
+Copyright: The c-ares project and its contributors
 License: MIT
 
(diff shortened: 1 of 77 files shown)


bradh352:
I'm willing to remove my name, the problem is the other copyright holders.  They've all licensed their contributions to modern MIT, but they are still the copyright holders and to remove their name from the headers, not sure how that works legally even with attribution in the AUTHORS file.

PR #958, opened 2025-01-08
Port: restore Windows XP minimal support

### LINKED ISSUE #956: Disable `NotifyIpInterfaceChange` symbol check in cmake (for windows XP target)
Hello!

I tried to use c-ares for windows-XP (only DNS-parsing capabilities are needed, no any I/O or syscalls), but, unfortunately, when I cross-compile it on linux (via mingw), it locates the `NotifyIpInterfaceChange` symbol , and then the compiled dll cannot be loaded on windows-xp as this symbol is missing.

```
-- Looking for NotifyIpInterfaceChange
-- Looking for NotifyIpInterfaceChange - found
```

I have tried

```
cmake "-DCFLAGS=-DWIN32_WINNT=0x0501" ...
```

which should strip-off the symbol from mingw-headers, but, as far as I notices, `CmakeLists.txt` defines this symbol by force.

What can be done?

bradh352:
Windows XP support was removed in 2021, but we've had a recent request for basic Windows XP support.  It looks like the needed changes are very minimal these days, the only real requirement is to build without threading support, everything else can be conditionally bypassed.

This changeset mostly just disables building the event subsystem if threading is disabled then makes a couple minor build system changes to allow overriding of the `_WIN32_WINNT` macro.

Fixes #956


bradh352:
needed to create this as a PR to spawn off AppVeyor tests that actually test building of Windows XP support

PR #961, opened 2025-01-16
Add New Hint in AddrInfo to send back QueryId in getaddrinfo 

arnnav:
Adding a new hint in `ares_addrinfo_hints` which will enable sending back query id as part of the `ares_addrinfo` in the call back of `getaddrinfo` .


bradh352:
a few comments here.
1. You can't modify the hints structure since a newer c-ares version would try to read out of bounds if the app was originally linked against an older version.  This breaks ABI.  Adding a new ai_flag would accomplish the same thing without ABI breakage.
2. qid is 16 bit, so shouldn't use int ... though that may not be the best thing to use in the first place.
3. maybe it would be better to optionally attach the returned dnsrec entries to the addrinfo result, that would contain the qid obviously as well as other metadata not typically present that might be desirable (such as the 'additional' records that may be attached)
4. @bagder recently asked (on the mailing list) about adding the ability to know if the result was served out of /etc/hosts vs a query, so if requesting the full record be attached and there wasn't one, that would indicate it was served via /etc/hosts.

arnnav:
Thanks @bradh352 for the comments. If I understand correctly, your comments are suggesting the following:

1. No modification to the `ares_addrinfo_hints` structure.
2. Do not make qid part of the `ares_addrinfo` structure.
3. Update the `ares_addrinfo` structure to have the dnsrec entires (`dnsrec_a`, `dnsrec_aaaa`). Pass that back as part of the addrinfo result. 
4. Based on the family specified in the `ares_addrinfo_hints`, the existence of dnsrec entries can determine if the requests were served out of /etc/hosts or a query.

I will go ahead and make these changes. 
Please let me know if I missed something or understood incorrectly. 
Thanks!

bradh352:
I think the missing thing is adding a new ai_flag to indicate if the dnsrec_a and dnsrec_aaaa should be attached to the answer or not.  Don't know if it makes sense to always attach ... I'd need to look at it closer I guess, I don't recall if it would be necessary to duplicate the response into the structure or if we can just attach it to the structure as a way to delay its destruction.

arnnav:
I think adding a new ai_flag can be beneficial, if the app is not interested in the dnsrec entries then c-ares should not be passing that information back.

PR #965, opened 2025-01-26
cmake: Fix exporting -pthread to pkgconfig

dg0yt:
Before this change, libcares.pc contained ` -l-pthread`.  
Noticed in the vcpkg port.


bradh352:
I'm assuming that means THREADS_PREFER_PTHREAD_FLAG is being set somewhere, since that's not otherwise the default behavior in cmake.

Fix seems reasonable though.

dg0yt:
> I'm assuming that means THREADS_PREFER_PTHREAD_FLAG is being set somewhere, since that's not otherwise the default behavior in cmake.

It is recommended, and it is there:
https://github.com/c-ares/c-ares/blob/8776bf5c15bf2f6ad4694e9b5e879be53102dd4f/CMakeLists.txt#L495

bradh352:
odd that docs show `THREADS_PREFER_PTHREAD_FLAG ` as the variable name. https://cmake.org/cmake/help/latest/module/FindThreads.html

dg0yt:
You are right, it is a different identifier.

The issue came up in the android builds. Maybe it came from the toolchain. THREADS_PREFER_PTHREAD_FLAG is more a toolchain setting than a project setting - users probably want the same configuration for a given target platform.

PR #975, opened 2025-03-26
ares_thread: cond_waitno monotonic clock if possible

crrodriguez:
The realtime clock is subject to administrative changes like NTP.

There is a newer, less verbose _clockwait() API in recent systems but the old condattr_setclock api is more widely available and that is what this patch uses for compatibility.


crrodriguez:
ok.. macos does not follow posix.. what a suprise.. noeed to add autoconf checks anyway..

bradh352:
seems reasonable, guess I need to fix the freebsd image for building that and on windows set google test to a specific tag since its now requiring c++17

crrodriguez:
Ok, I fixed it by excluding apple stuff that should not define macros telling it has this feature but it hasn't. otherwise it would involve adding extra cmake/autoconf macros for very little gain.

bradh352:
If you update your branch the CI builds should work.

crrodriguez:
ci passes except unrelated failure.

src/lib/util/ares_threads.c
```diff
@@ -487,6 +487,7 @@ void ares_thread_mutex_unlock(ares_thread_mutex_t *mut)
 
 struct ares_thread_cond {
   pthread_cond_t cond;
+  pthread_condattr_t attr;
```
bradh352, line 490: The pthread_condattr_t doesn't need to be saved after you've passed it to pthread_cond_init(), so please just make it a local scoped variable instead of saving it.

bradh352 (changes requested):

src/lib/util/ares_threads.c
crrodriguez: Yes, i'll fix that. thanks

bradh352: ping

PR #992, opened 2025-05-12
Changes to server selection and polling behaviour

### LINKED ISSUE #987: More fine-grained control of server priority list
Currently a server will be prioritised/deprioritised based on the success/failure of a single query. It would be useful to us to be able to have more fine-grained control to allow tuning of the priority list based on the network in which it is used. Specifically I am looking to prioritise/deprioritise based on _n_ consecutive successes/failures.

Examples:
- If servers can sometimes be unreliable, re-prioritising a server after a single success could lead to it to being repeatedly re-added to the rotation and used for requests that it frequently fails.  Requiring (for example) 3 successive requests to succeed would provide more confidence that it's fully working before starting using it again.
- In a network with a local primary DNS server and a remote backup, switching to the backup on a single failure may lead to brief glitches leading to longer periods of higher latency.  Requiring (for example) 3 successive requests to fail would ensure that it's not switching unnecessarily.

I am happy to implement above enhancement but I want to ensure that the spirit of the enhancements fits the overall design for the library.

DavidMcNallyAlianza:
Fixes #987

This PR contains the following changes:
- The addition of configurable rise/fall server behaviour (only mark servers as up or down after a certain number of successes or failures) (this is issue #987) 
- Changes to server selection when all servers are failed, instead of picking the highest priority server at all times, a query will prioritise trying all failed servers at least once.
- Polling now polls all failed servers that are outside of the timeout rather than just the highest priority one.

### DIFF
--- a/include/ares.h
+++ b/include/ares.h
@@ -547,6 +547,12 @@ CARES_EXTERN void ares_getaddrinfo(ares_channel_t *channel, const char *node,
 
 CARES_EXTERN void ares_freeaddrinfo(struct ares_addrinfo *ai);
 
+CARES_EXTERN void ares_set_min_server_successes(ares_channel_t *channel,
+                                                unsigned int minimum_successes);
+
+CARES_EXTERN void ares_set_max_server_failures(ares_channel_t *channel,
+                                               unsigned int maximum_failures);
+
 /*
  * Virtual function set to have user-managed socket IO.
  * Note that all functions need to be defined, and when
--- a/include/ares.h
+++ b/include/ares.h
@@ -241,30 +241,31 @@ typedef enum {
 #define ARES_FLAG_DNS0x20     (1 << 10)
 
 /* Option mask values */
-#define ARES_OPT_FLAGS           (1 << 0)
-#define ARES_OPT_TIMEOUT         (1 << 1)
-#define ARES_OPT_TRIES           (1 << 2)
-#define ARES_OPT_NDOTS           (1 << 3)
-#define ARES_OPT_UDP_PORT        (1 << 4)
-#define ARES_OPT_TCP_PORT        (1 << 5)
-#define ARES_OPT_SERVERS         (1 << 6)
-#define ARES_OPT_DOMAINS         (1 << 7)
-#define ARES_OPT_LOOKUPS         (1 << 8)
-#define ARES_OPT_SOCK_STATE_CB   (1 << 9)
-#define ARES_OPT_SORTLIST        (1 << 10)
-#define ARES_OPT_SOCK_SNDBUF     (1 << 11)
-#define ARES_OPT_SOCK_RCVBUF     (1 << 12)
-#define ARES_OPT_TIMEOUTMS       (1 << 13)
-#define ARES_OPT_ROTATE          (1 << 14)
-#define ARES_OPT_EDNSPSZ         (1 << 15)
-#define ARES_OPT_NOROTATE        (1 << 16)
-#define ARES_OPT_RESOLVCONF      (1 << 17)
-#define ARES_OPT_HOSTS_FILE      (1 << 18)
-#define ARES_OPT_UDP_MAX_QUERIES (1 << 19)
-#define ARES_OPT_MAXTIMEOUTMS    (1 << 20)
-#define ARES_OPT_QUERY_CACHE     (1 << 21)
-#define ARES_OPT_EVENT_THREAD    (1 << 22)
-#define ARES_OPT_SERVER_FAILOVER (1 << 23)
+#define ARES_OPT_FLAGS            (1 << 0)
+#define ARES_OPT_TIMEOUT          (1 << 1)
+#define ARES_OPT_TRIES            (1 << 2)
+#define ARES_OPT_NDOTS            (1 << 3)
+#define ARES_OPT_UDP_PORT         (1 << 4)
+#define ARES_OPT_TCP_PORT         (1 << 5)
+#define ARES_OPT_SERVERS          (1 << 6)
+#define ARES_OPT_DOMAINS          (1 << 7)
+#define ARES_OPT_LOOKUPS          (1 << 8)
+#define ARES_OPT_SOCK_STATE_CB    (1 << 9)
+#define ARES_OPT_SORTLIST         (1 << 10)
+#define ARES_OPT_SOCK_SNDBUF      (1 << 11)
+#define ARES_OPT_SOCK_RCVBUF      (1 << 12)
+#define ARES_OPT_TIMEOUTMS        (1 << 13)
+#define ARES_OPT_ROTATE           (1 << 14)
+#define ARES_OPT_EDNSPSZ          (1 << 15)
+#define ARES_OPT_NOROTATE         (1 << 16)
+#define ARES_OPT_RESOLVCONF       (1 << 17)
+#define ARES_OPT_HOSTS_FILE       (1 << 18)
+#define ARES_OPT_UDP_MAX_QUERIES  (1 << 19)
+#define ARES_OPT_MAXTIMEOUTMS     (1 << 20)
+#define ARES_OPT_QUERY_CACHE      (1 << 21)
+#define ARES_OPT_EVENT_THREAD     (1 << 22)
+#define ARES_OPT_SERVER_FAILOVER  (1 << 23)
+#define ARES_OPT_SERVER_RISE_FALL (1 << 24)
 
 /* Nameinfo flag values */
 #define ARES_NI_NOFQDN        (1 << 0)
@@ -345,10 +346,16 @@ struct apattern;
  * to.
  * The retry delay is the minimum time in milliseconds to wait between doing
  * such retries (applied per-server).
+ * The min_consec_successes is the minimum number of consecutive successful
+ * queries before a server is considered "up" again.
+ * The max_consec_failures is the maximum number of consecutive failed
+ * queries before a server is considered "down".
  */
 struct ares_server_failover_options {
   unsigned short retry_chance;
   size_t         retry_delay;
+  size_t         min_consec_successes;
+  size_t         max_consec_failures;
 };
 
 /* NOTE about the ares_options struct to users and developers.
--- a/src/lib/ares_init.c
+++ b/src/lib/ares_init.c
@@ -96,10 +96,21 @@ static int server_sort_cb(const void *data1, const void *data2)
   const ares_server_t *s1 = data1;
   const ares_server_t *s2 = data2;
 
-  if (s1->consec_failures < s2->consec_failures) {
+  /* The logic for sorting is as follows:
+   * 1. Failed servers are sorted to the end of the list.
+   * 2. Among non-failed servers, sort by index in the original list.
+   * 3. Among failed servers, sort by consecutive failures then index.
+   */
+  if (!s1->is_failed && s2->is_failed) {
     return -1;
   }
-  if (s1->consec_failures > s2->consec_failures) {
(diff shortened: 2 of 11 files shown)


src/lib/ares_process.c
```diff
@@ -1247,12 +1356,7 @@ ares_status_t ares_send_query(ares_server_t *requested_server,
     server = requested_server;
   } else {
     /* If rotate is turned on, do a random selection */
```
DavidMcNallyAlianza, line 1352: Remove comment

src/lib/ares_init.c
```diff
@@ -96,18 +96,36 @@ static int server_sort_cb(const void *data1, const void *data2)
...
-  }
-  if (s1->consec_failures > s2->consec_failures) {
-    return 1;
-  }
-  if (s1->idx < s2->idx) {
-    return -1;
-  }
-  if (s1->idx > s2->idx) {
-    return 1;
+
```
thomaswhiteway, line 99: I think this needs a comment to justify the logic.  I think it might also be clearer to structure this as
* compare `is_failed`
* if not `is_failed` compare `consec_failures`
* compare `is_idx`

Currently I think the logic looks a bit more complex than it actually is as the "compare `is_failed`" section is split up, and the "compare `is_idx` " section is duplicated.

src/lib/ares_process.c
```diff
@@ -147,6 +155,16 @@ static void server_set_good(ares_server_t *server, ares_bool_t used_tcp)
 
   if (server->consec_failures > 0) {
     server->consec_failures = 0;
+    reinsert                = ARES_TRUE;
+  }
+
+  server->consec_successes++;
+  if (server->consec_successes >= channel->min_consec_successes) {
```
thomaswhiteway, line 162: Should this also check if `server->is_failed` is false?  Currently, if a server is up and working, you'll be removing it from the list and reinserting it after every request (after the first `channel->min_consec_successes`) despite its position not changing.

src/lib/ares_init.c
```diff
@@ -598,3 +618,25 @@ int ares_set_sortlist(ares_channel_t *channel, const char *sortstr)
   ares_channel_unlock(channel);
   return (int)status;
 }
+
+void ares_set_min_server_successes(ares_channel_t *channel,
```
thomaswhiteway, line 622: I'd assumed these options would be added to `struct ares_options` rather than having separate function calls to set them, is there a particular reason for doing it this way?

src/lib/ares_init.c
DavidMcNallyAlianza: That's significantly cleaner, thanks!

src/lib/ares_init.c
DavidMcNallyAlianza: From the description of the `ares_options` struct it looks like that isn't the pattern that they want anymore, I will leave it to the maintainer to comment if there is something else I should be doing

```
     When new options are added to c-ares, they are not added to this
     struct. And they are not "saved" with the ares_save_options() function but
     instead we encourage the use of the ares_dup() function. Needless to say,
     if you add config options to c-ares you need to make sure ares_dup()
     duplicates this new option.
```

bradh352:
I will review this weekend. 

I did see a comment related to configuration, its ok to add members to the end of the ares_options as long as they aren't members that require allocation/deallocation since ares_destroy_options doesn't take the optmask.  I think the comment found was more aspirational than practical and predates me being a maintainer ... we've added lots of options to ares_options since then.

We have a ticket to redesign the whole bringup / settings, I'd rather not add helper functions to set new settings since it will make more work for that in the future.

DavidMcNallyAlianza:
Hi @bradh352, I would ideally get this change into the next c-ares release, do you know when you will be able to review?

DavidMcNallyAlianza:
Hi @bradh352, sorry to bother you again but would it be possible for you to let me know when you will be able review this change?

DavidMcNallyAlianza:
Hi Brad, do you think you will be able to review sometime soon? I totally get being busy, just an idea of timeframe would be great.

include/ares.h
```diff
@@ -345,10 +346,16 @@ struct apattern;
...
+ * queries before a server is considered "up" again.
+ * The max_consec_failures is the maximum number of consecutive failed
+ * queries before a server is considered "down".
  */
 struct ares_server_failover_options {
   unsigned short retry_chance;
   size_t         retry_delay;
+  size_t         min_consec_successes;
+  size_t         max_consec_failures;
 };
```
bradh352, line 359: this is sort of an odd approach to add a new ARES_OPT but then modify an existing option.  I think I'd rather just a setter/getter approach rather than using ares_init_options() for this.  Infact, I hope to remove ares_init_options() one day.

src/lib/ares_process.c
```diff
@@ -886,12 +905,18 @@ static ares_status_t process_answer(ares_channel_t      *channel,
...
+  server_increment_successes(server, query->using_tcp);
   end_query(channel, server, query, ARES_SUCCESS, rdnsrec);
 
   status = ARES_SUCCESS;
 
 cleanup:
+  /* If we were probing for the server to come back online, lets mark it as
+   * no longer being probed */
+  if (server != NULL) {
+    server->probe_pending = ARES_FALSE;
```
bradh352, line 917: this is done universally by end_query(), was there an instance you found this not being called?  What was the call path for that?

src/lib/ares_process.c
```diff
@@ -1409,12 +1529,6 @@ static void end_query(ares_channel_t *channel, ares_server_t *server,
                       ares_query_t *query, ares_status_t status,
                       const ares_dns_record_t *dnsrec)
 {
-  /* If we were probing for the server to come back online, lets mark it as
-   * no longer being probed */
-  if (server != NULL) {
-    server->probe_pending = ARES_FALSE;
```
bradh352, line 1415: why was this moved?

src/lib/ares_private.h
```diff
@@ -181,6 +184,8 @@ struct ares_query {
   size_t        timeouts;   /* number of timeouts we saw for this request */
   ares_bool_t   no_retries; /* do not perform any additional retries, this is
                              * set when a query is to be canceled */
+  ares_array_t *failed_servers_attempted; /* Failed servers attempted to be used
```
bradh352, line 187: why do we need to track this?  the prior logic would choose it in highest priority order.  failed servers wouldn't be probed until the timer expired for retrying them.

src/lib/ares_process.c
```diff
@@ -147,6 +155,17 @@ static void server_set_good(ares_server_t *server, ares_bool_t used_tcp)
 
   if (server->consec_failures > 0) {
     server->consec_failures = 0;
```
bradh352, line 157: should not unset consec_failures until consec_successes passes threshold.  again, we don't want to track with yet another variable of is_failed.

src/lib/ares_init.c
```diff
@@ -96,10 +96,21 @@ static int server_sort_cb(const void *data1, const void *data2)
...
+    return -1;
+  }
+  if (s1->is_failed && !s2->is_failed) {
+    return 1;
+  }
+  if (s1->is_failed && s1->consec_failures < s2->consec_failures) {
     return -1;
   }
-  if (s1->consec_failures > s2->consec_failures) {
+  if (s1->is_failed && s1->consec_failures > s2->consec_failures) {
```
bradh352, line 113: sort should now likely be:
```c
if (s1->consec_failures == 0 && s2->consec_failures > 0) {
  return -1;
}
if (s2->consec_failures == 0 && s1->consec_failures > 0) {
  return 1;
}
if (s2->consec_failures && s1->consec_failures) {
  /* See if connection is coming back online but hasn't hit the rise limit */
  if (s1->consec_successes > s2->consec_successes) {
    return -1;
  }
  if (s1->consec_successes < s2->consec_successes) {
    return 1;
  }
}
if (s1->consec_failures < s2->consec_failures) {
  return -1;
}
if (s1->consec_failures > s2->consec_failures) {
  return 1;
}
```

bradh352:
sorry for the long delay here, i just did a review of this.  I think it could be greatly simplified by not adding an `is_failed` flag since it can be trivially calculated, and to better rely on the server sort functionality that is there rather than trying to come up with a new algorithm which adds a lot of complexity and tracking.   The server sorting, which would re-sort based on each additional consecutive failure or consecutive success should accomplish exactly what you want without those needed functions.

Flynnzaa:
Hi @bradh352, I appreciate the comment above is a big wall of text, but it's been a few months so I just wanted to ask if you might have a response for me on this? Especially the question of per-request server retry distribution.

bradh352:
Sorry for the long delay in getting back to you on this. I went through your three concerns in detail, and the sticky-recovery scenario in your diagram is real — I traced it against current main and it actually exists today independent of this PR: probing is gated on the *triggering* query having gone to a healthy server, so in the all-servers-failed case no probes are ever sent, and the auto-rotate effect I described in my earlier review only holds while failure counts stay within ±1 of each other. A recovery blip breaks that symmetry exactly as your diagram shows. Thank you for the writeup and the diagram — that's what pinned it down.

Since it's a live bug independent of the rise/fall feature, I've fixed it directly on main rather than folding it into this PR: **#1192** (credited to you both in the description). It takes the "probe extra servers" + "minimise the failure counts down to a cap" direction you suggested as alternatives, specifically:

1. Probes of downed servers are now spawned even when the triggering query itself went to a failed server (the probe scan skips the triggering server since that query already exercises it). In your 1 qps scenario, B gets probed within ~`retry_chance` (default 10) queries of recovering, i.e. seconds instead of the outage-length catch-up window.
2. `consec_failures` is capped at 16 — only relative order matters for selection, so magnitude beyond "clearly down" carries no signal, and the cap bounds the stale-counter divergence as a backstop.
3. Sort ties on equal failure counts now break by `next_retry_time` (least-recently-failed first) instead of configuration order. This keeps retries rotating once counts saturate at the cap, and it means a stale failed server that may have quietly recovered gets tried before one that just failed — which resolves your scenario even for `retry_chance = 0` configurations where probing can't help.

Could you review #1192 and confirm it covers your first concern (including the variants you were worried about)? Note it composes with the rise threshold from this PR: once N consecutive successes are required to mark a server healthy, the probe path is what delivers those N successes to a server the sort would otherwise never select.

On your second and third concerns: you're right that threshold-aware sorting trades away the per-request retry-distribution guarantee (and your ~3% estimate checks out — (3/4)^12). Rather than reintroducing per-query attempted-server tracking, I'd suggest using state we already have: when picking a server for a *retry* (`try_count != 0`), start the scan at offset `try_count % num_servers` into the sorted list instead of always at the head. Fresh queries still take the head, so the rise/fall semantics apply to new traffic exactly as intended, but a single request's retries spread across servers — eliminating both the deterministic worst case (all tries burned on one sub-threshold server) and the probabilistic one, without allocations or a new option. It's not perfectly collision-free if the list re-sorts mid-request, but the worst case is one wasted retry rather than a failed request. If that works for you, this PR reduces to: the rise/fall counters + thresholds, the sort changes from my earlier review, and the retry-offset — no `is_failed`, no per-query server memory.

On the `probe_pending`/SERVFAIL bug: a probe-timeout test (`ProbeAfterProbeTimeout`) has landed on main since your branch's base, and on current main I believe the SERVFAIL path also clears the flag on every route (`ares_requeue_query()` captures the server before detaching from the connection, and `end_query()` unconditionally clears it) — so it's possible the requeue rearchitecture fixed this since your base. Could you rebase your test onto current main and confirm whether it still reproduces? If it does, that's a bug I want fixed regardless of this PR.

bradh352:
closing in favor of #1192

PR #994, opened 2025-05-13
Use xor for random state generation

pps83:
It doesn't make sense to add lots of random sources and OR them, as this will lead to all bits set.

### DIFF
--- a/src/lib/util/ares_rand.c
+++ b/src/lib/util/ares_rand.c
@@ -56,7 +56,7 @@ static unsigned int ares_u32_from_ptr(void *addr)
 {
   /* LCOV_EXCL_START: FallbackCode */
   if (ares_is_64bit()) {
-    return (unsigned int)((((ares_uint64_t)addr >> 32) & 0xFFFFFFFF) |
+    return (unsigned int)((((ares_uint64_t)addr >> 32) & 0xFFFFFFFF) ^
                           ((ares_uint64_t)addr & 0xFFFFFFFF));
   }
   return (unsigned int)((size_t)addr & 0xFFFFFFFF);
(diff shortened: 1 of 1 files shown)


pps83:
the reason I looked into it all was because of a warning I'm getting in 32bit builds:

```
In file included from D:/work-pps/c-ares/libcares.c:155:
D:/work-pps/c-ares/src/lib/util/ares_rand.c: In function 'ares_u32_from_ptr':
D:/work-pps/c-ares/src/lib/util/ares_rand.c:59:29: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   59 |     return (unsigned int)((((ares_uint64_t)addr >> 32) & 0xFFFFFFFF) |
      |                             ^
D:/work-pps/backtest-engine/ext/c-ares/src/lib/util/ares_rand.c:60:28: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   60 |                           ((ares_uint64_t)addr & 0xFFFFFFFF));
      |                            ^

```

bradh352:
you are right, that should be XOR

pps83:
Also, the code uses that `ares_u32_from_ptr`. I'd simply drop that ares_is_64bit part and simply use `return (unsigned int)((size_t)addr & 0xFFFFFFFF);` regardless of platform.

dfandrich:
How are these random numbers used? It sounds like the seed will not infrequently end up being or almost being a constant 0xFFFFFFFF. That might make it a CVE-worthy issue, especially if it results in something similar to security/advisories/GHSA-8r8p-23f3-64c2.

pps83:
> How are these random numbers used? It sounds like the seed will not infrequently end up being or almost being a constant 0xFFFFFFFF. That might make it a CVE-worthy issue, especially if it results in something similar to security/advisories/[GHSA-8r8p-23f3-64c2](https://github.com/c-ares/c-ares/security/advisories/GHSA-8r8p-23f3-64c2).

no idea, I just saw the warning and wanted to get rid of it. I any case, it seems like that part is some sort of fallback (based on the comments)

bhouse-nexthop:
I'm honestly not sure what platform that fallback code actually gets used on and under what conditions.

Previously it was easy to get c-ares to use that old fallback (at least on non-windows platforms).  But we added in support for getrandom() and arc4random() which at least one of those is provided by most systems, and if not /dev/urandom is.  

Maybe its used on DOS?

PR #996, opened 2025-05-15
Fix builds with clang on windows

seanharmer:
We are using c-ares in a project and hit issues with running clang-format as part of our CI on Windows. These patches now result in a clean build when using llvm/clang tooling.

### DIFF
--- a/src/lib/util/ares_math.h
+++ b/src/lib/util/ares_math.h
@@ -26,13 +26,9 @@
 #ifndef __ARES_MATH_H
 #define __ARES_MATH_H
 
-#ifdef _MSC_VER
-typedef __int64          ares_int64_t;
-typedef unsigned __int64 ares_uint64_t;
-#else
-typedef long long          ares_int64_t;
-typedef unsigned long long ares_uint64_t;
-#endif
+#include <stdint.h>
+typedef int64_t  ares_int64_t;
+typedef uint64_t ares_uint64_t;
 
 ares_bool_t   ares_is_64bit(void);
 size_t        ares_round_up_pow2(size_t n);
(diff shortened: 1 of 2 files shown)


src/lib/util/ares_math.h
```diff
@@ -26,13 +26,9 @@
...
-#ifdef _MSC_VER
-typedef __int64          ares_int64_t;
-typedef unsigned __int64 ares_uint64_t;
-#else
-typedef long long          ares_int64_t;
-typedef unsigned long long ares_uint64_t;
-#endif
+#include <stdint.h>
+typedef int64_t  ares_int64_t;
+typedef uint64_t ares_uint64_t;
```
bhouse-nexthop, line 32: We can't guarantee stdint exists unfortunately.  We need to support C89.

seanharmer: Is the C89 support required on Windows too or can we assume C99 there? It's only on windows that clang complains.

Failing that, we would have to suppress the warning.

bhouse-nexthop: Microsoft's compilers didn't support C99 until fairly late.  So if someone was trying to use MSVC6 to target say WindowsXP it would fail.

seanharmer: OK no problem. I'll revert this commit and do another to silence the problematic warning from clang on Windows then.

bhouse-nexthop: you could do something like `#elif defined(HAVE_STDINT_H)` to add that codepath.  Just don't remove the other.

seanharmer: I'm not clear why the test is failing (https://cirrus-ci.com/task/6083938910732288) given it worked on the previous commit. Is it anything obvious to you?

seanharmer:
Does this look OK to you now please?

PR #1002, opened 2025-07-03
Add the option to disable system configuration monitoring

oargon:
I have seen a case on iOS that ends up failing to set up system change notification. As opposed to the general Apple initialization, where c-ares can continue if `ares_init_by_sysconfig` did not succeed, failure to set up system change notifications would fail the initialization of c-ares.
Specifically for my use case, I not actually interested in notifications for system DNS configuration changes, in a similar manner to how the socket callback option works.
 
I suggest to have a way to disable this automatic re-initialization with an options flag.
Alternatively, it might be good to allow c-ares to continue initialization if this setup failed.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -258,3 +258,4 @@ however on other unix-like systems a polling mechanism is used that checks every
 
 This feature requires the c-ares channel to persist for the lifetime of the
 application.
+If you want to disable this feature, you can set `ARES_OPT_DISABLE_MONITOR`.
--- a/docs/ares_init_options.3
+++ b/docs/ares_init_options.3
@@ -328,7 +328,8 @@ any events on the file descriptors, so \fIares_process(3)\fP nor
 \fIares_process_fd(3)\fP should ever be called when this option is enabled.
 
 As of c-ares 1.29.0, when enabled, it will also automatically re-load the
-system configuration when changes are detected.
+system configuration when changes are detected, unless
+\ARES_OPT_DISABLE_MONITOR\fP is set to explicitly disable this feature.
 
 Use \fIares_threadsafety(3)\fP to determine if this option is available to be
 used.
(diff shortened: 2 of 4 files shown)


bhouse-nexthop:
I'd rather not add a configuration option for that but handle odd failure scenarios better.  We could just omit monitoring automatically on windows if the registry keys aren't available.  I think the wine case is a fairly unique one.  Does the event thread actually function there?  If so, that's impressive since it uses some undocumented windows features.

oargon:
@bhouse-nexthop 
Thanks for you comment.

My use case is on Apple and I don't know why it fails.
Would a silent failure, not listening to changes be a better option?

bradh352:
what version of MacOS are you on?  I'm pretty sure I've tried a large subset of them, my daily driver is currently a Mac.  It could be something about your system config.  Can you isolate which function call is failing that is causing the error?  Its better to fix these things if possible, but yeah, we could make config monitoring in general non-critical for event thread to only use it if it doesn't fail to start up.

oargon:
> what version of MacOS are you on?
I'm running a unit test on iPhone simulator, using iOS 18.5.
The failure is because calling `notify_register_file_descriptor` does not return `NOTIFY_STATUS_OK` as expected here: https://github.com/c-ares/c-ares/blob/71663adcd95acdb1c198806d626dfc8640d132be/src/lib/event/ares_event_configchg.c#L506
It returns `NOTIFY_STATUS_OPT_DISABLE`.
These unit tests, run in a separate remote environment on a CI, locally it seems to work.

In the unit tests, the initial setup in `ares_init_sysconfig_macos` also fails, here: https://github.com/c-ares/c-ares/blob/71663adcd95acdb1c198806d626dfc8640d132be/src/lib/ares_sysconfig_mac.c#L337

I'm not certain why it fails. However, because I don't really need the feature of reloading configurations and don't really need the system configuration at all, I'm more interested in having it not fail or disabled. Similar to how the init fails silently.

oargon:
I opened another PR for the other solution.
https://github.com/c-ares/c-ares/pull/1009

dankmeme01:
> I think the wine case is a fairly unique one. Does the event thread actually function there?

By the way, yes, I was very surprised as well to learn that it works just fine if you manually create the registry keys

PR #1009, opened 2025-07-19
Ignore ares_event_configchg_init failures

oargon:
As discussed on https://github.com/c-ares/c-ares/pull/1002, ignore failure when trying to set up configuration notifications.

Similar to failures when calling `ares_init_by_sysconfig` https://github.com/c-ares/c-ares/blob/71663adcd95acdb1c198806d626dfc8640d132be/src/lib/ares_init.c#L335

### DIFF
--- a/src/lib/ares_init.c
+++ b/src/lib/ares_init.c
@@ -362,7 +362,8 @@ int ares_init_options(ares_channel_t           **channelptr,
     e      = channel->sock_state_cb_data;
     status = ares_event_configchg_init(&e->configchg, e);
     if (status != ARES_SUCCESS && status != ARES_ENOTIMP) {
-      goto done; /* LCOV_EXCL_LINE: UntestablePath */
+      DEBUGF(fprintf(stderr, "Error: ares_event_configchg_init failed: %s\n",
+                     ares_strerror(status)));
     }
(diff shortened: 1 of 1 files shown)


oargon:
> did you happen to determine what function your failure occurred in?

I wrote a comment on the original PR. See https://github.com/c-ares/c-ares/pull/1002#issuecomment-3092500898

> If you keep the check as `if (status != ARES_SUCCESS && status != ARES_ENOTIMP)` rather than just != success to print the warning, I'll accept the PR.

Thanks. I updated the PR to reflect your suggestion.

bradh352:
> > did you happen to determine what function your failure occurred in?
> 
> I wrote a comment on the original PR. See [#1002 (comment)](https://github.com/c-ares/c-ares/pull/1002#issuecomment-3092500898)
> 
> > If you keep the check as `if (status != ARES_SUCCESS && status != ARES_ENOTIMP)` rather than just != success to print the warning, I'll accept the PR.
> 
> Thanks. I updated the PR to reflect your suggestion.

Ah, I missed that.  Ok, so this is mostly in the simulator/testing framework that apparently its simulated/emulated in such a way that there is missing functionality.  Fair enough.

PR #1010, opened 2025-07-24
Restore compatibility with Microsoft GDK

sykhro:
Add a new symbol check for `RegNotifyChangeKeyValue` and use it to guard `ares_event_configchg_regnotify`.
This is necessary to compile for flavors of Windows that don't implement the `DESKTOP` WinAPI partition (e.g. Microsoft GDK aka `WINAPI_PARTITION_GAMES`).

Avoids macro checks with `WINAPI_PARTITION` to preserve compatibility with WinSDK < 10.


bradh352:
i know it sounds weird, but can you also update the configure.ac with the same change?  We do actually have a few windows users that use that.

PR #1021, opened 2025-08-16
Fix DNS server fetching for users with non-ASCII characters

### LINKED ISSUE #1016: `c-ares` fails to resolve for some users while getaddrinfo works

```

;; QUESTION SECTION:

;; MSG SIZE  rcvd: 0
```

```

DNS servers: 
127.0.0.1:53

```

MichalPetryka:
`c-ares` is currently broken when user configuration contains non-ASCII characters, remove the useless checks to make it work.

Fixes #1016.

### DIFF
--- a/src/lib/str/ares_buf.c
+++ b/src/lib/str/ares_buf.c
@@ -438,7 +438,6 @@ ares_status_t ares_buf_tag_fetch_string(const ares_buf_t *buf, char *str,
 {
   size_t        out_len;
   ares_status_t status;
-  size_t        i;
 
   if (str == NULL || len == 0) {
     return ARES_EFORMERR;
@@ -454,13 +453,6 @@ ares_status_t ares_buf_tag_fetch_string(const ares_buf_t *buf, char *str,
 
   /* NULL terminate */
   str[out_len] = 0;
-
-  /* Validate string is printable */
-  for (i = 0; i < out_len; i++) {
-    if (!ares_isprint(str[i])) {
-      return ARES_EBADSTR;
-    }
-  }
 
   return ARES_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


MichalPetryka:
There are more places that use `ares_isprint` and could possibly be broken for non-ASCII text but this is enough to fix the issue at hand.

Every single usecase of the macro should however be reviewed in the future and the macro itself should be renamed to `ares_isasciichar` or something.

MichalPetryka:
> DNS in general is ascii only, punycode is used for non-ascii representations. Can you explain this change. It will likely cause all sorts of issues (security or otherwise) to change this in the CORE parsing library which expects ONLY ascii.

Affected users are getting non-ASCII values being appended [here](https://github.com/c-ares/c-ares/blob/ee2a1c3eff3c8164b09123005f4b49c571788b59/src/lib/ares_sysconfig_win.c#L580) which then makes the split code fail.

I believe the split helper by itself shouldn't fail for non-ASCII text, there's no reasonable explanation for such behaviour that I can think of without it having a separate "require ASCII" flag, so I've made this change that makes split no longer fail for non ASCII.

gvanem:
> *"user configuration contains non-ASCII characters"*

Like where?

bradh352:
So the DhcpDomain registry key is non-ascii.  That means the proper fix would be to punycode the value before joining.

MichalPetryka:
> So the DhcpDomain registry key is non-ascii. That means the proper fix would be to punycode the value before joining.

I'm not sure if that's the only one that can have such but one log I've got had it as such, the change here fixed it for other users though.

Still having the splitting reject non ASCII was very unexpected and made figuring the issue out very annoying even when I had the problematic value. If such behaviour is desired, I'd really recommend adding a separate ASCII_ONLY flag to the split flags.

bhouse-nexthop:
which split function is failing? If its ares_strsplit() or ares_buf_split_str() there's no reason to define ascii as by definition all strings in DNS are ascii and there is zero unicode support in this library at all, so adding a flag for ascii would just confuse things.  If we have a binary split function like ares_buf_split() that is doing something wrong by verifying ascii, that need to be fixed.

MichalPetryka:
> which split function is failing? If its ares_strsplit() or ares_buf_split_str() there's no reason to define ascii as by definition all strings in DNS are ascii and there is zero unicode support in this library at all, so adding a flag for ascii would just confuse things. If we have a binary split function like ares_buf_split() that is doing something wrong by verifying ascii, that need to be fixed.

It was the `ares_strsplit` that's used there.

> I should also mention the docs explicitly call out "The data is validated to be ASCII-printable" so the docs conform with the behavior.

Didn't notice that but even then, the split function itself has no such remark.

If you prefer to handle it differently though, I can close the PR and leave it up to you.

bradh352:
closing this since it isn't the right solution

PR #1022, opened 2025-08-16
Cleanup DNS server fetching and add a GetNetworkParams fallback

### LINKED ISSUE #1020: `c-ares` fails to load DNS servers for users with broken `GetAdaptersAddresses`
For some users, `GetAdaptersAddresses` (which is used by `c-ares`) and `DhcpRequestParams` seem broken for some reason and both return no DNS servers. However both the legacy GetNetworkParams and the registry keys contain proper data. 
`c-ares` should fallback to `GetNetworkParams` to handle such cases.

Split off from #1016

MichalPetryka:
Cleans up the Windows DNS server list fetching code to make it more readable and adds a `GetNetworkParams` fallback to make it handle weird user configurations.

Fixes #1020.

### DIFF
--- a/src/lib/ares_sysconfig_win.c
+++ b/src/lib/ares_sysconfig_win.c
@@ -211,11 +211,19 @@ static int compareAddresses(const void *arg1, const void *arg2)
  * to sort the DNS servers correctly.
  */
 static ULONG getBestRouteMetric(IF_LUID * const luid, /* Can't be const :( */
-                                const SOCKADDR_INET * const dest,
-                                const ULONG                 interfaceMetric)
+                                const struct sockaddr * const dest,
+                                const ULONG                   interfaceMetric)
 {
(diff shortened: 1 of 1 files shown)


MichalPetryka:
> Can you better summarize the bug in c-ares. It seemed on one of the tickets you seemed to indicate it was a parse failure and that the fallback itself wasn't actually necessary, but you're adding fallback logic still.

This one ssems to be some bug in Windows and separate from the other issue which is why I split it off into #1020.
Dumping the outputs of `GetAdaptersAddresses` and `GetNetworkParams` for some affected users shows that `GetAdaptersAddresses` returns null `FirstDnsServerAddress` on every single adapter.
`GetNetworkParams` however still works for some reason and returns the valid IPv4 servers.

I wasn't sure if the code should only call it if the list is empty or alwayss but I've settled on the latter to make the code easier to validate by not depending on this weird behaviour.

PR #1023, opened 2025-08-17
Add ares_queue_empty()

pallas:
Add a function to indicate whether the queries queue is empty.  Unlike ares_queue_active_queries(), does not walk the entire queue.

Signed-off-by: Derrick Lyndon Pallas (@pallas)

### DIFF
--- a/docs/ares_queue.3
+++ b/docs/ares_queue.3
@@ -4,13 +4,15 @@
 .\"
 .TH ARES_QUEUE 3 "16 February 2024"
 .SH NAME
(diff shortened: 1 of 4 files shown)


pallas:
I am currently using `ares_queue_active_queries` to tell if there is more work to do, but that unnecessarily walks the whole list of active queries.

dimbleby:
>  that unnecessarily walks the whole list of active queries.

Are you sure about that?

pallas:
> Are you sure about that?

Ha, my mistake. I don't know what I was reading.

PR #1024, opened 2025-08-19
Convert user config to punycode on Windows

### LINKED ISSUE #1016: `c-ares` fails to resolve for some users while getaddrinfo works

MichalPetryka:
Alternative fix for #1016 as per https://github.com/c-ares/c-ares/pull/1021#issuecomment-3193752663

Fixes #1016.


MichalPetryka:
The OS API used here prepends the Punycode with the `xn--` prefix, since we're dealing with domains here I assume that's correct though?

bradh352:
closing in favor of what has already merged and the future hand-rolled non-OS-specific punycode implementation

PR #1031, opened 2025-09-07
Punycode/IDNA support and encoding of Windows unicode search suffixes

### LINKED ISSUE #1016: `c-ares` fails to resolve for some users while getaddrinfo works

bradh352:
This PR implements native IDNA (punycode) support in a non-OS-specific manner with no external dependencies, and uses it to convert unicode search domains from system configuration (Windows registry, resolv.conf, environment, ...) to the punycode form DNS needs.

Properly fixes the search-suffix half of #1016, which was closed by the #1034 stop-gap (it dropped any non-ASCII Windows search suffix; this PR replaces that). The other half of that issue — `GetAdaptersAddresses` returning no DNS servers on some machines — is tracked separately in #1020.

## Components

**Punycode codec** (`src/lib/str/ares_punycode.c`) — RFC 3492 encode/decode operating directly on UTF-8 via `ares_buf`, derived from Ben Noordhuis's MIT implementation. Each domain label is coded separately with the `xn--` prefix; the RFC's 32-bit overflow checks are implemented on both directions; the decoder uses `ares_array` for insertion rather than raw `memmove`; encoded labels beyond the 63-octet DNS limit are rejected; blank labels are preserved so fully-qualified names with a trailing dot round-trip.

**UTF-8 primitives** (`src/lib/str/ares_buf.c`) — `ares_buf_fetch_codepoint()`, `ares_buf_append_codepoint()`, `ares_buf_len_utf8()` and friends, with strict validation: malformed/truncated sequences, overlong encodings, UTF-16 surrogate halves, and codepoints beyond U+10FFFF are rejected in both directions, so hostile `xn--` labels cannot decode into invalid unicode.

**UTS #46 mapping table** (`src/lib/str/ares_idnamap.c`, generated by `unicode-codegen.py` from Unicode 17.0.0) — codepoint ranges with disallowed/ignored/mapped status, searched with libc `bsearch()` (16 bytes/entry). Mappings are variable length (up to 18 codepoints), stored in a shared deduplicated UTF-8 pool referenced by offset/length; generation hard-errors on anything unrepresentable rather than truncating. The generated file carries SPDX + Unicode license tags and is wrapped in `clang-format off/on`.

**IDNA encode** (`ares_idna_encode_domain{,_buf}()`) — the function consumers should use: applies the UTS #46 mapping step (casefold, strip ignored codepoints, reject disallowed ones) across the whole domain (so separators like U+3002 IDEOGRAPHIC FULL STOP work), then punycode encodes each label. The raw `ares_punycode_*` API remains mapping-free, which is what the RFC 3492 test vectors exercise.

**Central config hook** (`ares_sysconfig_domains_idna()` in `ares_sysconfig.c`) — runs once in `ares_init_by_sysconfig()` after search domains are gathered from *all* platform sources (Windows registry, resolv.conf, LOCALDOMAIN, macOS, Android, ...): non-ASCII domains are IDNA encoded, and entries that cannot be converted are dropped individually so the remaining configuration still applies. Two pre-existing ASCII-only gates silently broke unicode search domains before any conversion could see them, and are fixed as part of this: the string fetch operations (`ares_buf_fetch_str_dup()`, `ares_buf_tag_fetch_string()`, `ares_buf_tag_fetch_strdup()`) now take an `ares_buf_charset_t` validation parameter (`ARES_BUF_CHARSET_ASCII` or `ARES_BUF_CHARSET_UTF8` — validated UTF-8, not blindly-accepted high bytes), with `ares_strsplit()` selecting UTF-8 via a new `ARES_BUF_SPLIT_UTF8` split flag and the resolv.conf value extraction requesting it directly. All other call sites keep ASCII-only validation, and future validation modes become new enum values rather than sibling functions.

## Design decisions

- **Nontransitional processing** only (deviation characters like ß are kept), matching modern browsers.
- **IDNA2008 NV8/XV8 exclusions are treated as disallowed** for unicode codepoints (so e.g. emoji are rejected). Since UTS #46 validates *after* mapping, entries whose mapping target is itself disallowed (e.g. U+2152 `⅒` maps through NV8-excluded U+2044) are flipped to disallowed at table-generation time, so the runtime never needs a second validation pass.
- **ASCII passes through the mapper directly** (lowercased): the NV8 exclusions would otherwise reject `_`, which is in widespread DNS use (`_ldap._tcp`, `_dmarc`, …). Hostname validity of ASCII is enforced at query-write time as today.
- **NFC normalization is not performed** (documented limitation). Real-world input is virtually always already NFC — in particular everything read from Windows configuration. If/when a general query-name hook is added, an NFC quick-check table can be generated with the same script to detect-and-reject rather than silently query a wrong label.

## Testing

- RFC 3492 §7.1 sample strings (B, C, D, E, G, I, K) round-trip through the raw punycode API.
- Malformed UTF-8 rejection matrix (truncated 2- and 4-byte sequences, bad continuation bytes, overlong, surrogates, >U+10FFFF) and invalid punycode decode (overflow, bad digits, decodes-to-surrogate).
- IDNA cases: unicode/ASCII casefolding, ignored and disallowed codepoints, ideographic full stop as label separator, nontransitional ß, multi-codepoint compatibility mappings (Ⅷ → `viii`), mapping-through-disallowed (⅒), underscore passthrough, 63-octet label limit, trailing-dot round-trip.
- Config path: `SysConfigDomainsIDNA` unit test (mixed ascii/unicode/unconvertible domain lists, plus the full `ares_sysconfig_parse_resolv_line()` → IDNA pass path, all platforms) and `ContainerIDNResolvInit` (real `ares_init()` against a resolv.conf containing unicode search domains, Linux containers).

## Future work (follow-up PRs)

- Hook IDNA encoding into user-supplied query names (`ares_getaddrinfo()` etc.), likely behind an opt-in, with the NFC quick-check above.  User-supplied search domains (`ARES_OPT_DOMAINS` / `ares_set_search()`) bypass the conversion hook by design and would be covered by the same opt-in.
- Decode-for-display (PTR/CNAME results, adig output), along with a fuzz entry point for `ares_idna_encode_domain()` / `ares_punycode_decode_domain()` before decode faces untrusted data.
- Hosts-file hostnames remain ASCII-only (a unicode hosts entry is silently skipped as a bad entry).
- The `gethostname()`-derived default search domain in `init_by_defaults()` bypasses the conversion hook (pre-existing); a non-ASCII hostname lands raw in the search list and fails at query-write time.


dfandrich:
I'd feel better with a more extensive test suite. This kind of parsing is always tricky to get right in C, especially at the edge cases. Things like maximum and over-length hostname, invalid UNICODE, characters not allowed in hostnames, a hostname that's different between IDNA2003 and IDNA2008 (e.g. ß), etc.

Why not just bypass all this by just relying on libidn2 everywhere?

renodr:
@bradh352 do you want me to look at the OpenWatcom failure at some point?

src/lib/ares_sysconfig.c
```diff
@@ -502,6 +502,43 @@ static ares_status_t
...
+      }
+
+      ares_free(domain);
+      domain = encoded;
+    }
+
+    sysconfig->domains[cnt++] = domain;
+  }
+
+  sysconfig->ndomains = cnt;
```
bradh352, line 545: **[Major]** If every domain is unconvertible (e.g. the only registry `SearchList` entry is an emoji domain -- NV8-disallowed, and exactly the malformed-config family from #1016), `cnt == 0` here but `sysconfig->domains` stays non-NULL. `ares_sysconfig_apply()` then calls `ares_strsplit_duplicate(domains, 0)` which returns NULL by design, so apply returns a spurious `ARES_ENOMEM` and skips lookups, sortlist, ndots, tries, timeout, rotate, usevc (servers survive only because they apply first). Pre-PR, `domains != NULL` implied `ndomains >= 1`, so this state is new.

```suggestion
  sysconfig->ndomains = cnt;
  if (cnt == 0) {
    ares_free(sysconfig->domains);
    sysconfig->domains = NULL;
  }
  return ARES_SUCCESS;
```

bradh352 (commented):
Deep review with mechanical verification (multi-agent assisted; every reported finding was independently re-verified against the code). Inline comments are tagged [Major]/[Medium]/[Low]/[Nit].

## Verified clean

- **RFC 3492 conformance** checked line-by-line against the spec and its Appendix C reference implementation: bootstring parameters, `adapt()`, both codec loops, and all four sec. 6.4 overflow pre-checks are exact; uppercase digits decode; last-delimiter rule and empty-label round-trip correct; `ares_array_insert_at` index semantics match "insert n at position i".
- **Generated UTS #46 table** verified against Unicode 17.0.0's official `IdnaMappingTable.txt` across **all 1,114,112 codepoints with zero mismatches** vs the documented semantics (NV8/XV8 -> disallowed, deviation kept, single-shot disallowed-flip = exactly 300 codepoints); table is sorted, non-overlapping, all pool offsets in-bounds and valid UTF-8; regeneration is byte-identical (modulo the provenance lines, see inline). The pool is emitted as an integer array, so no C89 string-literal length limits or hex-escape adjacency hazards.
- **UTF-8 primitives** are strictly conformant with Unicode Table 3-7 (all overlong/surrogate/out-of-range/truncation/bad-lead-and-continuation boundaries land on the right side, both directions); all byte handling is `unsigned char`, no signed-char promotion bug.
- **Charset retrofit**: all 22 production call sites audited -- the ASCII path is bit-for-bit the old `ares_isprint` behavior, and the only wire-format consumer (URI RR target) already required printable ASCII pre-PR, so **no DNS interoperability regression**. UTF-8 is surgically limited to search-domain values; nameservers/options/hosts/HOSTALIASES stay ASCII.
- **Integration**: the sysconfig hook is genuinely the single choke point (Windows registry, resolv.conf, LOCALDOMAIN, macOS dnsinfo, Android, QNX, libresolv -- and `ares_reinit()` reuses the same path); the #1034 ASCII gate is fully removed; registry UTF-16 -> UTF-8 via existing `WideCharToMultiByte(CP_UTF8)` is sound; no double-encode of already-punycoded labels; `ARES_OPT_DOMAINS`/`ares_set_search()` correctly remain out of scope per the description.
- **Build/ABI**: no installed header touched, no public API/ABI change; Makefile.inc flows into all three build systems; builds warning-free including under `-Wall -Wextra -Wconversion -Wsign-conversion`; all new tests pass under ASAN; changed lines are clang-format clean.

## Must fix before merge

The all-domains-unconvertible state (`domains != NULL`, `ndomains == 0`) makes `ares_sysconfig_apply()` fail with a spurious ENOMEM and skip lookups/sortlist/ndots/tries/timeout/rotate/usevc -- see the [Major] inline comment in ares_sysconfig.c. Small fix; the "all-bad" test case suggested inline doubles as its regression test.

## Decode direction (latent -- zero production callers today)

Case-sensitive `xn--` matching (RFC 5890 wants case-insensitive), the zero-basic-codepoints delimiter deviation (`xn---baa` aliases `xn--baa`), no input length cap (O(n^2) insertion), and a header-doc contradiction. If decode is intentional groundwork for the stated display-decoding follow-up, keeping it is fine -- but these should be fixed before decode ever touches wire data, ideally now while it's fresh. Recommend also adding a fuzz entry point for `ares_idna_encode_domain()`/`ares_punycode_decode_domain()` alongside the existing `ares-fuzz.c` harnesses; this is exactly the codec shape where fuzzing pays off.

## Design confirmations (deliberate per the PR description; suggest one sentence in user-facing docs)

- **NV8/XV8-as-disallowed** rejects some names browsers resolve (registered emoji domains like `xn--e28h.ws`, U+2260, vulgar fractions) -- strict IDNA2008/libidn2 posture rather than WHATWG. Defensible for a DNS library, but it makes some resolvable names unresolvable through c-ares, so it deserves a documented sentence.
- **NFC omission** is documented; CheckJoiners/CheckBidi/CheckHyphens omissions should be too (see inline).

## Known edge behaviors, inherited or accepted

- `LOCALDOMAIN` containing one invalid-UTF-8 byte aborts the entire env config as ENOMEM (the split is all-or-nothing) -- pre-existing shape for non-ASCII, but the central hook was arguably the chance to make this drop-individually too.
- Android JNI `GetStringUTFChars()` returns modified UTF-8 (CESU-8), so a non-BMP search suffix would fail validation and drop all Android search domains before the hook. Practically rare (IDN suffixes are almost always BMP) and pre-PR all non-ASCII was dropped anyway -- still a net improvement.
- The 255-octet total name limit stays deferred to query-write time, consistent with existing behavior for overlong ASCII domains.

## Prior feedback follow-through

@dfandrich's asks are now substantially covered: nontransitional eszett, the 63-octet post-encode limit, and a solid malformed-UTF-8/invalid-punycode rejection matrix all have vectors; the 255-octet full-name case is the one still untested (unenforced at this layer by design). Staying dependency-free instead of libidn2 looks right given the table now verifies mechanically against upstream Unicode data.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,708 @@
...
+    return ARES_SUCCESS;
+  }
+
+  /* Make sure input is all ascii-printable or its an error */
+  if (!ares_buf_isprint(inbuf)) {
+    return ARES_EFORMERR;
+  }
+
+  /* If it doesn't start with "xn--" then its all ascii */
+  if (!ares_buf_begins_with(inbuf, (const unsigned char *)"xn--", 4)) {
```
bradh352: **[Medium]** `ares_buf_begins_with()` is a raw `memcmp`, but RFC 5890 sec. 2.3.2.1 defines the ACE prefix as case-insensitive; `XN--MNCHEN-3YA` silently passes through un-decoded (empirically confirmed against the built library). Latent -- decode has zero production callers -- but this API's purpose is decoding third-party data, so it should be fixed before decode is ever wire-exposed. Compare the first 4 bytes case-insensitively before consuming.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,708 @@
...
+      unsigned int *cp = NULL;
+
+      status = ares_array_insert_last((void **)&cp, codepoints);
+      if (status != ARES_SUCCESS) {
+        goto done;
+      }
+      *cp = data[j];
+    }
+    /* Consume '-' */
+    ares_buf_consume(inbuf, 1);
```
bradh352, line 443: **[Medium]** RFC 3492 sec. 6.2 deviation: "if **more than zero** code points were consumed then consume one more (which will be the last delimiter)". The reference implementation (`in = b > 0 ? b + 1 : 0`) consequently rejects an extended part starting with `-`. Here `num_ascii_chars == 0 != SIZE_MAX` still consumes the `-`, so `xn---baa` decodes identically to `xn--baa` -- two distinct A-labels aliasing one U-label, a canonicalization ambiguity if decode ever sees untrusted data. Only consume the delimiter when `num_ascii_chars > 0`; a leading `-` then fails naturally in `decode_digit()`.

src/lib/str/ares_punycode.h
```diff
@@ -0,0 +1,116 @@
...
+ *  \param[out] out     Output punycode encoded domain
+ *  \return ARES_SUCCESS on success, or otherwise an ares_status_t error.
+ */
+ares_status_t ares_punycode_encode_domain(const char *domain, char **out);
+
+/*! Punycode decode a domain as per RFC3492 from buf to buf.
+ *
+ *  This function will split the domain into each component then punycode decode
+ *  it and rejoin the components.  If the passed in domain doesn't have any
+ *  sections starting with "xn--" this will return the same as the input, but
```
bradh352: **[Medium]** Doc contradicts behavior: any label containing a byte outside printable ASCII fails `ares_buf_isprint()` and returns `ARES_EFORMERR`, so `xn--bcher-kva.münchen.example` errors rather than "return the same as the input". Either pass non-ASCII labels through unchanged, or document the printable-ASCII precondition and the error code.

src/lib/str/ares_buf.c
```diff
@@ -787,6 +1005,40 @@ size_t ares_buf_consume_until_charset(ares_buf_t          *buf,
...
+  ares_ssize_t         pos;
+  ares_bool_t          found = ARES_FALSE;
+
+  if (ptr == NULL || charset == NULL || len == 0) {
+    return 0;
+  }
+
+  for (pos = (ares_ssize_t)remaining_len - 1; pos >= 0; pos--) {
+    size_t j;
+    for (j = 0; j < len; j++) {
```
bradh352, line 1013: **[Medium]** This violates the documented `require_charset == ARES_FALSE` contract. The header promises "if ARES_FALSE it will simply consume the rest of the buffer", but when the byte isn't found the loop exits with `pos == -1`, the consume is skipped, and this returns `(size_t)-1` == SIZE_MAX -- the not-found sentinel reserved for the TRUE case. Latent (the only caller passes ARES_TRUE, and no test executes the FALSE branch), but it's a booby trap in a general-purpose primitive.

Related inconsistency: an empty buffer returns 0 even when `require_charset == ARES_TRUE` (reachable via a bare `"xn--"` label), where SIZE_MAX would be consistent with the non-empty not-found case.

test/ares-test-internal.cc
```diff
@@ -508,6 +509,344 @@ TEST_F(LibraryTest, URI) {
...
+      "plain ascii" },
+    /* ASCII is lowercased for canonical form */
+    { ARES_SUCCESS,  "WWW.EXAMPLE.COM",   "www.example.com",
+      "ascii lowercased" },
+    /* Unicode casefolding is applied before punycode (unlike the raw
+     * punycode API which preserves case) */
+    { ARES_SUCCESS,  "www.München.com",   "www.xn--mnchen-3ya.com",
+      "unicode casefold" },
+    /* Ignored codepoints are removed: U+00AD SOFT HYPHEN */
+    { ARES_SUCCESS,  "exam­ple.com", "example.com",
```
bradh352: **[Medium]** This literal contains a raw soft hyphen (U+00AD) -- invisible, so if an editor/formatter ever strips it the test degenerates to `"example.com" == "example.com"` and passes vacuously. Should be `"exam\xC2\xADple.com"` regardless of anything else.

More broadly: 9533ff97 escaped U+1F600 for MSVC, but ~30 other narrow literals in these tests still carry raw UTF-8 bytes and no `/utf-8` flag exists in any build file. On cp1252 runners bytes pass through 1:1, but MSVC on DBCS codepages (cp932/936/949) will mangle the CJK/Hangul/Hebrew literals -- compiles fine, fails at runtime. Either add `/utf-8` for MSVC (which also makes the 9533ff97 workaround unnecessary) or escape them all.

test/ares-test-internal.cc
```diff
@@ -508,6 +509,344 @@ TEST_F(LibraryTest, URI) {
   EXPECT_NE(ARES_SUCCESS, ares_uri_parse_buf(NULL, NULL));
   EXPECT_NE(ARES_SUCCESS, ares_uri_parse_buf(NULL, NULL));
 }
+
+TEST_F(LibraryTest, PUNYCODE) {
```
bradh352, line 514: **[Medium]** Codec test gaps (bundled):

1. No vector ever executes the 4-byte UTF-8 emit path -- every success vector is BMP-only. Suggested row (cross-checked against Python's punycode codec), U+10348 GOTHIC LETTER HWAIR:
```c
{ ARES_TRUE, "\xF0\x90\x8D\x88.com", "xn--2c8c.com" },
```
2. No uppercase-prefix decode vector (`XN--MNCHEN-3YA.com`) -- would have caught the case-sensitive ACE prefix bug.
3. Of the omitted RFC 3492 sec. 7.1 vectors, (L) `3年B組金八先生` is the only shape with digits inside the basic-codepoint segment -- worth adding.
4. The `success` field is dead (no `ARES_FALSE` row exercises it) and `EXPECT_STREQ(encoded, tests[i].output)` below has actual/expected reversed vs gtest convention.

test/ares-test-internal.cc
```diff
@@ -508,6 +509,344 @@ TEST_F(LibraryTest, URI) {
...
+      EXPECT_STREQ(tests[i].output, encoded) << tests[i].descr;
+    }
+    ares_free(encoded);
+  }
+
+  EXPECT_NE(ARES_SUCCESS, ares_idna_encode_domain(NULL, NULL));
+  EXPECT_NE(ARES_SUCCESS, ares_idna_encode_domain("www.bücher.com", NULL));
+}
+
+TEST_F(LibraryTest, SysConfigDomainsIDNA) {
```
bradh352, line 872: **[Medium]** Drop semantics are only tested for bad-in-middle. Add bad-first / bad-last / **all-bad** cases -- all-bad currently trips the `ndomains == 0` / non-NULL `domains` bug flagged in ares_sysconfig.c, so it doubles as the regression test.

Separately, both test files use `SetAllocFail` heavily but none of `ares_idna_encode_domain()` / `ares_sysconfig_domains_idna()` runs under it, despite ~1500 lines of allocation-heavy new code; that's the repo's standard leak/crash net (ASAN/LSAN CI then proves no leak at any failure point). E.g.:
```cpp
TEST_F(LibraryTest, IDNAAllocFail) {
  for (int ii = 1; ii <= 30; ii++) {
    char *out = NULL;
    ClearFails();
    SetAllocFail(ii);
    ares_status_t st = ares_idna_encode_domain("b\xC3\xBC" "cher.M\xC3\xBC" "nchen.example", &out);
    if (st == ARES_SUCCESS)
      EXPECT_STREQ("xn--bcher-kva.xn--mnchen-3ya.example", out);
    else
      EXPECT_EQ(ARES_ENOMEM, st);
    ares_free(out);
  }
}
```

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,708 @@
...
+    status = ARES_ENOMEM;
+    goto done;
+  }
+
+  status = ares_punycode_encode_domain_buf(inbuf, outbuf);
+  if (status != ARES_SUCCESS) {
+    goto done;
+  }
+
+  *out   = ares_buf_finish_str(outbuf, NULL);
```
bradh352: **[Low]** `ares_buf_finish_str()` NULL result unchecked in all three string wrappers (also the decode and idna variants). OOM yields `ARES_SUCCESS` with `*out == NULL` plus a leaked buf (`ares_buf_finish_bin()` doesn't free on failure), and `ares_sysconfig_domains_idna()` would store a NULL domain. OOM-only, practically unhittable; three-line fix per site matching codebase convention (e.g. `record/ares_dns_name.c`):
```c
  *out = ares_buf_finish_str(outbuf, NULL);
  if (*out == NULL) {
    status = ARES_ENOMEM;
    goto done;
  }
  outbuf = NULL;
```
Related: `ares_punycode_encode_domain("")` returns `ARES_ENOMEM` because `ares_buf_create_const(ptr, 0)` returns NULL by design -- wrong error code for empty input; consider special-casing `ares_strlen(domain) == 0`.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,708 @@
...
+
+    ares_buf_tag(inbuf);
+    while (ares_buf_len(inbuf)) {
+      status = ares_buf_fetch_codepoint(inbuf, &cp);
+      if (status != ARES_SUCCESS) {
+        return status;
+      }
+      if (cp < n) {
+        delta++;
+        if (delta > PUNY_MAXINT) {
```
bradh352: **[Low]** (practically impossible to hit -- flagged for RFC sec. 6.3 fidelity only) On platforms where `SIZE_MAX == 0xFFFFFFFF`, `delta++` wraps to 0 before `delta > PUNY_MAXINT` can ever be true, and the `delta++` in the outer `for (; h < utf8_cnt; n++, delta++)` is unguarded on all platforms. Reaching it requires >= ~3855 already-emitted codepoints, so the 63-octet label check rejects the result regardless; worst case is garbage bytes in a discarded buffer, no memory unsafety. Cheap fix: `if (delta == PUNY_MAXINT) { return ARES_EBADNAME; } delta++;` (and same for the outer increment).

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,708 @@
...
+    ares_buf_consume(inbuf, 1);
+  }
+
+  ares_buf_tag_clear(inbuf);
+
+  i    = 0;
+  n    = INITIAL_N;
+  bias = INITIAL_BIAS;
+
+  while (ares_buf_len(inbuf) > 0) {
```
bradh352, line 452: **[Low]** Encode enforces the 63-octet label limit but decode doesn't cap input length, and `ares_array_insert_at()` memmoves per codepoint, so a multi-kilobyte "label" gives O(n^2) work and unbounded allocation. No untrusted callers today; add `ares_buf_len(inbuf) > MAX_DNS_LABEL_LEN -> ARES_EBADNAME` before the decode loop before this is ever pointed at wire data.

src/lib/str/ares_buf.c
```diff
@@ -447,12 +518,27 @@ ares_status_t ares_buf_tag_fetch_constbuf(const ares_buf_t *buf,
   return ARES_SUCCESS;
 }
 
+ares_status_t ares_buf_tag_fetch_buf(const ares_buf_t *buf, ares_buf_t *outbuf)
```
bradh352: **[Low]** `ares_buf_tag_fetch_buf()` has zero callers anywhere in src/ or test/, and it's `CARES_EXTERN` so it adds an exported symbol with no users. Use it, drop it, or at least drop the export.

src/lib/str/ares_buf.c
```diff
@@ -293,6 +293,77 @@ ares_status_t ares_buf_append_be32(ares_buf_t *buf, unsigned int u32)
   return ARES_SUCCESS;
 }
 
+ares_status_t ares_buf_append_codepoint(ares_buf_t *buf, unsigned int codepoint)
+{
+  ares_status_t status;
+
+  /* UTF-16 surrogate halves are not valid Unicode scalar values */
+  if (codepoint >= 0xD800 && codepoint <= 0xDFFF) {
```
bradh352, line 302: **[Low]** Error-code asymmetry: the append direction rejects surrogates and > U+10FFFF with `ARES_EFORMERR`, while the fetch/decode direction uses `ARES_EBADSTR` for the same invalid scalar values. Callers checking `ARES_EBADSTR` for "bad unicode" will mis-handle encode failures. Suggest `ARES_EBADSTR` here too, and document in ares_buf.h that surrogates/out-of-range are rejected -- ares_punycode.c explicitly relies on that behavior. (Same theme in the codec itself: encode failures are `ARES_EBADNAME` while equally-malformed decode input is `ARES_EFORMERR`; worth unifying on EBADNAME with EFORMERR reserved for NULL-parameter misuse.)

src/lib/str/ares_punycode.h
```diff
@@ -0,0 +1,116 @@
...
+ *  \return ARES_SUCCESS on success, or otherwise an ares_status_t error.
+ */
+ares_status_t ares_punycode_decode_domain(const char *domain, char **out);
+
+/*! IDNA encode a domain from buf to buf: apply the UTS #46 mapping step
+ *  (nontransitional, casefolding, removing ignored codepoints, and rejecting
+ *  disallowed codepoints), then punycode encode each label as per RFC3492.
+ *
+ *  This is the function to use for converting a user-visible unicode domain
+ *  into the ASCII form used on the wire.  NFC normalization of the mapped
```
bradh352: **[Low]** The NFC omission is documented here, but CheckJoiners (ContextJ), CheckBidi, and CheckHyphens from UTS #46 sec. 4 are also not implemented -- ZWJ/ZWNJ pass through context-free as deviation-valid. One sentence here prevents future "is this intentional?" churn.

Also on this header: `\param[in] inbuf` on both `_buf` entry points should be `[in,out]` (the split consumes it), on error `outbuf` may hold a partial label (undocumented), and the six declarations lack `CARES_EXTERN` even though the test suite calls them -- a `CARES_SYMBOL_HIDING=ON` shared-only build would fail to link the tests (precedent is mixed: ares_buf.h has it, ares_strsplit.h doesn't).

src/lib/str/unicode-codegen.py
```diff
@@ -0,0 +1,303 @@
...
+    for entry in codepoints:
+        if entry["code_min"] <= cp <= entry["code_max"]:
+            return entry["status"]
+    return 0  # valid
+
+for entry in codepoints:
+    if entry["status"] != 3:
+        continue
+    for cp in entry["mapping"]:
+        if cp >= 0x80 and find_status(cp) == 1:
```
bradh352: **[Low]** Hardening the hard-error philosophy (bundled):

1. No validation that parsed ranges are sorted, non-overlapping, and `code_min <= code_max` -- overlapping input would silently survive sort+merge and corrupt the bsearch semantics. Suggest asserting `codepoints[i]["code_min"] > codepoints[i-1]["code_max"]` after the sort.
2. The mapped->disallowed flip is a single pass over a list it mutates; if a mapping target were itself a mapped codepoint flipped later in the pass, the result would be order-dependent. Impossible for Unicode 17 (verified mechanically -- no mapping target is itself mapped), but an assert or fixpoint iteration future-proofs it.
3. The status magic numbers 1/2/3 duplicate `ares_idnamap_status_t` in ares_idnamap.h with no cross-reference in either direction -- add "must match" comments in both places.
4. The output header embeds `sys.argv[0]` and the raw URL argument verbatim, so regenerating from another directory produces spurious provenance diffs; use `os.path.basename(sys.argv[0])`.
5. Nits: `casefold` is a misleading leftover name for the whole file contents; one f-string has no placeholders.

src/lib/str/ares_idnamap.c
```diff
@@ -0,0 +1,7740 @@
+/* clang-format off */
```
bradh352, line 1: **[Low]** REUSE polish: the tag is `MIT` only, with the Unicode license as prose. `reuse lint` passes and the prose satisfies the Unicode notice requirement, but the strict form for derived Unicode data would be `SPDX-License-Identifier: MIT AND Unicode-3.0` plus `LICENSES/Unicode-3.0.txt` (Unicode-3.0 is a registered SPDX id).

test/ares-test-internal.cc
```diff
@@ -508,6 +509,344 @@ TEST_F(LibraryTest, URI) {
...
+    EXPECT_STREQ("xn--bcher-kva.de", sysconfig.domains[0]);
+    EXPECT_STREQ("first.com", sysconfig.domains[1]);
+
+    ares_strsplit_free(sysconfig.domains, sysconfig.ndomains);
+    ares_buf_destroy(buf);
+    ares_destroy(channel);
+  }
+}
+
+TEST_F(LibraryTest, BufCharset) {
```
bradh352, line 977: **[Low]** The new buf primitives (`append_codepoint` / `fetch_codepoint` / `len_utf8` / `isprint` / `tag_fetch_buf` / `consume_last_charset`) are tested only through punycode -- no direct boundary tests (U+007F/0080, U+07FF/0800, U+FFFF/10000, U+10FFFF on the append side), and `BufMisuse` NULL-arg rows weren't added for them, breaking that test's completeness convention. `consume_last_charset(..., ARES_FALSE)` is executed by nothing, test or production.

`ARES_BUF_SPLIT_UTF8` has no direct `ares_buf_split` test, and its all-or-nothing failure on one invalid-UTF-8 token is unpinned -- worth a test documenting that contrast with the IDNA-stage drop-individually semantics.

Also worth pinning in the IDNA table (all verified working, none asserted): already-punycoded input `xn--bcher-kva.de` passes through un-double-encoded, fullwidth uppercase maps to ASCII lowercase, ZWNJ nontransitional handling. And this test collapses status to a success boolean rather than pinning ARES_EBADSTR vs ARES_EBADRESP.

src/lib/include/ares_buf.h
```diff
@@ -338,6 +357,24 @@ CARES_EXTERN ares_status_t ares_buf_tag_fetch_strdup(const ares_buf_t *buf,
 CARES_EXTERN ares_status_t ares_buf_tag_fetch_constbuf(const ares_buf_t *buf,
                                                        ares_buf_t **newbuf);
 
+/*! Fetch the bytes starting from the tagged position up to the _current_
+ *  position and write into the provided buffer.
+ *
+ *  \param[in]     buf    Initialized buffer object
+ *  \param[out]    outbuf Buffer to write bytes into.
+
```
bradh352: **[Nit]** Bare blank line inside the doxygen block (missing ` *` prefix), inconsistent with the rest of the header; and `\return` documents only success, not `ARES_EFORMERR` (no tag / NULL outbuf) or `ARES_ENOMEM`.

src/lib/include/ares_buf.h
```diff
@@ -365,6 +402,18 @@ CARES_EXTERN ares_status_t ares_buf_fetch_be32(ares_buf_t   *buf,
                                                unsigned int *u32);
 
 
+/*! Fetch a UTF 32bit codepoint from a UTF-8 encoded string.  Will consume
```
bradh352: **[Nit]** "Will consume up to the length of the codepoint" is ambiguous; the actual (good) semantics are: consumes exactly the sequence length on success, consumes nothing on error -- worth stating since it differs from lenient decoders. `ARES_EBADRESP` (NULL/empty buffer) is returned but undocumented. Nearby: typo at the `ares_buf_consume_last_charset` doc ("end of the of the buffer"), and the `ARES_BUF_CHARSET_UTF8` doc should note that non-ASCII codepoints are not printability-checked (C1 controls, zero-width, and bidi-control codepoints pass).

src/lib/record/ares_dns_parse.c
```diff
@@ -1071,7 +1071,8 @@ static ares_status_t ares_dns_parse_rr_uri(ares_buf_t *buf, ares_dns_rr_t *rr,
   }
 
   /* NOTE: Not in DNS string format */
-  status = ares_buf_fetch_str_dup(buf, remaining_len, &name);
+  status =
+    ares_buf_fetch_str_dup(buf, remaining_len, &name, ARES_BUF_CHARSET_ASCII);
```
bradh352, line 1076: **[Nit]** The `ares_str_isprint()` re-check a few lines below is now doubly redundant -- the ASCII-charset fetch already guarantees printable ASCII -- and can be removed.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,708 @@
...
+    if (i / (cnt + 1) > PUNY_MAXINT - n) {
+      /* OVERFLOW */
+      status = ARES_EFORMERR;
+      goto done;
+    }
+
+    n += i / (cnt + 1);
+    i %= (cnt + 1);
+
+    status = ares_array_insert_at((void **)&cp, codepoints, i);
```
bradh352, line 522: **[Nit]** The decoder's core loop lacks the RFC-section breadcrumbs the encoder has (line 199 cites sec. 6.3). A `/* RFC 3492 6.2: insert n into output at position i */` here and a one-liner at the var-int loop would help future readers.

test/ares-test-init.cc
```diff
@@ -451,6 +451,27 @@ CONTAINED_TEST_F(LibraryTest, ContainerFullResolvInit,
...
 
+// Unicode (IDN) search domains from resolv.conf are converted to their
+// punycode form; entries that cannot be converted are dropped.
+NameContentList idnresolv = {
+  {"/etc/resolv.conf", "nameserver 1.2.3.4\n"
+                       "search bücher.de first.com \xF0\x9F\x98\x80.com"
+                       " münchen.example\n"}};
+CONTAINED_TEST_F(LibraryTest, ContainerIDNResolvInit,
+                 "myhostname", "mydomainname.org", idnresolv) {
+  ares_channel_t *channel = nullptr;
```
bradh352, line 462: **[Nit]** `channel` is dereferenced after only `EXPECT_EQ` on init, and `domains[2]` is indexed guarded only by `EXPECT_EQ` on `ndomains` -- a regression crashes the container child instead of failing cleanly. Neighboring container tests share the pattern (and `ASSERT_*` won't compile in the non-void body), so awareness only; the idiomatic guard is `if (channel == nullptr || channel->ndomains != 3) return true;` after the EXPECTs.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,790 @@
...
+    unsigned int *cp = NULL;
+
+    /* Decode a generalized variable-length integer into delta as per
+     * RFC 3492 Section 6.2, with the Section 6.4 overflow checks */
+    for (w = 1, k = BASE;; k += BASE) {
+      unsigned char b;
+      size_t        digit;
+      size_t        t;
+
+      status = ares_buf_fetch_bytes(inbuf, &b, 1);
```
bradh352, line 466: **[Low]** Truncated punycode var-int returns `ARES_EBADRESP`, docs promise `ARES_EBADNAME`. When an ACE label ends mid variable-length-integer, `ares_buf_fetch_bytes()` fails with `ARES_EBADRESP` and that status propagates verbatim: `ares_punycode_decode_domain("xn--9", &out)` returns `ARES_EBADRESP` (empirically confirmed against the built library) -- `'9'` decodes to digit 35 >= t, the loop continues, and the next fetch hits the empty buffer. The header promises `ARES_EBADNAME` when a label "is not valid punycode", and 872873f5's stated goal was "EBADNAME for malformed punycode"; `EBADRESP` ("malformed DNS response") is a nonsensical code for this API, which matters once decode sees wire data. `invalid_puny[]` has no truncation vector and only asserts `NE ARES_SUCCESS`, so nothing pins this.

```c
      status = ares_buf_fetch_bytes(inbuf, &b, 1);
      if (status != ARES_SUCCESS) {
        /* Truncated variable-length integer is malformed punycode */
        status = ARES_EBADNAME;
        goto done;
      }
```
plus an `"xn--9"` row in `invalid_puny[]`, ideally pinning the code.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,790 @@
...
+      if (status != ARES_SUCCESS) {
+        return status;
+      }
+      continue;
+    }
+
+    switch (entry->status) {
+      case ARES_IDNA_STATUS_DISALLOWED:
+        return ARES_EBADNAME;
+      case ARES_IDNA_STATUS_IGNORED:
```
bradh352: **[Low]** A label consisting solely of IGNORED codepoints silently collapses to an empty label. The mapping step strips IGNORED codepoints with no notion of label boundaries, so a label that is entirely ignored characters becomes empty and the whole conversion *succeeds*: `ares_idna_encode_domain("a.\xC2\xAD.b")` (middle label = one soft hyphen) returns `ARES_SUCCESS` with `"a..b"`, and a lone soft hyphen returns `ARES_SUCCESS` with `""` (both empirically confirmed). `ares_sysconfig_domains_idna()` then *keeps* an entry that can never form a valid query name -- the opposite of its drop-unusable-entries purpose. This is the other half of UTS #46 ToASCII VerifyDnsLength: the 63-octet maximum is enforced but the 1-octet minimum is not. The blank-label preservation comment covers labels already empty in the *input*, not labels the mapper newly empties.

Input is practically impossible (a search suffix label made entirely of invisible characters), so documenting the divergence is acceptable. If fixing: track whether any codepoint was emitted since the last emitted `'.'`/start, and return `ARES_EBADNAME` when a label that consumed >=1 input codepoint emits none -- preserving the deliberate pass-through of input `".."` and trailing dots.

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,790 @@
...
+        break;
+      case ARES_IDNA_STATUS_MAPPED:
+        /* The generator guarantees pool references are in bounds (also
+         * verified by the IDNAMapTable unit test); guard anyhow so a corrupt
+         * table can never cause an out-of-bounds read */
+        if ((size_t)entry->map_offset + entry->map_len >
```
bradh352: **[Nit]** `(size_t)entry->map_offset + entry->map_len` can wrap on ILP32 when `map_offset` is within 255 of `SIZE_MAX` -- the sum wraps small, the check passes, and the `pool` read is OOB. Only reachable with a corrupt table, but a corrupt table is this check's sole reason to exist; the legitimate table (max offset 12466) can never trip either form. Flagged per the tiny-hardening rule since the overflow-proof form is free:

```c
        if (entry->map_len > ares_idnamap_data_pool_len ||
            (size_t)entry->map_offset >
              ares_idnamap_data_pool_len - entry->map_len) {
```

src/lib/str/ares_punycode.c
```diff
@@ -0,0 +1,790 @@
...
+    }
+
+    /* An encoded label that exceeds the DNS label limit can never be used */
+    if (ares_buf_len(outbuf) - label_start > MAX_DNS_LABEL_LEN) {
+      status = ARES_EBADNAME;
+      goto fail;
+    }
+  }
+
+fail:
```
bradh352: **[Nit]** Both `_domain_buf` functions fall through into `fail:` on success -- behavior correct, name misleading. The rest of this file and the codebase use `done:`; same at line 576.

src/lib/include/ares_punycode.h
```diff
@@ -0,0 +1,126 @@
...
+ *  really necessary before calling this function.
+ *
+ *  \param[in,out] inbuf  Input domain name.  Consumed by this function.
+ *  \param[in,out] outbuf Output punycode encoded domain.  On error may
+ *                        contain a partially-encoded name.
+ *  \return ARES_SUCCESS on success, ARES_EBADNAME if an encoded label
+ *          exceeds the DNS label length limit or the delta arithmetic would
+ *          overflow, or otherwise an ares_status_t error.
+ */
+CARES_EXTERN ares_status_t ares_punycode_encode_domain_buf(ares_buf_t *inbuf,
```
bradh352, line 47: **[Low]** `CARES_EXTERN` here conflicts with the AGENTS.md visibility rule ("`CARES_EXTERN` only in public headers and the semi-public `src/lib/include/` headers") -- this is now the only header outside `src/lib/include/` using it. It was added in response to the prior review, but it isn't actually needed for its stated purpose: every test calling these functions sits inside the existing `#ifndef CARES_SYMBOL_HIDING` block, so symbol-hiding builds compile them out regardless. As written it adds 6 exported symbols with no public header -- de facto ABI. Pick a side deliberately: either (a) move ares_punycode.h to `src/lib/include/` (matching ares_buf.h; update Makefile.inc HHEADERS and the includes), making the semi-public status explicit, or (b) drop the six `CARES_EXTERN`s, matching how ares_strsplit and `ares_sysconfig_domains_idna` are handled.

src/lib/str/ares_punycode.h
```diff
@@ -0,0 +1,126 @@
...
+ */
+#ifndef __ARES_PUNYCODE_H
+#define __ARES_PUNYCODE_H
+
+
+/*! Punycode encode a domain as per RFC3492 from buf to buf.
+ *
+ *  This function will split the domain into each component then punycode encode
+ *  it and rejoin the components.  If there are no UTF8 codepoints outside the
+ *  ascii range this will return the same as the input, but it is, however, an
```
bradh352: **[Nit]** "will return the same as the input" is overstated: the 63-octet cap applies to pass-through ASCII labels too, so an all-ASCII domain with a >63-octet label returns `ARES_EBADNAME`, not the input. Suggest "...this will return the same as the input (provided every label is within the 63-octet DNS limit)..."; same wording at line 64 on the decode side.

src/lib/include/ares_punycode.h
```diff
@@ -0,0 +1,126 @@
...
+ */
+CARES_EXTERN ares_status_t ares_punycode_decode_domain(const char *domain,
+                                                       char      **out);
+
+/*! IDNA encode a domain from buf to buf: apply the UTS #46 mapping step
+ *  (nontransitional, casefolding, removing ignored codepoints, and rejecting
+ *  disallowed codepoints), then punycode encode each label as per RFC3492.
+ *
+ *  This is the function to use for converting a user-visible unicode domain
+ *  into the ASCII form used on the wire.  Known divergences from a full
```
bradh352, line 96: **[Nit]** The three documented divergences (no NFC, no CheckHyphens/CheckBidi/CheckJoiners, NV8-strict) are accurate, but the fourth lives only in a .c comment: ASCII never consults the table, so table-disallowed ASCII (`_`, `~`, space, controls) passes through lowercased -- effectively UseSTD3ASCIIRules=false plus genuinely-disallowed ASCII accepted. A caller reading this list will assume table semantics apply to ASCII. One more bullet: "ASCII bypasses the mapping table entirely (lowercased only); hostname validity of ASCII is enforced at query-write time."

src/lib/str/ares_idnamap.h
```diff
@@ -0,0 +1,62 @@
...
+#define __ARES_IDNAMAP_H
+
+#include <stddef.h>
+
+/*! UTS #46 IDNA mapping status.  Codepoints not present in the table are
+ *  valid and used as-is (this includes deviation codepoints since we only
+ *  perform nontransitional processing).  Numeric values must match the
+ *  status values emitted by unicode-codegen.py. */
+typedef enum {
+  ARES_IDNA_STATUS_DISALLOWED = 1, /*!< Codepoint not permitted in a domain */
```
bradh352: **[Nit]** Type is `ares_idnamap_status_t` but constants are `ARES_IDNA_STATUS_*`; repo convention mirrors the type prefix (`ares_buf_charset_t` -> `ARES_BUF_CHARSET_*` in this same PR). Rename constants to `ARES_IDNAMAP_STATUS_*` (or the type to `ares_idna_status_t`) -- three uses in ares_punycode.c plus the test; no codegen change since the script emits numeric values.

bradh352 (commented):
Follow-up deep review of the head after the three response commits (multi-agent assisted; every reported finding independently re-verified, several empirically against the built library). Inline comments tagged [Low]/[Nit] -- no Major or Medium findings this round.

## Prior review resolution

**20 of 22 inline items fully fixed and verified correct** -- including the [Major] all-dropped sysconfig state -- with no new bugs introduced by the fixes. Residuals: the container-test guard ordering (inline), and the remaining `SetAllocFail` sliver for `ares_sysconfig_domains_idna()` which is **explicitly waived** (malloc-failure coverage there is not required; the ENOMEM cleanup path was hand-verified correct). The prior summary's fuzz-harness suggestion was not acted on; optional -- either add a `LLVMFuzzerTestOneInput` for `ares_idna_encode_domain()`/`ares_punycode_decode_domain()` or treat it as deferred to the display-decoding follow-up where decode first faces untrusted data.

## Items without inline anchors (files untouched by this PR)

- **[Low]** `init_by_defaults()` (ares_init.c:217): the `gethostname()`-derived default search domain is the one populate-search-domains path that bypasses the central IDNA hook -- a non-ASCII hostname lands raw in the search list, and `ares_is_hostnamech()` then makes every query using it fail `ARES_EBADNAME`. Pre-existing behavior, but the new all-dropped state (`domains = NULL`, `ndomains == 0`) makes the fallback slightly *more* reachable, and Windows unicode hostnames are the likeliest real-world source of non-ASCII here. Fine to defer -- worth a line in the future-work list.
- **[Low]** `test/Makefile.msvc` didn't get the `/utf-8` that test/CMakeLists.txt did; on cp932/936/949 hosts the nmake test build compiles the UTF-8-literal tests but fails at runtime -- append `/utf-8` to `CC_CFLAGS`. (MinGW/m32 and autotools need nothing: narrow literals pass through as UTF-8 bytes there.)
- **[Low]** "Fixes #1016" is stale: #1016 was already closed by the #1034 stop-gap merge, and its other scenario (GetAdaptersAddresses returning no servers) lives on as #1020 / open PR #1022. Suggest: "Properly fixes the search-suffix half of #1016 (closed by the #1034 stop-gap, which dropped non-ASCII Windows search suffixes); the remaining half is tracked in #1020."
- **[Nit]** Future-work list: add hosts-file hostnames (still ASCII-only; a unicode hosts entry is silently skipped) and `ARES_OPT_DOMAINS`/`ares_set_search()` (bypass the hook by design) so the conversion boundary is explicit.
- **[Nit]** If not squash-merged: the last three commits' `Signed-off-by: Brad House (@bradh352)` lacks the DCO `Name <email>` form; ensure the squash message carries the proper trailer.
- **Coordination**: open PR #1022 rewrites the server-fetching half of ares_sysconfig_win.c; whichever lands second rebases. Complementary, not rival -- coordination is already on record across #1021/#1024/#1034.
- **Awareness, no action needed**: `ARES_BUF_SPLIT_UTF8` is exercised only through `ares_strsplit()` (its all-or-nothing semantics are pinned there); domains containing any non-ASCII get their ASCII labels lowercased by the mapper while pure-ASCII domains keep case (DNS is case-insensitive, harmless).

## Verified clean (highlights)

- **RFC 3492** re-verified line-by-line including the new 872873f5 code: case-insensitive ACE prefix, delimiter-only-when-basic>0 vs the reference implementation (`xn---baa` correctly rejected), both delta guards checked pre-increment (correct where `size_t` is exactly 32 bits), 63-octet cap both directions, manual digit-by-digit traces of two encode and two decode vectors match, non-ACE labels pass through per the updated docs.
- **Generated table**: regenerates byte-identical from official Unicode 17.0.0 data (modulo provenance URL line); an independent from-scratch reimplementation of nontransitional UTS #46 semantics compared status **and exact mapping bytes** across all 1,114,112 codepoints with zero mismatches (NV8/XV8 flips, 300 fixpoint-flipped entries, U+FDFA's 33-byte mapping, deviation chars kept). Structural test validates what it claims; pool bounds guard is correct (modulo the ILP32 wrap nit inline).
- **UTF-8 primitives**: strictly conformant with Unicode Table 3-7 on every boundary, both directions; no OOB reads (every remaining-length check precedes the bytes it covers); no signed-char promotion; exact consume-on-success/nothing-on-error semantics as documented.
- **Charset retrofit**: all 22 call sites audited -- ASCII path bit-identical to pre-PR; UTF-8 surgically limited to resolv.conf values and `ares_strsplit()`; the only wire-format consumer (URI RR) unchanged in behavior and error code -- **no interoperability regression**; strictness was only ever loosened, never tightened.
- **Integration**: every writer of `sysconfig->domains` on every platform (registry, resolv.conf `search`+`domain`, LOCALDOMAIN, macOS dnsinfo, Android, QNX confstr, libresolv dnsrch) verified to run before the single IDNA hook, `ares_reinit()` included; #1034 fully superseded (ASCII gate gone, its registry-UTF-16 and macro relocations correctly retained); registry surrogate garbage becomes U+FFFD which is IDNA-disallowed and individually dropped -- the desired semantic.
- **Build/ABI/style**: no installed header or exported-ABI change (modulo the CARES_EXTERN placement question inline); Makefile.inc is the single source-enumeration point feeding all seven build systems; C89-clean under `-std=c90 -pedantic -Wdeclaration-after-statement -Wconversion -Wsign-conversion` (generated file included); clang-format 22 clean on all changed .c/.h; SPDX/REUSE correct incl. `MIT AND Unicode-3.0`; AGENTS.md conventions pass throughout (ares_malloc/ares_free, no libc ctype, LCOV tags, include order).
- **Tests**: 14/14 targeted suites green in normal and ASAN builds, zero reports; `IDNAAllocFail` exhaustively covers all 11 measured allocation points; every vector requested in the prior review is present; byte-level scan found no invisible codepoints in any touched file; CI 28/28 on the head push.

src/lib/str/unicode-codegen.py
```diff
@@ -0,0 +1,332 @@
...
+        if entry["code_min"] <= cp <= entry["code_max"]:
+            return entry["status"]
+    return 0  # valid
+
+# Iterate to a fixpoint: a flip can in principle make another entry's
+# mapping target disallowed, and a single pass over a list it mutates would
+# be order-dependent.  (No current Unicode version needs a second pass since
+# mapping targets are never themselves mapped, but this keeps the result
+# well-defined regardless of input.)
+changed = True
```
bradh352, line 232: **[Nit]** The fixpoint handles mapping targets that become *disallowed*, but a target that is itself *mapped*, *ignored*, or uppercase ASCII would be emitted verbatim, and the single-pass runtime mapper would then produce non-canonical output. Zero such cases exist in Unicode 17 (verified mechanically: no mapping target is mapped, ignored, or in A-Z), but the script's philosophy is hard-error over silent wrongness and a future data file could regress silently. After the fixpoint, hard-error if any remaining mapped entry has a target whose status is mapped/ignored (for cp >= 0x80) or falls in A-Z.

src/lib/str/unicode-codegen.py
```diff
@@ -0,0 +1,332 @@
...
+        offset = pool_index[utf8]
+    else:
+        offset = len(pool)
+        pool_index[utf8] = offset
+        pool.extend(utf8)
+    entry["map_offset"] = offset
+    entry["map_len"] = len(utf8)
+
+if len(pool) > 0xFFFFFFFF:
+    print(f"pool exceeds unsigned int range")
```
bradh352: **[Nit]** Last sliver of the prior codegen bundle: this f-string has no placeholders -- drop the `f`.

src/lib/include/ares_buf.h
```diff
@@ -609,6 +680,16 @@ CARES_EXTERN ares_bool_t ares_buf_begins_with(const ares_buf_t    *buf,
  */
 CARES_EXTERN size_t ares_buf_len(const ares_buf_t *buf);
 
+/*! Length of unprocessed remaining data in utf8/wide characters.
+ *
+ *  \param[in]  buf Initialized buffer object
+ *  \param[out] len Number of utf8 characters
+ *  \return ARES_SUCCESS on success, or error.
+ */
+CARES_EXTERN ares_status_t ares_buf_len_utf8(const ares_buf_t *buf,
```
bradh352, line 696: **[Nit]** "utf8/wide characters" suggests `wchar_t` units -- it counts Unicode codepoints; and `\return ARES_SUCCESS on success, or error.` is the only new declaration in this header without enumerated codes, while the sibling codepoint functions added in the same commit enumerate theirs. Actual: `ARES_EFORMERR` (NULL args), `ARES_EBADSTR` (invalid UTF-8), success with `*len == 0` on an empty buffer.

src/lib/include/ares_buf.h
```diff
@@ -338,6 +364,13 @@ CARES_EXTERN ares_status_t ares_buf_tag_fetch_strdup(const ares_buf_t *buf,
 CARES_EXTERN ares_status_t ares_buf_tag_fetch_constbuf(const ares_buf_t *buf,
                                                        ares_buf_t **newbuf);
 
+/*! Determine if entire remaining buffer is ASCII-printable.  If there are
+ *  no remaining bytes, returns ARES_FALSE.
+ *
+ *  \param[in] buf Initialized buffer object
+ *  \return ARES_TRUE if data is ascii-printable */
+CARES_EXTERN ares_bool_t ares_buf_isprint(const ares_buf_t *buf);
```
bradh352, line 373: **[Nit]** The doc block closes with `*/` on the `\return` line -- the only one of this header's ~75 function doc blocks formatted that way. While touching it: `ares_buf_isprint(NULL)` returning `ARES_FALSE` is tested in BufMisuse but undocumented (only the empty case is called out).

src/lib/str/ares_buf.c
```diff
@@ -787,6 +995,44 @@ size_t ares_buf_consume_until_charset(ares_buf_t          *buf,
...
+                                     const unsigned char *charset, size_t len,
+                                     ares_bool_t require_charset)
+{
+  size_t               remaining_len = 0;
+  const unsigned char *ptr           = ares_buf_fetch(buf, &remaining_len);
+  ares_ssize_t         pos;
+  ares_bool_t          found = ARES_FALSE;
+
+  if (ptr == NULL || charset == NULL || len == 0) {
+    return require_charset ? SIZE_MAX : 0;
```
bradh352, line 1008: **[Nit]** Invalid-parameter corner contradicts the `ARES_FALSE` contract: non-empty buffer but `charset == NULL || len == 0` with `require_charset == ARES_FALSE` returns 0 without consuming, while the header promises ARES_FALSE "will simply consume the rest of the buffer" -- semantically an empty charset is "not found", which should consume. Pure parameter-misuse corner, zero callers hit it, and it mirrors `consume_until_charset`'s guard -- so a one-line doc addition ("on invalid parameters nothing is consumed and 0/SIZE_MAX is returned") is enough.

src/lib/str/ares_strsplit.c
```diff
@@ -72,10 +72,14 @@ char **ares_strsplit(const char *in, const char *delms, size_t *num_elm)
     return NULL;
   }
 
-  status = ares_buf_split_str(
-    buf, (const unsigned char *)delms, ares_strlen(delms),
-    ARES_BUF_SPLIT_NO_DUPLICATES | ARES_BUF_SPLIT_CASE_INSENSITIVE, 0, &out,
-    num_elm);
+  /* UTF8 is permitted since every current consumer splits search domain
```
bradh352, line 75: **[Nit]** This rationale lives only here: `ares_strsplit()` previously failed on any non-printable-ASCII input and now accepts validated UTF-8 tokens, but ares_strsplit.h still describes only delimiters/dedup. A future caller reads the header, not this comment. One sentence there: "Tokens may contain validated UTF-8 in addition to printable ASCII; input containing invalid UTF-8 fails the entire split (returns NULL)."

src/lib/ares_sysconfig.c
```diff
@@ -502,6 +502,52 @@ static ares_status_t
...
+{
+  size_t i;
+  size_t cnt = 0;
+
+  for (i = 0; i < sysconfig->ndomains; i++) {
+    char *domain = sysconfig->domains[i];
+
+    sysconfig->domains[i] = NULL;
+
+    if (!ares_str_isprint(domain, ares_strlen(domain))) {
```
bradh352, line 515: **[Nit]** An ASCII control byte survives the conversion path un-dropped: the `!ares_str_isprint()` gate sends control-byte-containing entries to IDNA encode, but the mapper's ASCII fast-path passes control bytes through -- `ares_idna_encode_domain("f\x01oo.\xC3\xBC.com")` returns `ARES_SUCCESS` with `"f\x01oo.xn--tda.com"` (empirically confirmed), so the entry is *kept* with the raw byte rather than dropped, and every query using it later fails at query-write. Practically impossible input (registry/confstr suffix with embedded control bytes -- resolv.conf values already reject controls at fetch), so take-or-leave; the cheap tightening is to drop entries whose encoded result still fails `ares_str_isprint()`.

src/lib/ares_sysconfig.c
```diff
@@ -502,6 +502,52 @@ static ares_status_t
...
+      char         *encoded = NULL;
+      ares_status_t status  = ares_idna_encode_domain(domain, &encoded);
+
+      if (status == ARES_ENOMEM) {
+        ares_free(domain);
```
bradh352: **[Nit]** `ares_free(domain)` appears on all three outcomes of this 15-line block (ENOMEM return, drop-continue, success-replace). Hoisting a single `ares_free(domain);` immediately after the `ares_idna_encode_domain()` call reads cleaner: free, then `if (status == ARES_ENOMEM) return status; if (status != ARES_SUCCESS) continue; domain = encoded;`. Cosmetic only.

test/ares-test-init.cc
```diff
@@ -451,6 +451,33 @@ CONTAINED_TEST_F(LibraryTest, ContainerFullResolvInit,
...
+  EXPECT_EQ(ARES_SUCCESS, ares_init(&channel));
+
+  EXPECT_EQ((size_t)3, channel->ndomains);
```
bradh352, line 471: **[Nit]** (residual of the prior container-test comment) The added guard sits after `EXPECT_EQ((size_t)3, channel->ndomains);`, which itself dereferences `channel` -- if `ares_init()` fails, `channel` stays nullptr and the container child still crashes on this line before the `channel == nullptr` check can help. Hoist `if (channel == nullptr) return true;` directly after the init EXPECT, keeping the `ndomains != 3` check where it is.

src/lib/str/ares_punycode.c
bradh352: Fixed in fab6eca2: a truncated variable-length integer now returns `ARES_EBADNAME`, with an explicit `xn--9` EBADNAME pin plus an `invalid_puny[]` row.

src/lib/str/ares_punycode.c
bradh352: Implemented in fab6eca2 rather than just documenting: `ares_idna_map_buf()` now tracks consumed/emitted codepoints per label and returns `ARES_EBADNAME` for a label that consumed input but emitted nothing; blank labels present in the input still pass through. Tests cover `a.\xC2\xAD.b` and a lone soft hyphen; the header documents the VerifyDnsLength-minimum behavior.

src/lib/str/ares_punycode.c
bradh352: Fixed in fab6eca2 with the suggested overflow-proof form (the guard moved slightly earlier in the function since separator detection now also reads the pool).

src/lib/str/ares_punycode.c
bradh352: Renamed to `done:` in both functions in fab6eca2.

src/lib/include/ares_punycode.h
bradh352: Went with (a) in fab6eca2: ares_punycode.h moved to `src/lib/include/` (Makefile.inc and includes updated), making the semi-public status explicit per AGENTS.md and matching ares_buf.h. Consumers will grow with the query-name hook and decode-for-display follow-ups, so keeping the export surface deliberate seems right.

src/lib/str/ares_punycode.h
bradh352: Reworded both the encode and decode pass-through claims with the 63-octet proviso in fab6eca2.

src/lib/include/ares_punycode.h
bradh352: Added the ASCII-bypass bullet to the header's divergence list in fab6eca2.

src/lib/str/ares_idnamap.h
bradh352: Renamed to `ARES_IDNAMAP_STATUS_*` in fab6eca2 (header, ares_punycode.c, tests; the codegen emits numeric values so the generated table is unchanged).

src/lib/str/unicode-codegen.py
bradh352: Added in fab6eca2: post-fixpoint hard-error if any remaining mapping target is mapped/ignored (cp >= 0x80) or uppercase ASCII. Regeneration verified byte-identical with the new validation in place.

src/lib/str/unicode-codegen.py
bradh352: Dropped in fab6eca2.

src/lib/include/ares_buf.h
bradh352: Rewritten in fab6eca2: counts Unicode codepoints, EBADSTR/EFORMERR enumerated, empty-buffer case noted.

src/lib/include/ares_buf.h
bradh352: Fixed in fab6eca2: `*/` on its own line, NULL behavior documented.

src/lib/str/ares_buf.c
bradh352: Documented in fab6eca2 on both `ares_buf_consume_until_charset()` and `ares_buf_consume_last_charset()`.

src/lib/str/ares_strsplit.c
bradh352: Added the sentence to ares_strsplit.h in fab6eca2.

src/lib/ares_sysconfig.c
bradh352: Took the cheap tightening in fab6eca2: entries whose encoded result still fails `ares_str_isprint()` are dropped, with a hand-built-array test case (`ares_strsplit()` can't construct one since it rejects control bytes itself).

src/lib/ares_sysconfig.c
bradh352: Hoisted to a single `ares_free(domain)` in fab6eca2.

test/ares-test-init.cc
bradh352: Hoisted the nullptr guard above the ndomains EXPECT in fab6eca2.

PR #1064, opened 2025-12-30
Fix incorrect check for empty wide string

### LINKED ISSUE #1056: c-ares error 11 Could not contact DNS servers on Windows
It starts after this commit https://github.com/c-ares/c-ares/commit/1d1b3d4a808dff9359cacb4539ac1b775876dcb6
On previous it works fine.

Linux and MacOs did not affected.

Is there anything I could do to help to debug this issue?

https://github.com/c-ares/c-ares/pull/1034

dankmeme01:
Fixes #1056 

The commit https://github.com/c-ares/c-ares/commit/1d1b3d4a808dff9359cacb4539ac1b775876dcb6 refactored the function to use wide strings, but didn't touch this check. Because an empty wide string would now be size 2 and not 1, the empty string would go on and cause the DNS domain list to be replaced with nothing.

### DIFF
--- a/src/lib/ares_sysconfig_win.c
+++ b/src/lib/ares_sysconfig_win.c
@@ -106,7 +106,7 @@ static ares_bool_t get_REG_SZ(HKEY hKey, const WCHAR *leafKeyName, char **outptr
 
   /* Get the value for real */
   res = RegQueryValueExW(hKey, leafKeyName, 0, NULL, (BYTE *)val, &size);
-  if (res != ERROR_SUCCESS || size == 1) {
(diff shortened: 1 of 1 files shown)


dankmeme01:
Appears to be because of this part https://github.com/c-ares/c-ares/blob/73bc590b8ee80b4eee8ae637f5089584b217f3b0/src/lib/ares_sysconfig_win.c#L641-L643

The faulty check leads to `get_SuffixList_Windows` returning a truthy value, despite `line` being an empty string. And I'm guessing an empty string causes `ares_strsplit` to return a null pointer, which leads to an unsuccessful return (thus freeing the entire sysconfig)

PR #1087, opened 2026-02-15
SVCB key 7 dohpath is UTF8 string

dimbleby:
I have ducked adding UTF-8 handling to adig.c - my interest is as a user of c-ares, not as a user of adig, and that feels like it might be a non-trivial effort to do properly.

In practice I guess treating as ASCII would likely work out most of the time

eg (via original dig):
```
$ dig SVCB _dns.resolver.arpa +short
1 one.one.one.one. alpn="h2,h3" port=443 ipv4hint=1.1.1.1,1.0.0.1 ipv6hint=2606:4700:4700::1111,2606:4700:4700::1001 key7="/dns-query{?dns}"
2 one.one.one.one. alpn="dot" port=853 ipv4hint=1.1.1.1,1.0.0.1 ipv6hint=2606:4700:4700::1111,2606:4700:4700::1001
```
which apparently does not recognise "key7" but does know or guess that it looks like a string


dimbleby:
added key 8 "ohttp" also - I did not find it in the wild but the spec is there

PR #1089, opened 2026-02-28
More record types

dimbleby:
Adding support for more record types: DS, SSHFP, RRSIG, NSEC, DNSKEY, NSEC3, NSEC3PARAM

Per the commit history: I let the AI loose here.  I have nudged it towards a couple of fixes and extra testcases, but this is generated code (and commit messages).  Of course I have reviewed to the best of my ability and am responsible for this merge request.  It looks to me as though the AI has done a fine job - well known protocols and easy-to-copy coding patterns should be a friendly place for it.

Up to you whether AI involvement makes you feel that this needs more - or less! - careful review than otherwise, or maybe is already enough reason to say "no thanks".   I shan't be upset either way.

I don't know whether you even want c-ares to add support for more record types!  If not, no problem.  If yes - I expect it should be straightforward to generate more like this.


bradh352:
Looks reasonable at first glance, I'd have to look at the RFCs though to validate the wire formats.  I don't have a problem with using AI assistants as long as the human responsible for the agent is validating the output :)

I'd think c-ares should support all the DNS records types we can.

bradh352 (commented):
## Review: More record types (DS, SSHFP, RRSIG, NSEC, DNSKEY, NSEC3, NSEC3PARAM)

Overall this is a high-quality PR that faithfully mirrors the established TLSA/SIG parser/writer/mapping patterns. I did a fairly deep pass (RFC/IANA cross-check, completeness sweep, build + ASAN test run). Summary of what I independently verified as **correct**:

- **Type numbers & wire formats** — all 7 match IANA / RFC 4034 / 4255 / 5155 field-for-field, and the datatype table agrees.
- **Every enum constant** (`ares_dnssec_algorithm_t`, `ares_ds_digest_t`, `ares_sshfp_algorithm_t`, `ares_sshfp_fp_type_t`, `ares_nsec3_hash_t`) matches IANA, including the correctly-skipped reserved/unassigned values (DNSSEC 4/9/11, SSHFP 5) and the RFC citations in the comments.
- **Completeness** — every new RR key is wired into `ares_dns_rr_key_datatype()`, `ares_dns_rr_data_ptr()`, `ares_dns_rr_key_tostr()`, the `rr_*_keys[]` arrays, `ares_dns_rr_free()`, and both the parse and write dispatch switches. No table was missed.
- **Parsing safety** — only `ares_buf_*()` is used (no manual parsing/structuring); NSEC/NSEC3 bitmaps are correctly kept as opaque `BIN`. NSEC3's length-prefixed fields can over-read the RR boundary but that is safely rejected afterward by the `processed_len > rdlength` check, and reads are bounded by the total buffer — no OOB. Clean under `-Wall -Wextra -Wconversion -Wsign-conversion -Werror` (zero warnings in the new code). `LibraryTest.DNSRecord` + `LibraryTest.DNSMapping` pass, including under **ASAN**.
- The **NSEC-rejects-empty-bitmap vs. NSEC3-allows-empty-bitmap** asymmetry is correct per RFC 5155 §7.1 (empty non-terminals / opt-out).

No Major issues and no runtime-panic risks. Findings are in the inline comments. A few cross-cutting items that couldn't be anchored inline:

### [Low / question] `src/lib/ares_qcache.c:259` `ares_qcache_calc_minttl()`
This skips `SIG`/`OPT`/`SOA` from the min-TTL calc but does not add `RRSIG`. I believe the omission is **correct** — the `SIG` exclusion exists so `SIG(0)` transaction signatures (TTL 0) don't disable caching, whereas an `RRSIG` always carries a meaningful TTL equal to its covered RRset. Just flagging to confirm it was intentional.

### [Low] Test methodology
The round-trip tests are symmetric (API → write → parse → API), so a defect mirrored in both write and parse (e.g. two same-typed fields swapped, or matching wrong-endianness) would still pass. I verified the wire order against the RFCs by hand, but a single hard-coded `dig +dnssec`-captured byte vector (e.g. one RRSIG) would harden the suite against future symmetric regressions.

### [Nit / follow-up, non-blocking]
DS/DNSKEY are commonly accompanied by their child-zone counterparts **CDS (59)** / **CDNSKEY (60)**, and **KEY (25)** pairs with the existing SIG (24). Not required here — noting as a natural follow-up, which you already anticipated.

Nice work by the author — the AI-generated code holds up well against the specs.

src/lib/record/ares_dns_write.c
```diff
@@ -681,6 +681,297 @@ static ares_status_t ares_dns_write_rr_opt(ares_buf_t          *buf,
...
+    return status;
+  }
+
+  status = ares_dns_write_rr_be16(buf, rr, ARES_RR_RRSIG_KEY_TAG);
+  if (status != ARES_SUCCESS) {
+    return status;
+  }
+
+  status = ares_dns_write_rr_name(buf, rr, namelist, ARES_FALSE,
+                                  ARES_RR_RRSIG_SIGNERS_NAME);
```
bradh352, line 793: **[Medium] RRSIG signer's name must not be compressed.** RFC 4034 §3.1.7: *"A sender MUST NOT use DNS name compression on the Signer's Name field when transmitting a RRSIG RR."* `ares_dns_write_rr_name()` forwards the shared `namelist`, so `ares_dns_name_write()` can emit a `0xC0` compression pointer whenever the signer's name (usually the zone apex) already appeared earlier in the message. The resulting wire output breaks canonical-form signature validation for downstream consumers.

Note this mirrors the pre-existing `SIG` (type 24) writer, which has the same behavior — worth a follow-up to fix both. Suggested change here: bypass the namelist to force an uncompressed name:

```c
  /* RFC 4034 3.1.7: signer's name MUST NOT be compressed */
  {
    const char *name = ares_dns_rr_get_str(rr, ARES_RR_RRSIG_SIGNERS_NAME);
    if (name == NULL) {
      return ARES_EFORMERR;
    }
    status = ares_dns_name_write(buf, NULL, ARES_FALSE, name);
  }
  if (status != ARES_SUCCESS) {
    return status;
  }
```

src/lib/record/ares_dns_write.c
```diff
@@ -681,6 +681,297 @@ static ares_status_t ares_dns_write_rr_opt(ares_buf_t          *buf,
...
+static ares_status_t ares_dns_write_rr_nsec(ares_buf_t          *buf,
+                                            const ares_dns_rr_t *rr,
+                                            ares_llist_t       **namelist)
+{
+  ares_status_t        status;
+  const unsigned char *data;
+  size_t               len = 0;
+
+  status = ares_dns_write_rr_name(buf, rr, namelist, ARES_FALSE,
+                                  ARES_RR_NSEC_NEXT_DOMAIN);
```
bradh352, line 819: **[Medium] NSEC next-domain name must not be compressed.** RFC 4034 §4.1.1: *"A sender MUST NOT use DNS name compression on the Next Domain Name field when transmitting an NSEC RR."* Same root cause and fix as the RRSIG comment above — pass `NULL` instead of `namelist` to `ares_dns_name_write()` so no `0xC0` pointer is emitted.

src/lib/record/ares_dns_write.c
```diff
@@ -681,6 +681,297 @@ static ares_status_t ares_dns_write_rr_opt(ares_buf_t          *buf,
...
+
+static ares_status_t ares_dns_write_rr_rrsig(ares_buf_t          *buf,
+                                             const ares_dns_rr_t *rr,
+                                             ares_llist_t       **namelist)
+{
+  ares_status_t        status;
+  const unsigned char *data;
+  size_t               len = 0;
+
+  (void)namelist;
```
bradh352, line 753: **[Nit]** `(void)namelist;` is misleading here — `namelist` is actually used further down for the signer's name (copied from the `SIG` writer). Either remove this cast, or, if you adopt the no-compression fix above (which stops referencing `namelist`), keep it. Just make the cast and the code agree.

include/ares_dns_record.h
```diff
@@ -56,6 +56,13 @@ typedef enum {
   ARES_REC_TYPE_SRV   = 33,    /*!< RFC 2782. Server Selection. */
   ARES_REC_TYPE_NAPTR = 35,    /*!< RFC 3403. Naming Authority Pointer */
   ARES_REC_TYPE_OPT   = 41,    /*!< RFC 6891. EDNS0 option (meta-RR) */
+  ARES_REC_TYPE_DS    = 43,    /*!< RFC 4034. Delegation Signer */
+  ARES_REC_TYPE_SSHFP = 44,   /*!< RFC 4255. SSH Key Fingerprint */
```
bradh352: **[Nit]** `=` alignment is inconsistent in this block — `SSHFP = 44`, `DNSKEY = 48`, and `NSEC3PARAM = 51` don't line up with their siblings. Trivial whitespace/`clang-format` cleanup.

test/ares-test-internal.cc
```diff
@@ -826,6 +826,136 @@ TEST_F(LibraryTest, DNSRecord) {
...
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_rr_set_u8(rr, ARES_RR_NSEC3_HASH_ALGORITHM,
+      ARES_NSEC3_HASH_SHA1));
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_rr_set_u8(rr, ARES_RR_NSEC3_FLAGS, 0));
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_rr_set_u16(rr, ARES_RR_NSEC3_ITERATIONS, 10));
+  const unsigned char nsec3_salt[] = { 0xaa, 0xbb, 0xcc, 0xdd };
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_rr_set_bin(rr, ARES_RR_NSEC3_SALT, nsec3_salt,
```
bradh352, line 934: **[Low] Missing coverage for the zero-length code paths.** The commit messages specifically call out special handling for zero-length salt/bitmap via `ares_dns_rr_set_bin_own(..., NULL, 0)`, but every test here uses a non-empty salt (4 bytes) and non-empty bitmap, so those `else` branches (in both parse and write) are never exercised. Suggest adding one NSEC3 record with an **empty salt** and **empty type-bitmap** (empty non-terminal), plus an NSEC3PARAM with an empty salt, asserting the round-trip yields `len == 0`.

dimbleby:
my bot agrees with your bot on all points, except that it proposes to avoid compression more simply: by passing a null `namelist` (rather than inlining a partial copy of `ares_dns_write_rr_name`).

Updates pushed, I will resolve comments.

bradh352 (approved):

PR #1096, opened 2026-03-24
Feature: New internal error codes

### LINKED ISSUE #697: Feature: New internal error codes
Right now its not possible to differentiate between server codes and internal error codes.  We use server error codes like EFORMERR to indicate API usage issues, but it is really a server error code and they shouldn't overlap.  We need to come up with new internal error codes and change to use them consistently.

kairosci:
Resolves #697 by introducing new internal error codes (100+) distinct from DNS server response codes, enabling differentiation between server errors and internal API errors.

### DIFF
--- a/DEVELOPER-NOTES.md
+++ b/DEVELOPER-NOTES.md
@@ -45,3 +45,72 @@ Developer Notes
 * Try to keep line lengths below 80 columns and formatted as the existing code.
   There is a `.clang-format` in the repository that can be used to run the
   automated code formatter as such: `clang-format -i */*.c */*.h */*/*.c */*/*.h`
+
+## Error Code Architecture
+
+### Overview
+
+c-ares uses a two-layer error code system to maintain backward compatibility
+while providing enhanced error diagnostics internally:
+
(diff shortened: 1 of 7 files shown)


kairosci:
I introduced an internal header file (include/ares_error_codes_internal.h) that defines more than twenty new error codes, starting with values greater than 1000. These codes are intended exclusively for internal use, allowing for much more detailed diagnostics without affecting the public interface.

Next, I implemented a conversion function (src/lib/ares_error_codes_internal.c), named ares_map_internal_error(). This function is responsible for translating internal error codes into their corresponding public codes. This creates a single point of control through which all conversions pass, making the system more secure and easier to maintain.

I then integrated everything into the build system by updating the Makefiles, so that the new components are compiled and included correctly without requiring manual intervention.

Finally, I added documentation and tests. In the DEVELOPER-NOTES.md file, I described how to use the new internal codes and when to perform the conversion. The tests verify that the mechanism works correctly and that the public codes remain unchanged.


Do you think this is a good enough solution? Is there anything that needs to be corrected?

bhouse-nexthop:
sorry, I haven't had a chance to do a deep dive, you didn't need to close it ... :)

PR #1113, opened 2026-03-30
ares_process avoid recursive retry stack growth

rootvector2:
Make retry flow iterative to avoid stack growth
Add a small high retry regression test

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -58,6 +58,10 @@ static ares_status_t process_answer(ares_channel_t      *channel,
                                     ares_conn_t          *conn,
                                     const ares_timeval_t *now,
                                     ares_array_t        **requeue);
+static ares_status_t ares_send_query_int(ares_server_t *requested_server,
+                                         ares_query_t *query,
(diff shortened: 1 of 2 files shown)


src/lib/ares_process.c
```diff
@@ -573,6 +577,54 @@ static ares_status_t ares_append_endqueue(ares_array_t     **requeue,
...
+    return status;
+  }
+
+  while (idx < ares_array_len(*requeue)) {
+    ares_query_t          *query;
+    const ares_requeue_t  *entry_ptr;
+    ares_requeue_t         entry;
+    ares_status_t   internal_status;
+
+    entry_ptr = ares_array_at(*requeue, idx++);
```
bradh352, line 597: the prior code used
```
internal_status = ares_array_claim_at(&entry, sizeof(entry), requeue, 0);
```

Any reason why you changed the logic here?  I'd rather it be explicit the ownership is changing.  Plus its popping the member off the array so if there is some recursion its not going to now transfer to heap growth.

bradh352:
Fixed in 1.34.7

PR #1114, opened 2026-04-01
enforce 255-octet expanded name limit during parsing

jmestwa-coder:
## Description

The DNS name parser validates individual labels and compression pointers, but does not enforce the RFC 1035 limit of 255 octets for the fully expanded name.

Add cumulative wire-length tracking in `ares_dns_name_parse` and reject names that exceed 255 octets with `ARES_EBADNAME`. This keeps parsing behavior consistent with the existing write-side limit.

### Tests
- Names exactly at 255 octets → accepted  
- Names exceeding 255 via compression → rejected  
- Normal compressed names → unchanged  

No change in behaviour for valid inputs.

### DIFF
--- a/src/lib/record/ares_dns_name.c
+++ b/src/lib/record/ares_dns_name.c
@@ -543,6 +543,7 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
   ares_status_t status;
   ares_buf_t   *namebuf     = NULL;
   size_t        label_start = ares_buf_get_position(buf);
+  size_t        name_wire_len = 1;
 
   if (buf == NULL) {
     return ARES_EFORMERR;
@@ -636,6 +637,14 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
     }
 
     /* New label */
+
+    /* RFC 1035: full domain name (wire format) must not exceed 255 octets
+     * including label length bytes and root terminator */
+    if (name_wire_len + (size_t)c + 1 > 255) {
+      status = ARES_EBADNAME;
+      goto fail;
+    }
+    name_wire_len += ((size_t)c + 1);
 
     /* Labels are separated by periods */
     if (ares_buf_len(namebuf) != 0 && name != NULL) {
--- a/test/ares-test-parse.cc
+++ b/test/ares-test-parse.cc
@@ -26,11 +26,104 @@
 #include "ares-test.h"
 #include "dns-proto.h"
 
+#include <cstring>
 #include <sstream>
 #include <vector>
 
 namespace ares {
 namespace test {
+
+TEST_F(LibraryTest, ParseNameExactly255WireOctets) {
+  std::vector<byte> data;
+  ares_dns_record_t *dnsrec = nullptr;
+  const char        *qname  = nullptr;
+
+  PushInt16(&data, 0x1234);
+  PushInt16(&data, 0x8400);
+  PushInt16(&data, 1);
+  PushInt16(&data, 0);
+  PushInt16(&data, 0);
+  PushInt16(&data, 0);
+
+  /* QNAME: 63+63+63+61 labels => 255 octets on wire including root */
+  data.push_back(63);
+  data.insert(data.end(), 63, 'a');
+  data.push_back(63);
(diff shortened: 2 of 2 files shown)


src/lib/record/ares_dns_name.c
```diff
@@ -637,6 +638,14 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
 
     /* New label */
 
+    /* RFC 1035: full domain name (wire format) must not exceed 255 octets
+     * including label length bytes and root terminator */
+    if (name_wire_len + (size_t)c + 1 > 255) {
+      status = ARES_EBADNAME;
+      goto fail;
+    }
+    name_wire_len += ((size_t)c + 1);
```
bradh352, line 647: why track name_wire_len separately when you can just do ares_buf_len(namebuf) ?

jmestwa-coder: I was tracking the wire length explicitly to match the RFC limit
(which includes label length octets and the root terminator).

You're right that namebuf already reflects the constructed name, so it may be
possible to derive the wire length from ares_buf_len(namebuf) with some
adjustment.

I’m happy to rework this to use ares_buf_len(namebuf) instead if you prefer.

bradh352: https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.4 says the name limit is 255 characters, nothing wire format related there, I read it to mean the final formatted domain name length.

jmestwa-coder: I took another look at this.

One concern I ran into with using `ares_buf_len(namebuf)` is that `namebuf` contains the escaped presentation form rather than the encoded DNS name. It doesn't include label length octets or the root terminator, and escaped bytes can expand to multiple characters.

There's also the `name == NULL` path where `namebuf` isn't allocated, so an `ares_buf_len(namebuf)`-based check wouldn't be able to enforce the limit there.

Was your suggestion mainly to avoid the extra state variable, or am I misunderstanding the intended interpretation of the RFC limit?

bradh352 (commented):
Thanks for this — the underlying goal is right and worth doing: without a cap, the strictly-decreasing-pointer rule stops infinite loops but not unbounded per-name expansion, so a reply that points answer RRs back at one long name re-expands it per record. Capping it is good hardening.

That said, I think this should be **closed in favor of your own #1164**, which supersedes it. Two concrete reasons:

**1. The length count here is stricter than the write side, so c-ares would reject names it can itself produce.** This PR tracks *wire* length (`name_wire_len = 1` + `Σ(label + 1)` = `S + N + 1 ≤ 255`). The write side in `ares_split_dns_name()` limits *presentation* length (`S + N - 1 ≤ 255`, `ares_dns_name.c:343`). That's a 2-octet gap. A name of four 63-octet labels is accepted by the writer (presentation length 255) but this parser computes `1 + 4·64 = 257` and returns `ARES_EBADNAME`. So the PR description's statement that it "keeps parsing behavior consistent with the existing write-side limit" is unfortunately the opposite of what it does. It's also one octet stricter than #1173, which was closed for being too strict.

**2. #1164 already does this correctly.** It tracks the unescaped presentation length (`if (name_len) name_len++; name_len += c;`) so it matches `ares_split_dns_name()` exactly, documents the amplification cost, and — relevant to the discussion we had on this thread — its separate scalar counter cleanly handles both the escaped-`namebuf` concern and the `name == NULL` skip path, which is what I was really after with the `ares_buf_len()` question.

On the RFC point from the thread: I'll grant §3.1 literally frames 255 as wire-format octets (and `NS_MAXCDNAME` reflects that), but for a *parser* I'd rather stay lenient and self-consistent with our own writer than reject names that our writer and other implementations emit. If we ever want strict wire-255, it needs to change on both the write and parse sides together — a separate discussion.

Let's consolidate on #1164. Thanks again for chasing this down.

jmestwa-coder:
agreed, you're right that tracking wire length here is stricter than ares_split_dns_name() and would reject names our own writer emits, which isn't what we want. #1164 counts the presentation length so it lines up with the writer, and its separate counter handles the escaped namebuf and the name == NULL skip path cleanly. closing this in favor of #1164.

PR #1116, opened 2026-04-02
prevent integer overflow in buffer size calculation

uwezkhan:
fix: prevent integer overflow in buffer size calculation

Validate buffer growth using a safe pre-check to ensure
data_len + needed_size + 1 does not overflow size_t.

Also harden growth logic to avoid wraparound during doubling
and handle zero-sized allocations safely.

### DIFF
--- a/src/lib/str/ares_buf.c
+++ b/src/lib/str/ares_buf.c
@@ -153,24 +153,22 @@ static ares_status_t ares_buf_ensure_space(ares_buf_t *buf, size_t needed_size)
   /* When calling ares_buf_finish_str() we end up adding a null terminator,
    * so we want to ensure the size is always sufficient for this as we don't
    * want an ARES_ENOMEM at that point.
-   *
-   * This change enforces a strict upper bound invariant: all buffer capacity
-   * calculations must be validated before arithmetic, ensuring no intermediate
-   * wraparound can influence allocation decisions.
    */
   if (buf->data_len >= SIZE_MAX - 1) {
     return ARES_ENOMEM;
   }
 
-  if (needed_size > SIZE_MAX - buf->data_len - 1) {
+  needed_size++;
+
+  if (needed_size > SIZE_MAX - buf->data_len) {
     return ARES_ENOMEM;
   }
 
-  total_required = buf->data_len + needed_size + 1;
+  total_required = buf->data_len + needed_size;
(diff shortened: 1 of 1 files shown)


src/lib/str/ares_buf.c
```diff
@@ -151,41 +152,52 @@ static ares_status_t ares_buf_ensure_space(ares_buf_t *buf, size_t needed_size)
...
+   * want an ARES_ENOMEM at that point.
+   *
+   * This change enforces a strict upper bound invariant: all buffer capacity
+   * calculations must be validated before arithmetic, ensuring no intermediate
+   * wraparound can influence allocation decisions.
+   */
+  if (buf->data_len >= SIZE_MAX - 1) {
     return ARES_ENOMEM;
   }
-  needed_size++;
```
bradh352, line 158: why is this removed then just tagged onto the calculations below ... this makes the code more complex.

src/lib/str/ares_buf.c
```diff
@@ -151,41 +152,52 @@ static ares_status_t ares_buf_ensure_space(ares_buf_t *buf, size_t needed_size)
...
-  if (remaining_size >= needed_size) {
+  if (remaining_size >= needed_size + 1) {
     return ARES_SUCCESS;
   }
 
   alloc_size = buf->alloc_buf_len;
 
-  /* Not yet started */
-  if (alloc_size == 0) {
-    alloc_size = 16; /* Always shifts 1, so ends up being 32 minimum */
```
bradh352, line 178: you removed the minimum size for an ares_buf?

src/lib/str/ares_buf.c
```diff
@@ -151,41 +152,52 @@ static ares_status_t ares_buf_ensure_space(ares_buf_t *buf, size_t needed_size)
...
-  }
-
   /* Increase allocation by powers of 2 */
-  do {
-    if (alloc_size > SIZE_MAX >> 1) {
-      return ARES_ENOMEM;
+  while (alloc_size < total_required) {
```
bradh352: whats this code do that the other didn't or why this change?

uwezkhan:
i have updated the PR. waiting for the review.

uwezkhan:
@bradh352 can i get the review on this PR?

PR #1120, opened 2026-04-03
Fix/ifa netmask null

dimbleby:
The first commit handles the possibility that `ifa->ifa_netmask` is null, by defaulting netmask to zero.

Then I got worried about whether that was a good default - and discovered that no caller ever asks for the netmask anyway.  So the second commit just deletes the code that deals with it.

### DIFF
--- a/src/lib/util/ares_iface_ips.c
+++ b/src/lib/util/ares_iface_ips.c
@@ -500,20 +500,28 @@ static ares_status_t ares_iface_ips_enumerate(ares_iface_ips_t *ips,
       addr.family = AF_INET;
       memcpy(&addr.addr.addr4, &sockaddr_in->sin_addr, sizeof(addr.addr.addr4));
       /* netmask */
-      sockaddr_in = (struct sockaddr_in *)((void *)ifa->ifa_netmask);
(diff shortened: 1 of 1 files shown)


bradh352:
I'd rather not remove the netmask code, we might need it in the future.  If there's a bug in it, i'd like it to be fixed instead.

dimbleby:
The first commit on this PR deals only with the possibility that `ifa->ifa_netmask` is null: but if we keep the code we have to take _some_ default action.

Since the netmask is never actually used it is hard to guess what the right default will be.

Do you want just the first commit?

bradh352:
> The first commit on this PR deals only with the possibility that `ifa->ifa_netmask` is null: but if we keep the code we have to take _some_ default action.
> 
> Since the netmask is never actually used it is hard to guess what the right default will be.
> 
> Do you want just the first commit?

I wonder in what circumstances netmask might be NULL, but I'd agree 0 probably isn't a sane default.  I think we should just use 32 for ipv4 and 128 for ipv6.

dimbleby:
>  I think we should just use 32 for ipv4 and 128 for ipv6.

Done

PR #1131, opened 2026-04-03
Fix malloc(0) in ares_htable_all_buckets on empty table

dimbleby:
When num_keys is 0, ares_malloc_zero(0) is called. On platforms where malloc(0) returns non-NULL, both callers leak the pointer since they check for NULL-or-empty and return without freeing. On platforms where malloc(0) returns NULL, callers cannot distinguish empty from OOM. Return NULL early when the table is empty.


bradh352:
technically the statement is true, however I don't think most implementations override the default ares_malloc implementation which guarantees malloc returns NULL if size == 0.

PR #1118, opened 2026-04-03
Avoid recursive requeue stack overflows

steadytao:
## Summary
Process immediate retry requeues iteratively so repeated retryable send failures do not recurse until stack exhaustion.

## Problem
Issue #1043 reports an unbounded recursion path in retry handling. When a query is requeued immediately after a retryable send failure, the resend path can synchronously re-enter `ares_send_query()` through `ares_requeue_query()` and keep building stack frames instead of draining retries iteratively.

The earlier mock-style approximation was not a valid regression for this issue because it did not fail on unpatched code. This change therefore uses an issue-specific reproducer-backed regression instead.

## Fix
- add an internal requeue drain that processes queued resend and end-query work iteratively
- use that drain for immediate retry paths in query sending and timeout processing
- keep observable retry behaviour the same while removing recursive retry chaining

## Tests
- add a dedicated `ares_requeue_regress` target for the exact issue #1043 input
- build that regression against a sanitised static `c-ares` library on GCC/Clang non-Windows builds
- verify that unpatched code reproduces the stack overflow and the patched build does not

## Validation
Locally verified on Ubuntu 24.04:
- unpatched `ares_requeue_regress` reproduces the ASan stack overflow
- patched `ares_requeue_regress` completes cleanly
- the original external reproducer shows the same before-and-after result

Fixes #1043

Issue #1043, opened 2025-11-05
A crafted resolver configuration drives c-ares into unbounded mutual recursion between ares_requeue_query() and ares_send_query(), ultimately causing a stack overflow (ASan: stack-overflow).


weijinjinnihao:
# Bug Report

## Describe the bug  
A crafted resolver configuration drives c-ares into unbounded mutual recursion between ares_requeue_query() and ares_send_query(), ultimately causing a stack overflow (ASan: stack-overflow).

Representative trace (abridged):
```
#0  strlen
#1  ares_strlen (str/ares_str.c:58)
#2  ares_socket_configure (lib/ares_socket.c:261)
#3  ares_open_connection (lib/ares_conn.c:396)
#4  ares_send_query (lib/ares_process.c:1256)
#5  ares_requeue_query (lib/ares_process.c:948)
#6  ares_send_query (lib/ares_process.c:1267)
#7  ares_requeue_query (lib/ares_process.c:948)
#8  ares_send_query …
(repeats hundreds of times)
```
This spirals until the process aborts with ASan stack-overflow.

## Root Cause Analysis
- In ares_process.c, the requeue logic computes:
    max_tries = ares_slist_len(channel->servers) * channel->tries;
  and (when requeueing immediately) does:
    return ares_send_query(NULL, query, now);
  from inside ares_requeue_query() if requeue == NULL.

- With attacker-controlled or fuzzed options (e.g., a very large channel->tries via ares_init_options and a non-zero server count), max_tries becomes large. When the connection attempt fails quickly (e.g., because the server list/ports/families are invalid), ares_send_query() promptly calls back into ares_requeue_query() again.

- Because that call path uses immediate re-dispatch (requeue == NULL) instead of queueing/deferred processing, you get a tight, synchronous ping-pong:
ares_requeue_query() → ares_send_query() → ares_requeue_query() → …
even though try_count is increasing, the recursion depth grows to the order of max_tries before unwinding, overflowing the stack long before the loop would naturally terminate.

- The top frame shows strlen/ares_strlen each cycle because socket setup re-evaluates string options on every attempt; this is incidental. The crash is not a bad pointer in strlen, but the recursion depth itself.

## Suggested fix
Any one (or a combination) of the following will eliminate the crash and make the control flow robust:
1. Break the mutual recursion by deferring requeues (prefer iteration over recursion).
[...]

PR #1133, opened 2026-04-03
Reproducer for #1043

dimbleby:
#1118 proposes a wildly elaborate testcase.  Here is a simpler reproducer.


dimbleby:
I just noticed that the tescase in #1113 is approximately the same as this one.

PR #1134, opened 2026-04-04
validate RR counts before preallocation to reject malformed packets early

jmestwa-coder:
## Summary

Validate DNS RR counts before preallocation to reject malformed packets early.

## What’s the issue?

The parser currently trusts RR counts from the DNS header (`ANCOUNT`, `NSCOUNT`, `ARCOUNT`) without verifying that the packet contains enough data.

A malformed packet can claim a large number of records without providing sufficient payload, causing the parser to proceed with unnecessary work and incorrect handling.

## What changed?

- Added validation to ensure RR counts are feasible based on the remaining buffer size  
- Moved RR preallocation from `ares_dns_parse_header` to `ares_dns_parse_buf`  
- Perform validation after question parsing and before preallocation  
- Return `ARES_EBADRESP` early for malformed packets  

## Why this approach?

This keeps the parser aligned with its natural flow:

parse header → parse question → validate → preallocate → parse records

It avoids trusting unverified input and ensures malformed packets are rejected early and consistently.

## Test

Added a regression test (`ParseMalformedRRCount`) that reproduces the issue using a malformed packet with inconsistent RR count.

### DIFF
--- a/src/lib/record/ares_dns_parse.c
+++ b/src/lib/record/ares_dns_parse.c
@@ -923,30 +923,6 @@ static ares_status_t ares_dns_parse_header(ares_buf_t *buf, unsigned int flags,
 
   (*dnsrec)->raw_rcode = rcode;
 
-  if (*ancount > 0) {
-    status =
-      ares_dns_record_rr_prealloc(*dnsrec, ARES_SECTION_ANSWER, *ancount);
-    if (status != ARES_SUCCESS) {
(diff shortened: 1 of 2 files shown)


bradh352:
I guess a malicious packet could make us allocate about 12MB of memory unnecessarily per parse which could exhaust memory on high load systems, so this seems reasonable.  Thanks.

PR #1117, opened 2026-04-02
prevent overflow in ares_array allocation size calculations

uwezkhan:
fix: prevent overflow in ares_array allocation size calculations

Validate rounded capacity and allocation size in ares_array_set_size
to avoid integer overflow leading to undersized allocations.

Add checks for power-of-two rounding overflow and ensure that
size * member_size stays within size_t limits before allocation.

PR #1141, opened 2026-04-18
harden ares_array against integer overflow and unsafe size arithmetic

metsw24-max:
This patch adds overflow checks to size calculations in ares_array to prevent integer wraparound during memory allocation and element movement.

Changes Done:

- Added checked multiplication helper to detect overflow in size calculations
- Hardened memmove byte-size computation to prevent overflow
- Added validation in capacity growth (ares_array_set_size) to detect:
- multiplication overflow
- power-of-two rounding overflow
- Added bounds checks in insertion logic to prevent size and offset overflow
- Ensured safe failure (returning error) instead of proceeding with unsafe memory operations
- Added regression checks to verify that:
     1. oversized member_size values are safely rejected
     2. large allocation requests (e.g., SIZE_MAX) fail instead of wrapping
     
The fix improves the safety of a core container used throughout the library and aligns allocation logic with safe arithmetic practices.

### DIFF
--- a/src/lib/dsa/ares_array.c
+++ b/src/lib/dsa/ares_array.c
@@ -147,7 +147,7 @@ static ares_status_t ares_array_move(ares_array_t *arr, size_t dest_idx,
   }
 
   nmembers = arr->cnt - (src_idx - arr->offset);
-  if (ares_size_mul_overflow(nmembers, arr->member_size, &nbytes)) {
+  if (ares_size_t_mul_overflow(nmembers, arr->member_size, &nbytes)) {
     return ARES_ENOMEM;
   }
 
@@ -206,8 +206,8 @@ ares_status_t ares_array_set_size(ares_array_t *arr, size_t size)
     return ARES_SUCCESS;
   }
 
-  if (ares_size_mul_overflow(arr->alloc_cnt, arr->member_size, &orig_size) ||
-      ares_size_mul_overflow(size, arr->member_size, &new_size)) {
+  if (ares_size_t_mul_overflow(arr->alloc_cnt, arr->member_size, &orig_size) ||
+      ares_size_t_mul_overflow(size, arr->member_size, &new_size)) {
     return ARES_ENOMEM;
   }
 
--- a/src/lib/dsa/ares_array.c
+++ b/src/lib/dsa/ares_array.c
@@ -194,7 +194,7 @@ void *ares_array_finish(ares_array_t *arr, size_t *num_members)
 
 ares_status_t ares_array_set_size(ares_array_t *arr, size_t size)
 {
-  void  *temp;
+  void *temp;
   size_t orig_size;
   size_t new_size;
   size_t rounded_size;
@@ -250,7 +250,7 @@ ares_status_t ares_array_insert_at(void **elem_ptr, ares_array_t *arr,
   }
 
   if (arr->cnt == SIZE_MAX) {
-    return ARES_ENOMEM;
+    return ARES_EFORMERR;
   }
 
   /* Allocate more if needed */
--- a/src/lib/dsa/ares_array.c
+++ b/src/lib/dsa/ares_array.c
@@ -37,20 +37,6 @@ struct ares_array {
   size_t                  alloc_cnt;
 };
 
-static ares_bool_t ares_array_mult_overflow(size_t a, size_t b, size_t *out)
-{
-  if (out == NULL) {
-    return ARES_TRUE;
-  }
-
-  if (a != 0 && b > SIZE_MAX / a) {
-    return ARES_TRUE;
-  }
-
-  *out = a * b;
-  return ARES_FALSE;
-}
-
 ares_array_t *ares_array_create(size_t                  member_size,
                                 ares_array_destructor_t destruct)
 {
@@ -161,7 +147,7 @@ static ares_status_t ares_array_move(ares_array_t *arr, size_t dest_idx,
   }
 
   nmembers = arr->cnt - (src_idx - arr->offset);
-  if (ares_array_mult_overflow(nmembers, arr->member_size, &nbytes)) {
+  if (ares_size_mul_overflow(nmembers, arr->member_size, &nbytes)) {
     return ARES_ENOMEM;
   }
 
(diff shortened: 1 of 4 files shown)


bradh352:
probably need to collaborate with #1147 on creating general helpers for detecting multiplication overflow that can be used everywhere rather than a one-off per purpose.

metsw24-max:
@bradh352 i am ok to collabrate with #1147

metsw24-max:
Done. Switched this over to the same `ares_size_t_mul_overflow` helper #1147 adds, using the portable divide check so there's no builtin dependency on older toolchains. That drops the one-off here, so whichever of the two lands first the other just sheds the duplicate math hunk.

What's left on this PR is the array-specific hardening #1147 doesn't cover: the bounds and overflow guards in `ares_array_move` and the offset check in `ares_array_insert_at`.

src/lib/dsa/ares_array.c
```diff
@@ -189,8 +206,12 @@ ares_status_t ares_array_set_size(ares_array_t *arr, size_t size)
     return ARES_SUCCESS;
   }
 
-  temp = ares_realloc_zero(arr->arr, arr->alloc_cnt * arr->member_size,
-                           size * arr->member_size);
+  if (ares_size_t_mul_overflow(arr->alloc_cnt, arr->member_size, &orig_size) ||
```
bradh352, line 209: **[Coordination note, not blocking]** #1147 also rewrites `ares_array_set_size`, but routes the checked multiply through a new wrapper: `temp = ares_realloc_zero_array(arr->arr, arr->alloc_cnt, size, arr->member_size);`.

So it's not just the `ares_math.c` helper that overlaps — this function body conflicts between the two PRs. "Whichever lands first, the other sheds the duplicate math hunk" holds for the helper definition, but *this function* still has to be reconciled by whoever merges second.

Suggest converging on #1147's `ares_realloc_zero_array(...)` form here rather than keeping the two inline `ares_size_t_mul_overflow` calls, so `ares_array_set_size` ends up with a single canonical implementation. Your `ares_array_move` guard would still use `ares_size_t_mul_overflow` directly, which is fine.

bradh352 (commented):
Coordination note re: #1147 (which owns the shared `ares_size_t_mul_overflow` helper this PR consumes). One heads-up on `ares_array_set_size` — see inline. The rest of this PR (the `ares_array_move` bounds/overflow guards, the `ares_array_insert_at` count/offset guards, and the added `ArrayMisuse` overflow test) is the complementary array-specific hardening #1147 doesn't cover, which is good.

src/lib/dsa/ares_array.c
```diff
@@ -141,8 +142,16 @@ static ares_status_t ares_array_move(ares_array_t *arr, size_t dest_idx,
     return ARES_EFORMERR;
   }
 
+  if (src_idx < arr->offset || (src_idx - arr->offset) >= arr->cnt) {
```
bradh352, line 145: **Major — this introduces a regression in `ares_array_finish()`.** When `cnt == 0`, `(src_idx - arr->offset) >= arr->cnt` is `0 >= 0` -> returns `ARES_EFORMERR`. But `ares_array_finish()` calls `ares_array_move(arr, 0, arr->offset)` for any `offset != 0`, and an array drained to empty via `ares_array_remove_first()` legitimately reaches `0 < offset < alloc_cnt, cnt == 0`. On `main` that's a harmless 0-byte `memmove` and `finish()` succeeds; with this check it returns `NULL` and leaves `*num_members` unset.

Verified with a repro (insert 2 -> `remove_first` 2 -> `finish`): it fails on this branch and passes when only this hunk is reverted. The other three `ares_array_move` call sites are already guarded (`idx < cnt` / `idx+1 < cnt`) or shadowed by the existing `src_idx >= alloc_cnt` check, so this new guard adds no reachable protection — only the regression.

bradh352 (commented):
Thanks for iterating on this, and for moving to the shared `ares_size_t_mul_overflow` helper as discussed. Reviewing against current `main`, the substantive content here is now largely covered by **#1147** (the coordinated home for the shared helper + `ares_array` size-overflow hardening) plus the already-merged **#1117**. Breakdown:

- **`ares_array_set_size`** hardening here is already merged via **#1117** (`b73bb616`) and re-implemented in #1147 — so it now conflicts with `main`.
- The **helper** is byte-for-byte identical to the one #1147 adds.
- The changes unique to this PR — the `ares_array_move` bounds check and the `ares_array_insert_at` `SIZE_MAX`/offset guards — guard states that are provably unreachable given the container invariants (`offset + cnt <= alloc_cnt`, and `alloc_cnt * member_size <= SIZE_MAX` by construction, since `set_size` is the only reachable multiply-overflow site). And the `ares_array_move` bounds check actually **introduces a regression** (see inline). So there's no reachable protection left to preserve from those hunks.
- The one piece genuinely worth keeping is your **overflow test** — neither #1117 nor #1147 currently has one, so I've asked on #1147 to port it there so the coverage isn't lost, with credit to you.

Really appreciate the contribution and the collaboration on the shared helper — it fed directly into getting #1147 right.

src/lib/dsa/ares_array.c
metsw24-max: Good catch. You're right that the cnt==0 path in finish() turns that guard into a regression rather than a safeguard, since a drained-to-empty array legitimately hits 0 < offset < alloc_cnt with cnt == 0. The other move call sites are already covered, so it protects nothing reachable and should just come out.

src/lib/dsa/ares_array.c
metsw24-max: Agreed. Converging on #1147's ares_realloc_zero_array form is the right call so set_size ends up with a single canonical body. Happy to defer that to #1147.

metsw24-max:
Makes sense. With set_size already covered by #1117 and redone in #1147, and the move/insert_at guards not protecting anything reachable, there's nothing substantive left here that #1147 doesn't cover. Fine by me to port the overflow test into #1147 so the coverage still lands. I'll leave this open on your call, or close it in favour of #1147 if you'd rather consolidate.

bradh352:
closed in favor of #1147

PR #1142, opened 2026-04-18
enforce strict RDLENGTH consumption for fixed-size RDATA types

metsw24-max:
c-ares currently accepts DNS resource records where the declared RDLENGTH exceeds the number of bytes actually consumed by the type-specific parser. 
This results in trailing bytes within RDATA being silently ignored.

This patch enforces strict validation by requiring that the parser consumes exactly RDLENGTH bytes for known RDATA types.

- Enforce exact-length validation:
- Implemented in `ares_dns_parse.c`
- Added regression tests to validate the fix:
  1. `ares-test-parse-a.cc` — A record with oversized RDLENGTH
   2. `ares-test-parse-aaaa.cc` — AAAA record with oversized RDLENGTH

### DIFF
--- a/src/lib/record/ares_dns_parse.c
+++ b/src/lib/record/ares_dns_parse.c
@@ -1164,16 +1164,11 @@ static ares_status_t ares_dns_parse_rr(ares_buf_t *buf, unsigned int flags,
   /* Determine how many bytes were processed */
   processed_len = remaining_len - ares_buf_len(buf);
 
-  /* If too many bytes were processed, error! */
-  if (processed_len > rdlength) {
+  /* RDATA parsers must consume exactly rdlength bytes to avoid accepting
+   * malformed trailing data. */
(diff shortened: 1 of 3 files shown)


bradh352:
I'm a bit worried about this change.  I could see some dumb implementations preallocating more than they need and hardcoding the size just assuming downstream parsers will ignore any left-over bytes.  This could actually cause production issues in the field, and I'm not 100% sure what we'd be solving by addressing it.  There's no security risk here ... best case you catch malformed data ... but as stated before, that malformation might be tolerable.

metsw24-max:
Fair point, and I think you're right. The interop risk outweighs the benefit here: over-declared RDLENGTH from implementations that pad or hardcode sizes is the kind of malformation that's tolerable in practice, and since there's no security angle, rejecting those packets would only cost compatibility. Being lenient about trailing RDATA bytes is the safer default. Closing this.

PR #1143, opened 2026-04-20
Use safe string construction in Windows sysconfig join path

jmestwa-coder:
## Summary
- Replace the legacy `commanjoin()`/`commajoin*()` string concatenation in the Windows sysconfig join path with the safe `ares_buf_*` builders.

## Changes
- Drop `commanjoin`/`commajoin`/`commajoin_asciionly` (no more `strcat`/`strncat`).
- Build the suffix and DNS server lists with `ares_buf_t`, appending each value through a small `sysconfig_commajoin()` helper.
- A `reg_commajoin()` helper collapses the per-registry-key read/append/free boilerplate.

## Behavior
- Output format is unchanged (comma-separated values).
- The global SearchList is now ASCII-filtered like every other suffix source, so non-ASCII values are consistently stripped (matches #1034 ahead of the punycode work in #1031). Previously it was the one unfiltered path.
- Empty registry values no longer emit a spurious trailing comma.

## Risk
- Low, limited to a single Windows-only file, no public API change.

### DIFF
--- a/src/lib/ares_sysconfig_win.c
+++ b/src/lib/ares_sysconfig_win.c
@@ -134,43 +134,31 @@ static ares_bool_t get_REG_SZ(HKEY hKey, const WCHAR *leafKeyName, char **outptr
   return ARES_TRUE;
 }
 
-static void commanjoin(char **dst, const char * const src, const size_t len)
+static ares_status_t ares_buf_commajoin(ares_buf_t *buf, const char *src,
+                                        ares_bool_t asciionly)
 {
-  char  *newbuf;
-  size_t newsize;
+  ares_status_t status;
 
-  /* 1 for terminating 0 and 2 for , and terminating 0 */
-  newsize = len + (*dst ? (ares_strlen(*dst) + 2) : 1);
-  newbuf  = ares_realloc(*dst, newsize);
-  if (!newbuf) {
-    return;
+  if (buf == NULL) {
+    return ARES_ENOMEM;
   }
-  if (*dst == NULL) {
-    *newbuf = '\0';
-  }
-  *dst = newbuf;
-  if (ares_strlen(*dst) != 0) {
-    strcat(*dst, ",");
-  }
-  strncat(*dst, src, len);
-}
 
-/*
- * commajoin()
- *
- * RTF code.
- */
-static void commajoin(char **dst, const char *src)
-{
-  commanjoin(dst, src, ares_strlen(src));
-}
-
-static void commajoin_asciionly(char **dst, const char *src)
-{
-  if (!ares_str_isprint(src, ares_strlen(src))) {
-    return;
+  if (src == NULL || *src == '\0') {
+    return ARES_SUCCESS;
   }
-  commanjoin(dst, src, ares_strlen(src));
+
+  if (asciionly && !ares_str_isprint(src, ares_strlen(src))) {
+    return ARES_SUCCESS;
+  }
+
+  if (ares_buf_len(buf) > 0) {
+    status = ares_buf_append_byte(buf, ',');
+    if (status != ARES_SUCCESS) {
+      return status;
+    }
+  }
+
+  return ares_buf_append_str(buf, src);
 }
 
 /* A structure to hold the string form of IPv4 and IPv6 addresses so we can
@@ -501,8 +489,10 @@ static ares_bool_t get_DNS_Windows(char **outptr)
 
   /* Join them all into a single string, removing duplicates. */
   {
-    size_t i;
-    for (i = 0; i < addressesIndex; ++i) {
+    size_t      i;
+    ares_buf_t *buf = ares_buf_create();
+
+    for (i = 0; buf != NULL && i < addressesIndex; ++i) {
       size_t j;
       /* Look for this address text appearing previously in the results. */
       for (j = 0; j < i; ++j) {
@@ -512,10 +502,22 @@ static ares_bool_t get_DNS_Windows(char **outptr)
       }
       /* Iff we didn't emit this address already, emit it now. */
       if (j == i) {
-        /* Add that to outptr (if we can). */
-        commajoin(outptr, addresses[i].text);
+        /* Add that to buf (if we can). */
+        if (ares_buf_commajoin(buf, addresses[i].text, ARES_FALSE) !=
+            ARES_SUCCESS) {
+          ares_buf_destroy(buf);
+          buf = NULL;
+          break;
+        }
       }
     }
+
+    if (buf != NULL && ares_buf_len(buf) == 0) {
+      ares_buf_destroy(buf);
+      buf = NULL;
+    }
+
+    *outptr = ares_buf_finish_str(buf, NULL);
   }
 
 done:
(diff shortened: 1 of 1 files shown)


bradh352:
this is legacy code, I'd rather kill commanjoin() completely and use our safe string builders (ares_buf_*()) unless you've actually identified a security concern here.

jmestwa-coder:
I’ve updated the patch to remove commanjoin() and switch to the ares_buf_* APIs for string construction. I also made sure to handle return values properly and keep the behavior unchanged (including comma separation and ASCII filtering).

Let me know if you’d prefer this to be split further or adjusted in any way

jmestwa-coder:
hi @bradh352 , any update on this PR?

bradh352:
yeah, looks good.  That said, it looks like your conversion exposed a bug in the prior code where not all domains were guaranteed to be ascii only.  Since I never got a chance to finish the punycode implementation (#1031), non-ascii should always be stripped.

If you can make that change to just always strip non-ascii I'll merge this.  Thanks and sorry for the delay.  I'm finally taking a vacation so am getting around to reviewing everything.

jmestwa-coder:
@bradh352, I've updated the global SearchList path to use the same ASCII-only filtering as the other domain sources, so non-ASCII values are now consistently stripped.

bradh352 (commented):
Thanks — this matches the approval from earlier and both requested changes are in: `commanjoin()`/`commajoin*()` are gone in favor of `ares_buf_*` (no `strcat`/`strncat` left, no dangling refs), and all suffix sources — including the previously-unfiltered global SearchList — are now ASCII-stripped via `ares_buf_commajoin(..., ARES_TRUE)`. I re-verified memory/handle management (no leaks, no double-free, all `HKEY` closed, `p` freed on every path, NULL-safe buf calls) and the empty-result contract — all clean. Findings, all minor:

- **[Low]** `get_SuffixList_Windows` doesn't nullify `*outptr` on the `ares_buf_create()` OOM early-return, despite the documented contract to do so (inline). Benign today because the caller pre-inits `line = NULL`, but it's a regression vs. the old code.
- **Note (intentional):** this makes Windows strictly ASCII-filter all search domains, while Linux/macOS/BSD (`config_search()` in `ares_sysconfig_files.c`) currently don't. That's the right call here — it closes the real global-SearchList gap and matches the #1034 stopgap ahead of the punycode work in #1031, where cross-platform handling can be unified. Not a concern, just noting it for the record.
- **[Low]** `ares_buf_commajoin` naming + repeated per-key boilerplate (inline suggestions).
- **[Nit]** the PR description's "No functional changes for valid inputs" isn't quite accurate: the global SearchList is now ASCII-filtered, and empty registry values no longer emit a spurious trailing comma (the old `strcat(",")` + `strncat(src, 0)` did). Both are intended improvements — worth updating the description to reflect the filtering change.

No tests added — expected here, as this is Windows registry-reading code with no existing test harness. Not a blocker.

src/lib/ares_sysconfig_win.c
```diff
@@ -547,31 +549,50 @@ static ares_bool_t get_DNS_Windows(char **outptr)
...
+  DWORD       keyNameBuffSize;
+  DWORD       keyIdx = 0;
+  char       *p      = NULL;
+  ares_buf_t *buf    = ares_buf_create();
+
+  if (buf == NULL) {
+    return ARES_FALSE;
+  }
 
   *outptr = NULL;
```
bradh352: **[Low]** The function contract says it "nullifies `*outptr`," but on the `buf == NULL` early return `*outptr` is never written (it's set on the next line, after the return). Benign today because the sole caller pre-inits `line = NULL` and skips on `ARES_FALSE`, but it's a contract regression vs. the old code, which set `*outptr = NULL` up top. `get_DNS_Windows` gets this right (nullifies before creating the buf). Suggest reordering:

```suggestion
  ares_buf_t *buf    = ares_buf_create();

  *outptr = NULL;

  if (buf == NULL) {
    return ARES_FALSE;
  }
```

src/lib/ares_sysconfig_win.c
```diff
@@ -134,43 +134,31 @@ static ares_bool_t get_REG_SZ(HKEY hKey, const WCHAR *leafKeyName, char **outptr
   return ARES_TRUE;
 }
 
-static void commanjoin(char **dst, const char * const src, const size_t len)
+static ares_status_t ares_buf_commajoin(ares_buf_t *buf, const char *src,
```
bradh352: **[Low/Nit]** `ares_buf_commajoin` is file-`static` but wears the `ares_buf_` prefix reserved for the public buf API in `str/ares_buf.c`, so it reads like a core buf function. Suggest a local name, e.g. `commajoin_buf` or `sysconfig_commajoin`.

src/lib/ares_sysconfig_win.c
```diff
@@ -547,31 +549,50 @@ static ares_bool_t get_DNS_Windows(char **outptr)
...
 
   *outptr = NULL;
 
   /* 1. Global DNS Suffix Search List */
   if (RegOpenKeyExA(HKEY_LOCAL_MACHINE, WIN_NS_NT_KEY, 0, KEY_READ, &hKey) ==
       ERROR_SUCCESS) {
-    get_REG_SZ(hKey, SEARCHLIST_KEY, outptr);
-    if (get_REG_SZ(hKey, DOMAIN_KEY, &p)) {
-      commajoin_asciionly(outptr, p);
+    if (buf != NULL && get_REG_SZ(hKey, SEARCHLIST_KEY, &p)) {
```
bradh352: **[Low]** The `commajoin → on failure destroy+NULL → free(p)` block is copy-pasted for every registry key (~7×). A small helper collapses each site to one line and removes a class of copy-paste error:

```c
/* Read a REG_SZ value and comma-append it (ASCII-only) to *buf.
 * On allocation failure, destroys *buf and sets it NULL. */
static void reg_commajoin(ares_buf_t **buf, HKEY key, const WCHAR *leaf)
{
  char *p = NULL;
  if (*buf == NULL || !get_REG_SZ(key, leaf, &p)) {
    return;
  }
  if (ares_buf_commajoin(*buf, p, ARES_TRUE) != ARES_SUCCESS) {
    ares_buf_destroy(*buf);
    *buf = NULL;
  }
  ares_free(p);
}
```

Each call site then becomes e.g. `reg_commajoin(&buf, hKey, SEARCHLIST_KEY);`. Optional.

src/lib/ares_sysconfig_win.c
jmestwa-coder: Done, moved the *outptr = NULL above the buf == NULL return so the nullify contract holds on the OOM path too.

src/lib/ares_sysconfig_win.c
jmestwa-coder: Renamed it to sysconfig_commajoin so it no longer looks like a core buf API.

src/lib/ares_sysconfig_win.c
jmestwa-coder: Took the helper, added reg_commajoin and each site is now a one-liner.

jmestwa-coder:
Pushed the review fixes:
- nullify `*outptr` before the `buf == NULL` early return
- renamed `ares_buf_commajoin` to `sysconfig_commajoin`
- added the `reg_commajoin` helper so each registry key is a single call

Also updated the description so the ASCII-filtering and trailing-comma changes are noted instead of "no functional changes".

bradh352 (approved):
LGTM — approving. All three prior review comments are addressed at HEAD: `*outptr` is now nullified before the `ares_buf_create()` OOM early-return, `ares_buf_commajoin` → `sysconfig_commajoin`, and the per-key boilerplate is collapsed into `reg_commajoin`.

I re-audited the current state:

- **Memory/error paths clean.** No leaks, double-frees, or use-after-free on any path, including every OOM path. `p` is freed on every branch of `reg_commajoin`; `buf` is always either destroyed or consumed by `ares_buf_finish_str`; all `HKEY`s are closed; NULL-safe `ares_buf_*` calls throughout. `*outptr` is well-defined (NULL or valid string) on every TRUE/FALSE return.
- **Dead code fully removed.** `commanjoin`/`commajoin`/`commajoin_asciionly` and all `strcat`/`strncat` are gone with no lingering references.
- **Behavioral change is the intended one** — the previously-unfiltered global SearchList is now ASCII-stripped like every other suffix source (`sysconfig_commajoin(..., ARES_TRUE)`), matching #1034 ahead of the punycode work in #1031. Empty values no longer risk a spurious separator. Concatenation order and the empty-result TRUE/FALSE contract are preserved; on mid-build allocation failure the new code fails cleanly rather than returning a partial string (safer).
- **C89-clean.** No new declaration-after-statement, no `//` comments.

No tests added, which is expected here (Windows-registry-only code, no existing harness).

One coordination note, not blocking: #1022 also rewrites `get_DNS_Windows` and #1024 (punycode) will eventually supersede this always-strip-non-ASCII stopgap, so whichever lands after this will need a rebase. This being the smaller self-contained interim fix, landing it first is reasonable.

PR #1147, opened 2026-04-29
Prevent integer overflow in allocation size calculations

jmestwa-coder:
### Summary

Prevent potential integer overflow in allocation size calculations and ensure safe failure behavior.

### Changes

* Add `ares_size_t_mul_overflow` helper for checked multiplication
* Introduce overflow-safe allocation helpers:

  * `ares_malloc_zero_array`
  * `ares_realloc_zero_array`
* Apply safe allocation in `ares_parse_ns_reply` (input parsing path)
* Guard against wraparound in `ares_array_set_size` during power-of-2 growth

### Behavior

* No functional changes except:

  * returning `NULL` / `ARES_ENOMEM` on overflow
* Existing allocation and initialization semantics preserved

### Scope

* Minimal and localized changes
* No API changes or refactoring
* Only affects allocation size calculations and related safety checks


src/lib/util/ares_math.c
```diff
@@ -156,3 +156,19 @@ unsigned char ares_count_bits_u8(unsigned char x)
   static const unsigned char lookup[256] = { B6(0), B6(1), B6(1), B6(2) };
   return lookup[x];
 }
+
+ares_bool_t ares_size_t_mul_overflow(size_t a, size_t b, size_t *res)
+{
+#if defined(__GNUC__) || defined(__clang__)
```
bradh352: lets not use compile builtins.  This failed on CentOS 7 as you can see.

jmestwa-coder:
@bradh352 
I've removed the compiler builtin and replaced it with a fully portable overflow check to ensure compatibility with older toolchains like CentOS 7.

src/lib/util/ares_math.c
```diff
@@ -156,3 +156,12 @@ unsigned char ares_count_bits_u8(unsigned char x)
   static const unsigned char lookup[256] = { B6(0), B6(1), B6(1), B6(2) };
   return lookup[x];
 }
+
+ares_bool_t ares_size_t_mul_overflow(size_t a, size_t b, size_t *res)
+{
+  if (a > 0 && b > ((size_t)-1) / a) {
```
bradh352: please use SIZE_MAX instead of (size_t)-1

jmestwa-coder: Updated to use `SIZE_MAX` and added the required include.

bradh352: looks like there's a merge conflict

jmestwa-coder: conflict resolved, updated.

src/lib/util/ares_math.c
```diff
@@ -24,6 +24,7 @@
  * SPDX-License-Identifier: MIT
  */
 
+#include <limits.h>
```
bradh352: **[Change request] Remove this `#include <limits.h>`.**

Two problems:
1. It's placed **before** `ares_private.h`. Every c-ares source includes `ares_private.h` first (it pulls in `ares_setup.h`, which explicitly documents that no other header may precede it and that all sources must include `ares_private.h` first). See `ares_rand.c` for the pattern (`ares_private.h`, then system headers).
2. `<limits.h>` doesn't even define `SIZE_MAX` (the only macro used here) — that comes from `<stdint.h>`, which the project already brings in where available. `SIZE_MAX` is already visible in this file transitively via `ares_private.h` (confirmed: deleting this line still compiles).

Please just delete the line — don't substitute an (unguarded) `<stdint.h>` either, since `SIZE_MAX` is already available project-wide.

src/lib/util/ares_math.h
```diff
@@ -49,5 +49,6 @@ size_t        ares_pow(size_t x, size_t y);
 size_t        ares_count_digits(size_t n);
 size_t        ares_count_hexdigits(size_t n);
 unsigned char ares_count_bits_u8(unsigned char x);
+ares_bool_t   ares_size_t_mul_overflow(size_t a, size_t b, size_t *res);
```
bradh352: **[Nit]** Worth a one-line contract comment here — e.g. that on success it writes the product to `*res` and returns `ARES_FALSE`, and on overflow returns `ARES_TRUE` leaving `*res` untouched. #1141 already has a good version you could lift.

src/lib/dsa/ares_array.c
```diff
@@ -200,13 +195,7 @@ ares_status_t ares_array_set_size(ares_array_t *arr, size_t size)
...
   }
 
-
-  if (size > SIZE_MAX / arr->member_size) {
-    return ARES_ENOMEM;
-  }
-
-  temp = ares_realloc_zero(arr->arr, arr->alloc_cnt * arr->member_size,
-                           size * arr->member_size);
+  temp = ares_realloc_zero_array(arr->arr, arr->alloc_cnt, size, arr->member_size);
```
bradh352: **[Nit]** This line is ~83 columns; clang-format (CI) will wrap it. Running clang-format before pushing will keep the CI format check green.

bradh352 (changes requested):
Thanks — this is a solid foundational hardening PR, and it's the right home for the shared `ares_size_t_mul_overflow` helper plus the `ares_malloc_zero_array`/`ares_realloc_zero_array` wrappers (with #1141 building on the same helper per the earlier coordination). Build is clean and the relevant array / NS-reply / alloc-failure tests all pass. Prior feedback (dropping the compiler builtin, using `SIZE_MAX`, the merge conflict) was all addressed.

Requesting changes on one item: the `#include <limits.h>` added before `ares_private.h` — see the inline comment. Everything else below is nits/notes.

### Coordination with #1141 (not duplication)
This PR owns the shared helper and the alloc-array wrappers; #1141 consumes the helper and adds the complementary `ares_array_move` / `ares_array_insert_at` hardening plus the overflow test. One thing to watch on merge: **both PRs rewrite `ares_array_set_size` with different bodies** — this PR routes through `ares_realloc_zero_array(...)`, while #1141 uses two inline `ares_size_t_mul_overflow` calls + `ares_realloc_zero`. Whichever lands second has to reconcile that function, not just shed the duplicate math hunk. This PR's wrapper form is the cleaner convergence target.

### Relative to merged #1117
The `ares_array_set_size` change is an equivalent refactor of the overflow checks #1117 just landed, moved into the helper/wrapper (and drops #1117's verbose comment — fine).

### Correctness
Looks correct. The helper is sound (`a > 0 && b > SIZE_MAX / a`; `a == 0` yields product 0), the wrappers propagate overflow as `NULL` → `ARES_ENOMEM`, and dropping the explicit `memset` in `ares_parse_ns_reply` is safe because `ares_malloc_zero_array` already zeroes. No new panic paths.

### Tests (Low)
The overflow test lives downstream in #1141. If this PR may land first, a minimal `ares_malloc_zero_array` overflow-path test here would keep the helper covered regardless of merge order.

### Notes
- Security: Low — these sizes come from bounded DNS parsing / internal array growth, not attacker-steerable to `SIZE_MAX`; defense-in-depth.
- `ancount + 1` in `ares_parse_ns_reply` is a `size_t` bounded by the parsed record count, so the pre-multiply add isn't reachable to overflow — no change needed.

bradh352:
One thing worth carrying over before this merges: #1141 adds a regression test for the array size-overflow path that neither #1117 nor this PR currently has. Please port these assertions into `test/ares-test-internal.cc` (`ArrayMisuse`):

```cpp
ares_array_t *overflow_array = ares_array_create((SIZE_MAX / 2) + 1, NULL);
EXPECT_NE((void *)NULL, overflow_array);
EXPECT_NE(ARES_SUCCESS, ares_array_set_size(overflow_array, 2));
EXPECT_NE(ARES_SUCCESS, ares_array_set_size(overflow_array, SIZE_MAX));
ares_array_destroy(overflow_array);
```

That exercises both the multiply guard (`member_size * size`) and the pow2-rounding overflow guard, which are the two things this PR hardens. Credit to @metsw24-max (#1141).

bradh352:
Wanted to check on this... will be cutting a release in a few days.

src/lib/util/ares_math.c
jmestwa-coder: Deleted the line. Confirmed SIZE_MAX is still visible via ares_private.h and it builds clean.

jmestwa-coder:
Pushed the rest:

- ported the ArrayMisuse overflow assertions into test/ares-test-internal.cc (credit @metsw24-max), passes locally against both the multiply guard and the pow2-rounding guard
- added the contract comment on ares_size_t_mul_overflow
- wrapped the two long lines so the format check stays green

Build is clean and the array/ns-reply tests pass. Should be good for the release.

test/ares-test-internal.cc
```diff
@@ -1241,6 +1241,13 @@ TEST_F(LibraryTest, ArrayMisuse) {
   EXPECT_NE(ARES_SUCCESS, ares_array_remove_last(NULL));
   EXPECT_NE(ARES_SUCCESS, ares_array_sort(NULL, NULL));
   EXPECT_NE(ARES_SUCCESS, ares_array_set_size(NULL, 0));
+
+  /* Overflow in the size calculation must be rejected, not wrapped */
+  ares_array_t *overflow_array = ares_array_create((SIZE_MAX / 2) + 1, NULL);
+  EXPECT_NE((void *)NULL, overflow_array);
+  EXPECT_NE(ARES_SUCCESS, ares_array_set_size(overflow_array, 2));
+  EXPECT_NE(ARES_SUCCESS, ares_array_set_size(overflow_array, SIZE_MAX));
+  ares_array_destroy(overflow_array);
```
bradh352, line 1632: **[Nit]** These assertions are well-constructed — I traced both: `set_size(..., 2)` exercises the helper's overflow branch through `ares_realloc_zero_array`'s new-size arm, and `set_size(..., SIZE_MAX)` trips the pow2-wrap guard. One remaining hole: `ares_malloc_zero_array`'s overflow `return NULL` is never executed by the suite — its only caller is `ares_parse_ns_reply` with a DNS-bounded `ancount + 1`, so that line is dead in every test run, and Coveralls will show it uncovered in the diff. Both wrappers are `CARES_EXTERN` and already visible here via `ares_private.h`, so three lines close this plus the untested original-size arm of the realloc wrapper:

```cpp
  /* Overflow must also be rejected in the malloc-style wrapper, and in
   * the original-size arm of the realloc wrapper */
  EXPECT_EQ(NULL, ares_malloc_zero_array(SIZE_MAX, 2));
  EXPECT_EQ(NULL, ares_malloc_zero_array(2, SIZE_MAX));
  EXPECT_EQ(NULL, ares_realloc_zero_array(NULL, SIZE_MAX, 1, 2));
```

(Don't add a `(0, n)` case — `malloc(0)` is platform-defined.)

test/ares-test-internal.cc
```diff
@@ -1241,6 +1241,13 @@ TEST_F(LibraryTest, ArrayMisuse) {
   EXPECT_NE(ARES_SUCCESS, ares_array_remove_last(NULL));
   EXPECT_NE(ARES_SUCCESS, ares_array_sort(NULL, NULL));
   EXPECT_NE(ARES_SUCCESS, ares_array_set_size(NULL, 0));
+
+  /* Overflow in the size calculation must be rejected, not wrapped */
+  ares_array_t *overflow_array = ares_array_create((SIZE_MAX / 2) + 1, NULL);
+  EXPECT_NE((void *)NULL, overflow_array);
+  EXPECT_NE(ARES_SUCCESS, ares_array_set_size(overflow_array, 2));
```
bradh352: **[Nit]** `EXPECT_NE(ARES_SUCCESS, ...)` matches the surrounding misuse checks, but those return `ARES_EFORMERR` while these two paths specifically return `ARES_ENOMEM` — `EXPECT_EQ(ARES_ENOMEM, ...)` would pin that the rejection comes from the overflow guards rather than the argument validation above. Also: `SIZE_MAX` in this file currently resolves transitively (the `<climits>` at the top doesn't provide it in C++ — `<cstdint>` does); an explicit include would be more robust.

src/lib/dsa/ares_array.c
```diff
@@ -173,23 +173,18 @@ ares_status_t ares_array_set_size(ares_array_t *arr, size_t size)
...
-   * This change enforces a strict upper bound invariant: all buffer capacity
-   * calculations must be validated before arithmetic, ensuring no intermediate
-   * wraparound can influence allocation decisions.
-   */
-  rounded_size = ares_round_up_pow2(size);
-  if (rounded_size < size) {
-    return ARES_ENOMEM;
+  /* Always operate on powers of 2 */
+  {
+    size_t new_size = ares_round_up_pow2(size);
```
bradh352: **[Nit]** This swaps main's top-declared `rounded_size` for a bare scoping block — C89-legal, but bare blocks in src/lib are otherwise only used where an `#ifdef` forces them, and this grows the diff for no behavior change. Keeping main's shape is both house style and a smaller diff, and restores the wraparound rationale that got dropped along with #1117's comment:

```c
  size_t rounded_size;
  ...
  /* Always operate on powers of 2 */
  rounded_size = ares_round_up_pow2(size);
  if (rounded_size < size) {
    return ARES_ENOMEM; /* rounding wrapped around */
  }
  size = rounded_size;
```

bradh352 (approved):
## Review summary

Deep re-review of the final state (temp checkout @ bfce35c8c; CMake+Ninja clean; clang-format clean on changed lines; merges cleanly with current main — no rebase needed; array/NS-reply suites pass on the branch, on the merged state, and — deliberately — on the base with your sources reverted, which empirically confirms this is a behavior-identical refactor and the new test is a proper regression lock). All prior feedback verified addressed. Approving — everything below is a nit; feel free to fold any of them into one small push, or I can pick them up post-merge.

Things I verified this round: the division check is mathematically exact (no false accept/reject at any boundary; the `a == 0` short-circuit avoids the division); the idiom matches OpenBSD `reallocarray` and OpenSSL `safe_math.h`'s portable fallback, and TRUE-on-overflow matches both `__builtin_mul_overflow` and C23 `ckd_mul` conventions; `SIZE_MAX` is safe on every CI toolchain (Watcom/DJGPP/CentOS 7 already compile it in this exact include environment via #1116/#1117, all green); `ares_realloc_zero`'s existing `new > orig` guard means no shrink/memset-underflow path exists; zero API/ABI surface (internal headers only, additive `CARES_EXTERN` symbols matching the `ares_malloc_zero` precedent).

**Nit — `ares_parse_ptr_reply.c` is the one true sibling.** I swept every product-sized allocation in src/lib: `legacy/ares_parse_ptr_reply.c:91-96` has the byte-for-byte identical pattern to the site you converted — `ares_malloc((ancount + 1) * sizeof(*hostent->h_aliases))` plus the follow-up `memset`. Converting it too is a ~4-line hunk that goes net-negative by deleting the memset, and makes the sweep consistent. That's the only identical sibling — everything else with a multiply is either bounded by construction (fd counts, config sizes, u16-bounded parse counts × small sizeofs) or additive (#1168's territory). `ares_addrinfo2hostent.c:156-182` and `ares_sortaddrinfo.c:411` are the same *shape* with provably bounded counts — follow-up material at most.

Coordination: once this merges, #1141 gets closed — everything in it is now superseded or ruled unreachable, the overflow test has been ported here, and @metsw24-max agreed on record. I'll include the test credit ("overflow test ported from #1141 by @metsw24-max") in the squash-merge commit message since the in-thread credit doesn't otherwise land in history.

Security framing: accurate — both converted sites are defense-in-depth (a u16-bounded ancount × 8 bytes can't wrap even a 32-bit `size_t`), not reachable vulnerabilities. Worth landing regardless: it gives future duplicate overflow reports a one-line "already handled by `ares_size_t_mul_overflow`" answer.

test/ares-test-internal.cc
jmestwa-coder: added all three, plus the comment. covers the malloc wrapper's overflow return and the orig-size arm of the realloc wrapper.

test/ares-test-internal.cc
jmestwa-coder: switched both to EXPECT_EQ(ARES_ENOMEM, ...) and added <cstdint> next to <climits>.

src/lib/dsa/ares_array.c
jmestwa-coder: restored main's rounded_size shape and the wraparound comment.

jmestwa-coder:
Folded all the nits into one push:

- tests: EXPECT_EQ(ARES_ENOMEM, ...) on both overflow paths, <cstdint> for SIZE_MAX, and the three wrapper assertions so ares_malloc_zero_array's overflow return and the realloc wrapper's orig-size arm are no longer uncovered
- ares_array.c: back to main's rounded_size shape with the wraparound comment
- converted the ares_parse_ptr_reply.c sibling to ares_malloc_zero_array and dropped its memset

ArrayMisuse and the ns/ptr reply suites pass locally and the changed lines are clang-format clean. The #1141 plan and the test credit in the squash message both sound right to me.

PR #1151, opened 2026-05-02
fix: prevent undefined-behavior left shift in ares_calc_query_timeout()

### LINKED ISSUE #1144: Shift Exponent Too Large in `ares_calc_query_timeout` via Repeated Connection Failure
# Shift Exponent Too Large in `ares_calc_query_timeout` via Repeated Connection Failure

## Summary
An undefined-behavior left shift (`shift exponent 64 is too large for 64-bit
type 'size_t'`) occurs in `ares_calc_query_timeout()` when a query is
requeued 64 or more times per configured server. The exponent of the
`timeplus <<= rounds` expression is `query->try_count / num_servers` and is
never bounded before the shift, so any long-lived query against an
unresponsive server whose `ARES_OPT_TRIES` budget is sufficiently large
invokes undefined behavior.

## Version
```
$ git describe
v1.31.0-332-g46b7f689
```

## Description
`ares_calc_query_timeout()` in `src/lib/ares_process.c` doubles a query's
per-try timeout once per completed round through the server list:

```c
/* src/lib/ares_process.c ~L1138-L1160 */
static size_t ares_calc_query_timeout(const ares_query_t   *query,
                                      const ares_server_t  *server,
                                      const ares_timeval_t *now)
{
  const ares_channel_t *channel     = query->channel;
  size_t                timeout     = ares_metrics_server_timeout(server, now);
  size_t                timeplus    = timeout;
  size_t                rounds;
  size_t                num_servers = ares_slist_len(channel->servers);

  if (num_servers == 0) {
    return 0;
  }

  /* For each trip through the entire server list, we want to double the
   * retry from the last retry */
  rounds = (query->try_count / num_servers);
  if (rounds > 0) {
    timeplus <<= rounds;              /* <-- UB when rounds >= 64 */
  }

  if (channel->maxtimeout && timeplus > channel->maxtimeout) {
    timeplus = channel->maxtimeout;   /* cap is applied AFTER the shift */
  }
  ...
}
```

`size_t` is 64 bits on 64-bit platforms, so the C standard declares the
shift undefined whenever `rounds >= 64`. The existing `channel->maxtimeout`
cap runs *after* the offending expression and therefore cannot prevent the
UB — it can only clamp the already-poisoned result.

`rounds` is driven directly by `query->try_count`. Each failed attempt to
reach the server increments `try_count` via the async requeue path:

```
read_conn_packets          ares_process.c:511
 └─ handle_conn_error       ares_process.c:964
     └─ ares_close_connection  ares_close_sockets.c:62
         └─ ares_requeue_queries  ares_close_sockets.c:40
             └─ ares_requeue_query  ares_process.c:999
                 └─ ares_send_query  ares_process.c:1356
                     └─ ares_calc_query_timeout  ares_process.c:1156   ← UB
```

With a single configured server (`num_servers == 1`), `rounds` equals
`try_count`; after 64 connection failures `timeplus <<= 64` is executed.
A user can trigger this directly with the bundled `adig` tool:

```
adig -s ::1 +tries=10000000 example.com
```

(`::1:53` has no DNS listener, so every attempt fails with `ECONNREFUSED`.)

Unlike the related synchronous-recursion issue reported earlier (large
`ARES_OPT_TRIES` + failing `asocket`), this path does *not* recurse:
failures arrive asynchronously through the event loop, so `try_count`
simply grows one iteration at a time until it crosses the 64-bit shift
threshold.

## PoC Code
```cpp
// Single-server c-ares channel pointed at an unresponsive [::1]:53.
// Each failed connection attempt increments query->try_count; after 64
// such failures the shift in ares_calc_query_timeout becomes UB.
#include <ares.h>

#include <cstring>
#include <sys/select.h>

static void query_cb(void *, int, int, const unsigned char *, int) {}

int main(void) {
    ares_library_init(ARES_LIB_INIT_ALL);

    struct ares_options opts;
    std::memset(&opts, 0, sizeof(opts));

    opts.tries      = 1000000;  /* huge retry budget */
    opts.timeout    = 1;        /* ms, keeps the loop fast */
    opts.maxtimeout = 1;        /* ms, cap is applied AFTER the shift   */

    int mask = ARES_OPT_TRIES | ARES_OPT_TIMEOUTMS | ARES_OPT_MAXTIMEOUTMS;
    ares_init_options(&channel, &opts, mask);

    /* Single non-responsive server => rounds == try_count */
    ares_set_servers_csv(channel, "[::1]:53");

    ares_query(channel, "example.com", ARES_CLASS_IN, ARES_REC_TYPE_A,
               query_cb, nullptr);

    for (;;) {
        fd_set rfds, wfds;
        FD_ZERO(&rfds); FD_ZERO(&wfds);
        int nfds = ares_fds(channel, &rfds, &wfds);
        if (nfds == 0) break;

        struct timeval tv;
        struct timeval *tvp = ares_timeout(channel, nullptr, &tv);
        select(nfds, &rfds, &wfds, nullptr, tvp);
        ares_process(channel, &rfds, &wfds);
    }

    ares_destroy(channel);
    ares_library_cleanup();
    return 0;
}
```

## Reproduction Step
```
$ clang++ -g -O0 -fsanitize=address,undefined -std=c++17 \
    -Wno-deprecated-declarations \
    -I <path-to-cares>/include \
    poc.cpp <path-to-cares>/lib/libcares.a \
    -o poc

$ UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1 
$ ./poc
```

Alternative reproduction using the bundled CLI:
```
$ UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1 
# adig should be built with ASAN
$ adig -s ::1 +tries=10000000 example.com
```

## Stack Trace
```
/root/src/c-ares/src/lib/ares_process.c:1156:14: runtime error: shift exponent 64 is too large for 64-bit type 'size_t' (aka 'unsigned long')
    #0 0x5bd216498354 in ares_calc_query_timeout /root/src/c-ares/src/lib/ares_process.c:1156:14
    #1 0x5bd2164960ae in ares_send_query /root/src/c-ares/src/lib/ares_process.c:1356:14
    #2 0x5bd2164957c3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:999:12
    #3 0x5bd2164f5265 in ares_requeue_queries /root/src/c-ares/src/lib/ares_close_sockets.c:40:5
    #4 0x5bd2164f4e0b in ares_close_connection /root/src/c-ares/src/lib/ares_close_sockets.c:62:3
    #5 0x5bd2164953e8 in handle_conn_error /root/src/c-ares/src/lib/ares_process.c:964:3
    #6 0x5bd216499f9a in read_conn_packets /root/src/c-ares/src/lib/ares_process.c:511:5
    #7 0x5bd21649943e in process_read /root/src/c-ares/src/lib/ares_process.c:685:12
    #8 0x5bd216493934 in ares_process_fds_nolock /root/src/c-ares/src/lib/ares_process.c:227:14
    #9 0x5bd216494bd8 in ares_process /root/src/c-ares/src/lib/ares_process.c:380:3
    #10 0x5bd216481a3a in main /root/FuzzAgent/output/c-ares/crash_reports/crash_012/poc.cpp:54:9
    #11 0x7c87f93041c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #12 0x7c87f930428a in __libc_start_main csu/../csu/libc-start.c:360:3
    #13 0x5bd216396574 in _start (/root/FuzzAgent/output/c-ares/crash_reports/crash_012/poc+0x8c574) (BuildId: 69c13227d424d8ade1a497106f93c6c049522d5a)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /root/src/c-ares/src/lib/ares_process.c:1156:14 
```

## Suggested Fix
Clamp `rounds` to the width of `timeplus` before shifting, and apply the
`maxtimeout` cap when the shift would otherwise saturate:

```diff
   rounds = (query->try_count / num_servers);
-  if (rounds > 0) {
-    timeplus <<= rounds;
-  }
-
-  if (channel->maxtimeout && timeplus > channel->maxtimeout) {
-    timeplus = channel->maxtimeout;
-  }
+  if (rounds >= sizeof(timeplus) * CHAR_BIT) {
+    timeplus = channel->maxtimeout ? channel->maxtimeout : (size_t)-1;
+  } else if (rounds > 0) {
+    timeplus <<= rounds;
+    if (channel->maxtimeout && timeplus > channel->maxtimeout) {
+      timeplus = channel->maxtimeout;
+    }
+  } else if (channel->maxtimeout && timeplus > channel->maxtimeout) {
+    timeplus = channel->maxtimeout;
+  }
```

Equivalently, saturate early whenever `rounds` would exceed the width of
`size_t`, so the shift is only performed in its defined range.

---
> Signed-off-by: FuzzAnything fuzzanything@gmail.com

SongTonyLi:
## Summary

Fixes #1144.

`ares_calc_query_timeout()` left-shifts `timeplus` by `rounds` without bounding the exponent. When a query is requeued 64+ times for a single server (persistent connection failures with `num_servers == 1`), `rounds` reaches 64 and `timeplus <<= 64` invokes undefined behavior on a 64-bit `size_t` (C11 §6.5.7¶3).

The existing `channel->maxtimeout` clamp runs **after** the shift and cannot prevent the UB.

## Fix

Saturate `timeplus` to `SIZE_MAX` when:
- `rounds >= sizeof(timeplus) * CHAR_BIT` (shift width ≥ type width), **or**
- `timeplus > (SIZE_MAX >> rounds)` (shift would overflow).

The subsequent `maxtimeout` clamp then normalizes the value as before, preserving retry-timeout semantics.

## Verification

Tested with UBSan-instrumented build (`-fsanitize=undefined`) at base commit `46b7f689`:
- **Before**: `src/lib/ares_process.c:1156:14: runtime error: shift exponent 64 is too large for 64-bit type 'size_t' (aka 'unsigned long')`
- **After**: No UBSan findings; harness exits cleanly (exit code 0)

Pre/post validation matrix:
| Check | Pre-patch | Post-patch |
|-------|-----------|------------|
| Build | exit 0 | exit 0 |
| Reproducer exit | exit 1 (UB triggered) | exit 0 (clean) |
| UBSan shift error | present | absent |

Reproducer: configure c-ares with a single non-responsive DNS server (`127.0.0.1:53` with nothing listening), set `opts.tries >= 65`, `opts.timeout = opts.maxtimeout = 1`. After 64 connection-failure requeues through the `read_conn_packets → handle_conn_error → ares_requeue_query → ares_send_query → ares_calc_query_timeout` path, the UB fires without this patch.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -1140,7 +1140,12 @@ static size_t ares_calc_query_timeout(const ares_query_t   *query,
    * retry from the last retry */
   rounds = (query->try_count / num_servers);
   if (rounds > 0) {
-    timeplus <<= rounds;
+    if (rounds >= sizeof(timeplus) * CHAR_BIT ||
+        timeplus > (SIZE_MAX >> rounds)) {
+      timeplus = SIZE_MAX;
+    } else {
+      timeplus <<= rounds;
+    }
   }
 
   if (channel->maxtimeout && timeplus > channel->maxtimeout) {


src/lib/ares_process.c
```diff
@@ -1140,7 +1140,12 @@ static size_t ares_calc_query_timeout(const ares_query_t   *query,
    * retry from the last retry */
   rounds = (query->try_count / num_servers);
   if (rounds > 0) {
-    timeplus <<= rounds;
+    if (rounds >= sizeof(timeplus) * CHAR_BIT ||
+        timeplus > (SIZE_MAX >> rounds)) {
```
bradh352, line 1144: these 2 checks do basically the same thing, remove the redundancy.

really though, this whole thing we'd be better off making sure to cap the max attempts per resolution you could configure.  If you're approaching SIZE_MAX for a timeout you're having a really bad ... uhh, eternity?

SongTonyLi: Thanks for the review.

**On the two checks:** That's a great point, and I can see why they seem redundant. However, each check plays a slightly different role. The width check `rounds >= sizeof(timeplus) * CHAR_BIT` prevents the UB itself. The second one `timeplus > (SIZE_MAX >> rounds)` handles `rounds < 64` where the shift wraps (unsigned shift overflow is well-defined to wrap, not UB). Without it, `timeplus` can wrap to something tiny, the `maxtimeout` clamp below sees a small value and skips, and the query gets a near-zero timeout.

The example I can think of: `timeplus = 5000`, `rounds = 63`. In 64-bit, `5000 << 63 = 0` (5000 is divisible by 4, same wrap at `rounds = 62`). Without Check B that's a 0 ms timeout instead of a clamped value.

The width check is also needed to guard the second, since `SIZE_MAX >> rounds` is itself UB at `rounds >= 64`.

**On capping max attempts:** I considered both approaches.

- **Candidate A (this PR):** guard the exponent locally in `ares_calc_query_timeout()`. The function owns the backoff calc, so bounding the exponent extends its existing contract without changing caller API. Protects every path into the function, current and future. Function still returns a bounded timeout, no semantics change.
- **Candidate B (upstream tries cap):** changes global retry policy and can terminate retries earlier than the configured `channel->tries`. Leaves this shift site unsafe against any unexpected `try_count` reaching it. Policy workaround, not an arithmetic fix.

Eternal waiting may look quite scary 😅 . By the time rounds == 64, I think that the user has set `ARES_OPT_TRIES` very high and the server has been unresponsive across 64 sequential rounds. That configuration already implies "willing to wait a long time on a bad server", and saturating to SIZE_MAX then letting maxtimeout clamp preserves exactly that semantics. My understanding of user experience is that users who don't want eternity set `ARES_OPT_MAXTIMEOUTMS` and users who set huge tries without it have opted in. What do you think of that reasoning?

bradh352: i'll accept this just because some other AI evaluation will probably flag it in the future.  This is really a non-issue.

PR #1153, opened 2026-05-11
Limit GitHub Actions token permissions

xsourabhsharma:
This declares explicit read-only `GITHUB_TOKEN` permissions for CI workflows that only need repository checkout/read access.

By setting `permissions: contents: read`, these workflows no longer depend on the repository's default token permission setting. This follows GitHub Actions least-privilege guidance and reduces the impact if a build step or third-party action is compromised.

Workflows with existing custom permissions or service-specific behavior were left unchanged.

Verification:

- Parsed all workflow YAML files successfully.
- `git diff --check`


xsourabhsharma:
Closing this PR. Sorry for the noise.

Issue #639, opened 2023-11-28
Feature: Ability to cancel a specific query


bradh352:
The `ares_cancel()` function that currently exists cancels all existing queries which seems a bit absurd, I'd be surprised if anyone is really using it.  Likely if that's being called they're calling `ares_destroy()` after not realizing it will essentially do the same thing.

There can be multiple reasons for an end user to want to cancel a specific query, such as if it is user-interactive and a user hits a cancel button, or maybe the application wants to enforce an overall query timeout that isn't based on `retries * timeout` (though for the latter, implementing #135 would get rid of this particular use-case).  The point, however, is that an application's cleanup process for an outstanding c-ares query if it can't cancel it becomes much more complex and error-prone since they pass an object with the callback and cleanup of that object may be related to other objects within their application for a specific request or action.

This would obviously require a new public API so that an application can be returned an ID of some sort.  Then there's the fact that each query in c-ares isn't 1:1, meaning there are wrappers like ares_getaddrinfo() that spawn multiple internal queries, so the ID would need to reference the "parent" object and dispatch a cancel event to that which would be responsible for the cleanup of any sub-queries. 

Queries which trigger a callback prior to completion of the request (e.g. `ares_getaddrinfo()`) due to returning a cached result, invalid request data, and so on may not return an ID.  _**However**_ maybe we should rethink this behavior on the new API and delay callback dispatching as to integrators this may very well be unexpected behaviour (I think when I first started with c-ares this caught me off guard and I had to rework some assumptions). The problem is what would trigger the callback in this case since it doesn't need an fd?

Since this is specific to application cleanup for a specific query I think it should NOT trigger the actual callback with ARES_ECANCELLED and instead just document it is up to the user to clean up ... either that or we have a flag for whether or not the callback should be triggered and leave it up to the user.

PR #1154, opened 2026-05-12
ares_cancel: Avoid Allocation

RobertLeahy:
Previously ares_cancel:

1. Extracted the current list of all queries
2. Allocated a new, empty list of queries and made that the list of all queries (note that this can fail if allocation fails)
3. Iterated the list extracted in step 1 canceling all queries therein
4. Destroyed the list extracted in step 1

Note that not only does the above mean that ares_cancel could fail (due to allocation failure in step 2), but also that due to the fact it returns void that failure was silently ignored.

Updated the implementation so that ares_cancel does not allocate, and therefore always succeeds. This is accomplished by:

1. Inserting a dummy node at the end of the list of all queries
2. Traversing the list of all queries canceling each query until the dummy node inserted in step 1 is reached (dummy nodes inserted by re-entrant calls are skipped)
3. Removing the dummy node inserted in step 1

Note that since queries are only appended to the list of all queries the above formulation ensures that:

- Re-entrant calls to ares_cancel, and
- Adding queries from the callback of a cancelled query

Continue to work as before.

### DIFF
--- a/src/lib/dsa/ares_llist.c
+++ b/src/lib/dsa/ares_llist.c
@@ -383,3 +383,38 @@ void ares_llist_node_mvparent_first(ares_llist_node_t *node,
   ares_llist_node_detach(node);
   ares_llist_attach_at(new_parent, ARES__LLIST_INSERT_HEAD, NULL, node);
 }
+
+ares_llist_node_t *ares_llist_move_all_last(ares_llist_t *new_parent,
+                                            ares_llist_t *old_parent)
+{
+  ares_llist_node_t *first;
+  ares_llist_node_t *node;
+
+  if (new_parent == NULL || old_parent == NULL || new_parent == old_parent) {
+    return NULL; /* LCOV_EXCL_LINE: DefensiveCoding */
+  }
+
+  first = old_parent->head;
+  if (first == NULL) {
+    return NULL;
+  }
+
+  if (new_parent->tail != NULL) {
+    new_parent->tail->next = old_parent->head;
+    old_parent->head->prev = new_parent->tail;
+  } else {
+    new_parent->head = old_parent->head;
+  }
+  new_parent->tail = old_parent->tail;
+  new_parent->cnt += old_parent->cnt;
+
+  for (node = old_parent->head; node != NULL; node = node->next) {
+    node->parent = new_parent;
+  }
+
+  old_parent->head = NULL;
+  old_parent->tail = NULL;
+  old_parent->cnt  = 0;
+
+  return first;
+}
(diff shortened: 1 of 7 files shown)


src/lib/include/ares_llist.h
```diff
@@ -43,10 +43,13 @@ struct ares_llist;
...
-
-/*! Opaque data structure for a node in a linked list */
-typedef struct ares_llist_node ares_llist_node_t;
+/*! Internal data structure for a node in a linked list */
+typedef struct ares_llist_node {
+  void                    *data;
+  struct ares_llist_node  *prev;
+  struct ares_llist_node  *next;
+  ares_llist_t            *parent;
+} ares_llist_node_t;
```
bradh352: We should not make this public.

src/lib/dsa/ares_llist.c
```diff
@@ -289,7 +292,20 @@ static void ares_llist_node_detach(ares_llist_node_t *node)
   }
 
   node->parent = NULL;
-  list->cnt--;
+  return list;
+}
+
+void ares_llist_node_detach_sentinel(ares_llist_node_t *node)
```
bradh352: you'll probably need to slightly rearchitect this so the ares_llist_node_t doesn't need t obe public.  Even if that means pre-allocating your sentinel concept as part of the initial llist.

bradh352:
I'm not actually entirely sure what this resolves.  If you are out of memory, cancel fails.  Even if you find a way to solve that, it's not unlikely that some other path may require an allocation to complete its task.  Yes, it does essentially drop the cancel to the floor, but this is an issue with the prototype itself which should really be able to return an error.

src/lib/include/ares_llist.h
RobertLeahy: I thought this header was private to the library implementation.

src/lib/dsa/ares_llist.c
RobertLeahy: You can't preallocate because you don't know how many you need because `ares_cancel` is reentrant.

RobertLeahy:
> Even if you find a way to solve that, it's not unlikely that some other path may require an allocation to complete its task.

I thought all the other paths were taken as part of `ares_destroy` and should therefore be presumptively unable to fail. 

> this is an issue with the prototype itself which should really be able to return an error.

Even if the library could inform the caller that cancel failed what would the user do with that information? Call `ares_destroy` which already cancels? That raises the question of why `ares_cancel` can fail if it does a subset of what `ares_destroy` does.

Also locally this change is a parent of a change which adds `ares_cancel_by_arg` (#639) which imo raises the above questions but more intensely, although it may be possible to redesign that so it doesn't need a sentinel.

RobertLeahy:
I realized that while preallocating sentinels didn't seem workable, preallocating a second list and moving queries being cancelled thereinto was workable.

This is similar to the existing approach except:

- The list is preallocated when the channel is created, so `ares_cancel` doesn't need to allocate, and
- The list is shared between re-entrant `ares_cancel` calls (each re-entrant call only cancels the queries it added thereto)

PR #1155, opened 2026-05-19
ares_default_socket_functions_ex

RobertLeahy:
The non-deprecated socket function API requires callers to provide every core callback. This makes it awkward to wrap only a few default socket operations, such as observing socket open and close while otherwise preserving c-ares' native socket behavior.

The older deprecated ares_set_socket_functions API allowed NULL callbacks to fall back to c-ares' defaults, but that API is not equivalent to ares_set_socket_functions_ex and does not expose newer callbacks.

Added ares_default_socket_functions_ex, which initializes an ares_socket_functions_ex structure with c-ares' default socket callbacks. Callers can then replace individual callbacks before installing the table with ares_set_socket_functions_ex.

Also added ARES_SOCKET_FUNCTIONS_EX_VERSION so callers do not need to use a magic constant.

### DIFF
--- a/docs/ares_set_socket_functions.3
+++ b/docs/ares_set_socket_functions.3
@@ -2,7 +2,7 @@
 .\" SPDX-License-Identifier: MIT
 .TH ARES_SET_SOCKET_FUNCTIONS 3 "8 Oct 2024"
 .SH NAME
-ares_set_socket_functions, ares_set_socket_functions_ex \- Set socket io callbacks
+ares_set_socket_functions, ares_set_socket_functions_ex, ares_default_socket_functions_ex \- Set socket io callbacks
 .SH SYNOPSIS
 .nf
 #include <ares.h>
@@ -28,7 +28,7 @@ typedef enum {
 } ares_socket_bind_flags_t;
 
 struct ares_socket_functions_ex {
-  unsigned int version; /* ABI Version: must be "1" */
+  unsigned int version; /* ABI Version: must be ARES_SOCKET_FUNCTIONS_EX_VERSION */
   unsigned int flags;
 
   ares_socket_t (*asocket)(int domain, int type, int protocol, void *user_data);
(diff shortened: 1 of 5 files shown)


bhouse-nexthop:
that was an intentional design decision difference.  We didn't see a usecase where you'd be replacing one but not the others that might not cause hard to debug compatibility issues.  I think we added callback tracking if you were just trying to track connections/file descriptors.  

What is your particular use-case for overriding some but not all?

RobertLeahy:
> What is your particular use-case for overriding some but not all?

I want to replace `asocket` with a callback which:

1. Calls the default `asocket`, and if that succeeds
2. Creates a per-FD state for the newly-created socket

And to replace `aclose` with a callback which:

1. Destroys the per-FD state for that FD
2. Calls the default `aclose`

For the purposes of integrating with a different I/O back-end. I don't think this per-FD state creation can be managed via `sock_state_cb` because that doesn't have an error channel (i.e. it returns `void`).

Also, unlike the legacy API, fallback to defaults isn't implicit. The user has to explicitly choose to start from the pre-filled defaults, then override/wrap only the specific callbacks they need.

I'm unsure what "callback tracking" is supposed to refer to, maybe I'm missing something.

bradh352:
> > What is your particular use-case for overriding some but not all?
> 
> I want to replace `asocket` with a callback which:
> 
> 1. Calls the default `asocket`, and if that succeeds
> 2. Creates a per-FD state for the newly-created socket
> 
> And to replace `aclose` with a callback which:
> 
> 1. Destroys the per-FD state for that FD
> 2. Calls the default `aclose`
> 
> For the purposes of integrating with a different I/O back-end. I don't think this per-FD state creation can be managed via `sock_state_cb` because that doesn't have an error channel (i.e. it returns `void`).
> 
> Also, unlike the legacy API, fallback to defaults isn't implicit. The user has to explicitly choose to start from the pre-filled defaults, then override/wrap only the specific callbacks they need.
> 
> I'm unsure what "callback tracking" is supposed to refer to, maybe I'm missing something.

it does sound like the sock_state_cb would do what you want, but are you saying you want to be able to error out within your callback and force c-ares to abort?

RobertLeahy:
> are you saying you want to be able to error out within your callback and force c-ares to abort?

Especially if there's no other easy to use creational hook (such as overriding just `asocket`) the callback may need to allocate or perform some other operation which may fail.

My current model is to perform the possibly-failing work in `asocket` (which can report an error) but without this change that's tedious (since I need to manually reimplement every other function).

PR #1157, opened 2026-05-25
parse sortlist netmask with strtoul instead of atoi

jmestwa-coder:
Noticed parse_sort() reads the numeric sortlist mask with atoi(). The mask field accepts up to 15 digits, and atoi() on a value past INT_MAX is undefined; on common platforms it wraps, so `1.2.3.0/4294967296` is accepted as mask 0 rather than rejected. strtoul() with a 128 cap is well defined and matches the numeric parsing already used elsewhere in this file.

### DIFF
--- a/src/lib/ares_sysconfig_files.c
+++ b/src/lib/ares_sysconfig_files.c
@@ -170,11 +170,9 @@ static ares_status_t parse_sort(ares_buf_t *buf, struct apattern *pat)
 
     if (ares_str_isnum(maskstr)) {
       /* Numeric mask */
-      int mask = atoi(maskstr);
-      if (mask < 0 || mask > 128) {
-        return ARES_EBADSTR;
-      }
-      if (pat->addr.family == AF_INET && mask > 32) {
+      unsigned int mask;
+      unsigned int maxmask = (pat->addr.family == AF_INET) ? 32 : 128;
+      if (!ares_str_to_num(maskstr, maxmask, &mask)) {
         return ARES_EBADSTR;
       }
       pat->mask = (unsigned char)mask;
--- a/src/lib/ares_update_servers.c
+++ b/src/lib/ares_update_servers.c
@@ -42,6 +42,7 @@
 #ifdef HAVE_STDINT_H
 #  include <stdint.h>
 #endif
+#include <limits.h>
 
 #if defined(USE_WINSOCK)
 #  if defined(HAVE_IPHLPAPI_H)
@@ -403,7 +404,9 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
 
   if (ares_str_isnum(ll_iface)) {
     char ifname[IF_NAMESIZE] = "";
-    ll_scope                 = (unsigned int)atoi(ll_iface);
+    if (!ares_str_to_num(ll_iface, UINT_MAX, &ll_scope)) {
+      return ARES_ENOTFOUND;
+    }
     if (channel->sock_funcs.aif_indextoname == NULL ||
         channel->sock_funcs.aif_indextoname(ll_scope, ifname, sizeof(ifname),
                                             channel->sock_func_cb_data) ==
--- a/src/lib/include/ares_str.h
+++ b/src/lib/include/ares_str.h
@@ -60,6 +60,19 @@ CARES_EXTERN size_t ares_strcpy(char *dest, const char *src, size_t dest_size);
 CARES_EXTERN ares_bool_t    ares_str_isnum(const char *str);
 CARES_EXTERN ares_bool_t    ares_str_isalnum(const char *str);
 
+/*! Parse a base-10 unsigned integer from a NULL-terminated string.  Unlike
+ *  atoi(), this is well-defined on out-of-range input: empty strings, trailing
+ *  non-digit characters, values that overflow, and values greater than the
+ *  provided maximum are all rejected.
+ *
+ *  \param[in]  str  String to parse
+ *  \param[in]  max  Maximum permitted value (inclusive)
+ *  \param[out] num  Parsed value on success
+ *  \return ARES_TRUE on success, ARES_FALSE on invalid or out-of-range input
+ */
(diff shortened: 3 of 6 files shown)


src/lib/ares_sysconfig_files.c
```diff
@@ -170,8 +170,8 @@ static ares_status_t parse_sort(ares_buf_t *buf, struct apattern *pat)
 
     if (ares_str_isnum(maskstr)) {
       /* Numeric mask */
-      int mask = atoi(maskstr);
-      if (mask < 0 || mask > 128) {
+      unsigned long mask = strtoul(maskstr, NULL, 10);
```
bradh352: **Medium (consistency / no open-coded parsing):** This open-codes `strtoul(maskstr, NULL, 10)` with no `errno`/`endptr` validation, whereas #1097 (same author) established `ares_parse_port()` as the project's validated numeric-parse helper. This is also the 3rd PR in a series (#1157 / #1158 / #1162) each replacing `atoi` with a bare `strtoul` at a different site — exactly the kind of duplication that argues for a shared bounded-integer helper (e.g. `ares_str_to_num(str, max, &out)`) rather than repeating the pattern per-site.

To be clear: as written this site is **functionally correct** — `ares_str_isnum(maskstr)` already guarantees non-empty, digits-only input, and the `> 128` cap catches `ERANGE` saturation on 32-bit-`long` platforms — so adding bare `errno` checks *here alone* would just be cargo-culting. The suggestion is to standardize the series on one validated helper. Happy to defer if you'd rather land these as-is and refactor later.

bradh352 (commented):
## Summary

### Request — please consolidate the aligned PRs into one
This PR is one of three open, closely-aligned `atoi → strtoul` hardening changes from you: **#1157** (this, sortlist netmask), **#1158** (link-local interface scope), and **#1162** (numeric interface scope in `name_match`). They're the same bug class and the same fix pattern spun out of #1097. Please **combine them into a single PR** that introduces one shared validated bounded-integer helper (see the inline comment) and applies it at every site. That's much easier to review as a unit, avoids three near-duplicate diffs, and keeps the codebase consistent rather than open-coding `strtoul` three different ways. The per-PR feedback below still applies to the sortlist portion.

---

The fix itself is **correct and a genuine improvement**. `atoi()` on a value past `INT_MAX` is undefined behavior; on common platforms `atoi("4294967296")` wraps to `0`, so `1.2.3.0/4294967296` was accepted as mask `0` instead of rejected. `strtoul()` is well-defined, and the `mask > 128` check correctly rejects the overflow on all platforms — including 32-bit-`long` targets (e.g. Windows), where `strtoul` saturates to `ULONG_MAX` with `errno == ERANGE`, which is still `> 128` and still rejected. No RFC/interoperability concern: the only behavioral change is rejecting input that was previously accepted via UB, so no valid `resolv.conf` is affected. No runtime-panic risk. The conversion input is clean by construction — `maskstr` is bounded to 15 digits by `maskstr[16]` + `ares_buf_tag_fetch_bytes` (which errors rather than truncates on overflow), and `ares_str_isnum()` guarantees non-empty, digits-only input.

Two items that couldn't be left inline (the test file isn't part of this diff):

### Low — Missing regression test
The PR adds no test, and no existing test covers a numeric mask that is in-charset and buffer-fitting but out of range. The failure cases in `SetSortlistFailures` (`test/ares-test-init.cc:324`) only exercise the *over-length* mask path (`test/ares-test-init.cc:329`, a 32-char string that overflows `maskstr[16]`) — that's a length failure, not a value failure. The exact bug this PR fixes is untested and could silently regress. Suggested addition after `test/ares-test-init.cc:331`:

```c
  // Numeric netmask out of range / integer overflow (PR #1157)
  EXPECT_EQ(ARES_EBADSTR, ares_set_sortlist(channel_, "1.2.3.0/129"));
  EXPECT_EQ(ARES_EBADSTR, ares_set_sortlist(channel_, "1.2.3.0/33"));          // >32 for IPv4
  EXPECT_EQ(ARES_EBADSTR, ares_set_sortlist(channel_, "1.2.3.0/4294967296"));  // was accepted as mask 0
```

### Nit — Related identical patterns (out of scope, for series coherence)
The same unvalidated `strtoul` pattern already exists at `src/lib/ares_sysconfig_files.c:413` (ndots/timeout/retry), and raw `atoi` remains at `src/lib/ares_update_servers.c:406` and `src/lib/util/ares_iface_ips.c:340` — same bug class. Not asking to fix them here, but if a shared helper lands (see inline comment), these are the other call sites to sweep alongside the sibling PRs #1158 / #1162.

_Verified locally: built with CMake + Ninja and ran the sortlist suite (`--gtest_filter='*Sortlist*:*SortList*'`) — 35 tests pass._

src/lib/ares_sysconfig_files.c
jmestwa-coder: went with the shared helper. added ares_str_to_num(str, max, &out) in ares_str next to ares_parse_port (errno + endptr validated), and it now backs all three sites instead of three open-coded strtoul calls. the > max cap still covers the ERANGE saturation on 32-bit long.

jmestwa-coder:
Consolidated here. #1158 (link-local scope) and #1162 (name_match scope) are now folded into this PR and routed through the new ares_str_to_num(), so the sortlist netmask and both interface-scope sites share one validated parser and there's no open-coded strtoul left. Closing those two as consolidated.

Also added the out-of-range netmask cases to SetSortlistFailures (/129, /33, /4294967296) so the value-failure path is covered, not just the over-length one. Built with CMake + Ninja and the sortlist suite passes (35).

Left the ndots/timeout/retry strtoul at ares_sysconfig_files.c:413 alone since it has different zero-value semantics, but happy to sweep it into the helper as a follow-up if you want.

jmestwa-coder:
Consolidating the atoi->strtoul work into #1158 per @bradh352's note, keeping just that one. Thanks.

bradh352:
Where did you consolidate this into?  It looks like you closed all PRs

PR #1158, opened 2026-05-25
route numeric config parsing through range-checked ares_str_parse_uint

jmestwa-coder:
Consolidates the atoi->strtoul sweep into one audited path, per review. This is the single consolidated PR; #1157 and #1162 are closed in favor of it.

- add `ares_str_parse_uint()`, a strtoul-based helper with errno/end/range checks, and rework `ares_parse_port()` on top of it
- route link-local interface scope, sortlist netmask (was #1157), and `name_match` (was #1162) through it
- fold the hand-rolled numeric-service port parser in `ares_getaddrinfo.c` into `ares_parse_port()` so there's a single validated path
- also validate the previously-unchecked `strtoul` in the resolv.conf options parser (`ares_sysconfig_files.c`)

`atoi()` is undefined behavior on overflow, and the numeric tokens here are only validated as digits (up to 15 for the zone id), so an out-of-range value like `[fe80::1]:53%4294967297` was accepted instead of rejected. The helper rejects overflow, trailing garbage, and anything that doesn't fit the target width.

Behavior change worth noting: a numeric `getaddrinfo()` service that the old hand-rolled parser tolerated now returns `ARES_ESERVICE`. This covers an empty service `""` (previously yielded port 0), and leading whitespace/sign like `" 80"` / `"+80"`. This matches musl; glibc treats `""` like NULL and accepts the other two. POSIX-defensible either way.

Tests: direct `ares_str_parse_uint` unit test (`StrParseUint`) covering the boundary/overflow/garbage cases, plus a `SetServersCSV` case rejecting an out-of-range numeric scope.

Rebased on main to resolve the `ares_getaddrinfo.c` conflict with #1186.

### DIFF
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -582,23 +582,6 @@ static void host_callback(void *arg, ares_status_t status, size_t timeouts,
   /* at this point we keep on waiting for the next query to finish */
 }
 
-static ares_bool_t numeric_service_to_port(const char *service,
-                                           unsigned short *port)
-{
-  char *end;
-  unsigned long val;
-
-  errno = 0;
-  val   = strtoul(service, &end, 10);
-
-  if (errno == 0 && *end == '\0' && val <= 65535) {
-    *port = (unsigned short)val;
-    return ARES_TRUE;
-  }
-
-  return ARES_FALSE;
-}
-
 /* Per POSIX getaddrinfo(), when no node/hostname is provided the returned
  * addresses are synthesized from the service: the wildcard address when
  * ARES_AI_PASSIVE is set (suitable for bind()ing a listening socket),
@@ -688,14 +671,14 @@ static void ares_getaddrinfo_int(ares_channel_t *channel, const char *name,
 
   if (service) {
     if (hints->ai_flags & ARES_AI_NUMERICSERV) {
-      if (!numeric_service_to_port(service, &port)) {
+      if (!ares_parse_port(service, &port, ARES_TRUE)) {
         callback(arg, ARES_ESERVICE, 0, NULL);
         return;
       }
     } else {
       port = lookup_service(service, 0);
       if (!port) {
-        if (!numeric_service_to_port(service, &port)) {
+        if (!ares_parse_port(service, &port, ARES_TRUE)) {
           callback(arg, ARES_ESERVICE, 0, NULL);
           return;
         }
--- a/src/lib/ares_sysconfig_files.c
+++ b/src/lib/ares_sysconfig_files.c
@@ -27,6 +27,8 @@
 
 #include "ares_private.h"
 
+#include <limits.h>
+
 #ifdef HAVE_SYS_PARAM_H
 #  include <sys/param.h>
 #endif
@@ -170,8 +172,8 @@ static ares_status_t parse_sort(ares_buf_t *buf, struct apattern *pat)
 
     if (ares_str_isnum(maskstr)) {
       /* Numeric mask */
-      int mask = atoi(maskstr);
-      if (mask < 0 || mask > 128) {
+      unsigned int mask;
+      if (!ares_str_parse_uint(maskstr, 128, &mask)) {
         return ARES_EBADSTR;
       }
       if (pat->addr.family == AF_INET && mask > 32) {
@@ -414,8 +416,11 @@ static ares_status_t process_option(ares_sysconfig_t *sysconfig,
 
   key = kv[0];
   if (num == 2) {
-    val    = kv[1];
-    valint = (unsigned int)strtoul(val, NULL, 10);
+    val = kv[1];
+    if (!ares_str_parse_uint(val, UINT_MAX, &valint)) {
+      status = ARES_EBADSTR;
+      goto done;
+    }
   }
 
   if (ares_streq(key, "ndots")) {
--- a/src/lib/ares_update_servers.c
+++ b/src/lib/ares_update_servers.c
@@ -27,6 +27,8 @@
  */
 #include "ares_private.h"
 
+#include <limits.h>
+
 #ifdef HAVE_ARPA_INET_H
 #  include <arpa/inet.h>
 #endif
@@ -403,7 +405,13 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
 
   if (ares_str_isnum(ll_iface)) {
     char ifname[IF_NAMESIZE] = "";
-    ll_scope                 = (unsigned int)atoi(ll_iface);
+
+    /* A numeric interface index can be up to 15 digits; parse with the
+     * range-checked helper so an out-of-range value is rejected rather than
+     * silently wrapped as atoi() would */
+    if (!ares_str_parse_uint(ll_iface, UINT_MAX, &ll_scope)) {
+      return ARES_ENOTFOUND;
+    }
     if (channel->sock_funcs.aif_indextoname == NULL ||
         channel->sock_funcs.aif_indextoname(ll_scope, ifname, sizeof(ifname),
                                             channel->sock_func_cb_data) ==
--- a/src/lib/ares_update_servers.c
+++ b/src/lib/ares_update_servers.c
@@ -406,9 +406,9 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
   if (ares_str_isnum(ll_iface)) {
     char ifname[IF_NAMESIZE] = "";
 
-    /* A numeric interface index can be up to 15 digits; parse with the
-     * range-checked helper so an out-of-range value is rejected rather than
-     * silently wrapped as atoi() would */
+    /* The interface identifier is all digits but may not fit in an
+     * unsigned int; parse with the range-checked helper rather than atoi(),
+     * whose behavior on overflow is undefined. */
     if (!ares_str_parse_uint(ll_iface, UINT_MAX, &ll_scope)) {
       return ARES_ENOTFOUND;
     }
--- a/src/lib/include/ares_str.h
+++ b/src/lib/include/ares_str.h
@@ -60,6 +60,21 @@ CARES_EXTERN size_t ares_strcpy(char *dest, const char *src, size_t dest_size);
 CARES_EXTERN ares_bool_t    ares_str_isnum(const char *str);
 CARES_EXTERN ares_bool_t    ares_str_isalnum(const char *str);
 
+/*! Parse a base-10 unsigned integer from a NULL-terminated string.  Unlike
+ *  atoi(), this rejects overflow, trailing garbage, and any value that does
+ *  not fit within the caller-provided maximum.
+ *
+ *  \param[in]  str  String to parse.  Must be non-NULL and non-empty.
+ *  \param[in]  max  Largest value considered valid.
+ *  \param[out] out  On success, receives the parsed value.  Untouched on
+ *                   failure.
+ *  \return ARES_TRUE if a valid in-range integer was parsed, ARES_FALSE
+ *          otherwise.
+ */
+CARES_EXTERN ares_bool_t    ares_str_parse_uint(const char *str,
+                                                unsigned long max,
+                                                unsigned int *out);
+
 CARES_EXTERN void           ares_str_ltrim(char *str);
 CARES_EXTERN void           ares_str_rtrim(char *str);
 CARES_EXTERN void           ares_str_trim(char *str);
--- a/src/lib/include/ares_str.h
+++ b/src/lib/include/ares_str.h
@@ -61,11 +61,14 @@ CARES_EXTERN ares_bool_t    ares_str_isnum(const char *str);
 CARES_EXTERN ares_bool_t    ares_str_isalnum(const char *str);
 
 /*! Parse a base-10 unsigned integer from a NULL-terminated string.  Unlike
- *  atoi(), this rejects overflow, trailing garbage, and any value that does
- *  not fit within the caller-provided maximum.
+ *  atoi(), this rejects overflow, leading or trailing garbage (a sign,
+ *  whitespace, or non-digit characters), and any value that does not fit
+ *  within the caller-provided maximum.
  *
- *  \param[in]  str  String to parse.  Must be non-NULL and non-empty.
- *  \param[in]  max  Largest value considered valid.
+ *  \param[in]  str  String to parse.  May be NULL or empty, in which case
+ *                   ARES_FALSE is returned.
+ *  \param[in]  max  Largest value considered valid.  Values above UINT_MAX
+ *                   are capped to UINT_MAX since out is an unsigned int.
  *  \param[out] out  On success, receives the parsed value.  Untouched on
  *                   failure.
  *  \return ARES_TRUE if a valid in-range integer was parsed, ARES_FALSE
--- a/src/lib/include/ares_str.h
+++ b/src/lib/include/ares_str.h
@@ -74,9 +74,8 @@ CARES_EXTERN ares_bool_t    ares_str_isalnum(const char *str);
  *  \return ARES_TRUE if a valid in-range integer was parsed, ARES_FALSE
  *          otherwise.
  */
-CARES_EXTERN ares_bool_t    ares_str_parse_uint(const char *str,
-                                                unsigned long max,
-                                                unsigned int *out);
+CARES_EXTERN ares_bool_t ares_str_parse_uint(const char *str, unsigned long max,
+                                             unsigned int *out);
 
 CARES_EXTERN void           ares_str_ltrim(char *str);
 CARES_EXTERN void           ares_str_rtrim(char *str);
--- a/src/lib/str/ares_str.c
+++ b/src/lib/str/ares_str.c
@@ -127,23 +127,36 @@ ares_bool_t ares_str_isnum(const char *str)
   return ARES_TRUE;
 }
 
-ares_bool_t ares_parse_port(const char *str, unsigned short *port,
-                            ares_bool_t allow_zero)
+ares_bool_t ares_str_parse_uint(const char *str, unsigned long max,
+                                unsigned int *out)
 {
-  char         *endptr = NULL;
+  char         *end = NULL;
   unsigned long val;
 
-  if (str == NULL || port == NULL || *str == 0) {
+  if (str == NULL || out == NULL || *str == 0) {
     return ARES_FALSE;
   }
 
   errno = 0;
-  val   = strtoul(str, &endptr, 10);
-  if (errno == ERANGE || endptr == str || *endptr != '\0') {
+  val   = strtoul(str, &end, 10);
+  if (errno == ERANGE || end == str || *end != '\0' || val > max) {
     return ARES_FALSE;
   }
 
-  if (val > 65535UL) {
+  *out = (unsigned int)val;
+  return ARES_TRUE;
+}
+
+ares_bool_t ares_parse_port(const char *str, unsigned short *port,
+                            ares_bool_t allow_zero)
+{
+  unsigned int val;
+
+  if (port == NULL) {
+    return ARES_FALSE;
+  }
+
+  if (!ares_str_parse_uint(str, 65535UL, &val)) {
     return ARES_FALSE;
   }
 
--- a/src/lib/str/ares_str.c
+++ b/src/lib/str/ares_str.c
@@ -29,6 +29,7 @@
 #include "ares_str.h"
 
 #include <errno.h>
+#include <limits.h>
 
 #ifdef HAVE_STDINT_H
 #  include <stdint.h>
@@ -133,13 +134,22 @@ ares_bool_t ares_str_parse_uint(const char *str, unsigned long max,
   char         *end = NULL;
   unsigned long val;
 
-  if (str == NULL || out == NULL || *str == 0) {
+  /* Require a leading digit so strtoul()'s tolerance of a sign or leading
+   * whitespace (e.g. "-1" wrapping to UINT_MAX) can't slip through here. This
+   * also rejects NULL and empty strings. */
+  if (str == NULL || out == NULL || !ares_isdigit(*str)) {
     return ARES_FALSE;
+  }
+
+  /* out is unsigned int, so a max above UINT_MAX would let strtoul's result
+   * truncate silently on assignment.  Cap it. */
+  if (max > UINT_MAX) {
+    max = UINT_MAX;
   }
 
   errno = 0;
   val   = strtoul(str, &end, 10);
-  if (errno == ERANGE || end == str || *end != '\0' || val > max) {
+  if (errno == ERANGE || *end != '\0' || val > max) {
     return ARES_FALSE;
   }
 
--- a/src/lib/util/ares_iface_ips.c
+++ b/src/lib/util/ares_iface_ips.c
@@ -25,6 +25,8 @@
  */
 #include "ares_private.h"
 
+#include <limits.h>
+
 #ifdef USE_WINSOCK
 #  include <winsock2.h>
 #  include <ws2tcpip.h>
@@ -337,8 +339,11 @@ static ares_bool_t name_match(const char *name, const char *adapter_name,
     return ARES_TRUE;
   }
 
-  if (ares_str_isnum(name) && (unsigned int)atoi(name) == ll_scope) {
-    return ARES_TRUE;
+  {
+    unsigned int scope;
+    if (ares_str_parse_uint(name, UINT_MAX, &scope) && scope == ll_scope) {
+      return ARES_TRUE;
+    }
   }
 
   return ARES_FALSE;
--- a/src/lib/util/ares_iface_ips.c
+++ b/src/lib/util/ares_iface_ips.c
@@ -331,6 +331,8 @@ static char *wcharp_to_charp(const wchar_t *in)
 static ares_bool_t name_match(const char *name, const char *adapter_name,
                               unsigned int ll_scope)
 {
+  unsigned int scope;
+
   if (name == NULL || *name == 0) {
     return ARES_TRUE;
   }
@@ -339,11 +341,8 @@ static ares_bool_t name_match(const char *name, const char *adapter_name,
     return ARES_TRUE;
   }
 
-  {
-    unsigned int scope;
-    if (ares_str_parse_uint(name, UINT_MAX, &scope) && scope == ll_scope) {
-      return ARES_TRUE;
-    }
+  if (ares_str_parse_uint(name, UINT_MAX, &scope) && scope == ll_scope) {
+    return ARES_TRUE;
   }
 
   return ARES_FALSE;
--- a/test/ares-test-internal.cc
+++ b/test/ares-test-internal.cc
@@ -26,6 +26,7 @@
 #include "ares-test.h"
 #include "dns-proto.h"
 
+#include <climits>
 #include <stdio.h>
 
 #ifdef HAVE_UNISTD_H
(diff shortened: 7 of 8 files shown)


src/lib/ares_update_servers.c
```diff
@@ -402,8 +404,20 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
...
   if (ares_str_isnum(ll_iface)) {
-    char ifname[IF_NAMESIZE] = "";
-    ll_scope                 = (unsigned int)atoi(ll_iface);
+    char          ifname[IF_NAMESIZE] = "";
+    char         *end                 = NULL;
+    unsigned long scope;
+
+    /* atoi() is undefined on overflow, and the numeric interface index can be
+     * up to 15 digits; parse with strtoul() and reject anything that overflows
+     * or doesn't fit an interface index */
```
bradh352: **[Major — change request]** This re-inlines the validated-`strtoul` idiom that #1097 already extracted into `ares_parse_port()`, and that sibling PRs #1157 and #1162 each re-inline *differently* (#1157 notably drops the `ERANGE` check entirely). Please introduce one generalized helper and route all numeric-parse sites through it:

```c
ares_bool_t ares_str_parse_uint(const char *str, unsigned long max,
                                unsigned int *out)
{
  char         *end = NULL;
  unsigned long val;

  if (str == NULL || out == NULL || *str == 0) {
    return ARES_FALSE;
  }

  errno = 0;
  val   = strtoul(str, &end, 10);
  if (errno == ERANGE || end == str || *end != '\0' || val > max) {
    return ARES_FALSE;
  }

  *out = (unsigned int)val;
  return ARES_TRUE;
}
```

Then here simply:

```c
if (!ares_str_parse_uint(ll_iface, UINT_MAX, &ll_scope)) {
  return ARES_ENOTFOUND;
}
```

This satisfies the "use a helper, don't hand-roll parsing" rule and eliminates the cross-PR divergence.

src/lib/ares_update_servers.c
```diff
@@ -402,8 +404,20 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
...
+    errno = 0;
```
bradh352: **[Nit]** Given the enclosing `if (ares_str_isnum(ll_iface))`, the string is guaranteed non-empty and all-digits, so `end == ll_iface` and `*end != '\0'` can never trigger here — harmless but dead. Only `errno == ERANGE` and the `unsigned int` fit-check are live. Moot if you adopt the shared helper above (which can legitimately drop the `isnum` pre-check).

src/lib/ares_update_servers.c
```diff
@@ -402,8 +404,20 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
 
 
   if (ares_str_isnum(ll_iface)) {
-    char ifname[IF_NAMESIZE] = "";
-    ll_scope                 = (unsigned int)atoi(ll_iface);
+    char          ifname[IF_NAMESIZE] = "";
```
bradh352: **[Nit]** The inner `scope` local sits right next to the function's `ll_scope`; minor readability wart. A shared helper writing straight into `ll_scope` removes the near-duplicate.

bradh352 (commented):
## Review summary

Thanks for the fix — the intent is sound: `atoi()` is genuinely undefined on overflow, and the numeric zone-id token is only validated as up-to-15 digits upstream, so replacing it with a range-checked `strtoul()` is the right call. Build is clean (CMake+Ninja, zero warnings on the diff) and the relevant offline tests pass. My main ask is structural.

### Please resubmit the related PRs as one (Major)
#1097 (merged), #1157, #1158, and #1162 are all one `atoi`→`strtoul` sweep, and each one re-inlines the same validation idiom *differently* — #1157 even drops the `ERANGE` check entirely. **Could you close #1157/#1158/#1162 in favor of a single consolidated PR** that:
1. introduces one generalized helper (e.g. `ares_str_parse_uint()`, with `ares_parse_port()` reworked on top of it),
2. applies it to link-local scope (this PR), sortlist netmask (#1157), and `name_match` (#1162), and
3. also fixes the still-uncovered **unchecked** `strtoul` at `src/lib/ares_sysconfig_files.c:413` (`(unsigned int)strtoul(val, NULL, 10)` — no `errno`/end check).

That gives one audited code path, one review, and one test set instead of four diverging variants. See the inline comment for a suggested helper.

### Correctness ✅
The `scope != (unsigned long)(unsigned int)scope` guard is correct and necessary on LP64 (where `strtoul("4294967297")` succeeds without `ERANGE` but must still be rejected) and a harmless no-op on LLP64/ILP32. No valid input is newly rejected.

### Security — Low
`ll_iface` comes from local configuration (`resolv.conf`, `ares_set_servers_csv`, sortlist), not a remote-attacker surface. This is hardening, not a remotely-exploitable vulnerability.

### PR description
The description slightly overstates the exploit — "wraps to index 1 (often loopback)" is implementation-specific (glibc's `atoi` clamps via `strtol` rather than wrapping mod 2^32). The accurate and sufficient justification is simply that `atoi()` is undefined behavior on overflow. Suggest rewording.

### Test coverage — Low (suggestion)
No test was added. The existing `SetServersCSV` test only exercises an interface *name* (`%iface0`), not the numeric-overflow path. A small rejection test — e.g. `ares_set_servers_csv(channel, "[fe80::1%99999999999]:53")` asserting the server is rejected — would lock in the fix, and works without a real interface since validation fails before `aif_indextoname()` is called.

### RFC / interoperability ✅
RFC 4007 numeric zone-ids are implementation-defined, and `if_indextoname()` takes an `unsigned int`, so rejecting values that don't fit `unsigned int` is compliant and breaks no valid input.

jmestwa-coder:
Consolidating the atoi->strtoul series into #1157 per bradh352's note. The link-local scope change now goes through the shared ares_str_to_num() helper there.

src/lib/ares_update_servers.c
jmestwa-coder: done. added ares_str_parse_uint() with the errno/end/max checks and reworked ares_parse_port() on top of it, then routed link-local scope, sortlist netmask, name_match, and the options parser through the same helper.

src/lib/ares_update_servers.c
jmestwa-coder: moot now with the shared helper. the isnum pre-check only stays where it selects the numeric-vs-name branch.

src/lib/ares_update_servers.c
jmestwa-coder: gone, the helper writes straight into ll_scope now.

jmestwa-coder:
Reworked into the single consolidated PR.

- added `ares_str_parse_uint()` (strtoul + errno/end/max) and rebuilt `ares_parse_port()` on it
- routed link-local scope, sortlist netmask (was #1157), and `name_match` (was #1162) through it
- fixed the unchecked `strtoul` at `ares_sysconfig_files.c:413` as well
- added a `SetServersCSV` test that rejects an overflowing numeric scope
- reworded the description; dropped the "wraps to index 1" line since that's glibc-specific, the real point is just that `atoi()` is UB on overflow

Closed #1157 and #1162 in favor of this. Builds clean and the offline suite passes.

bradh352:
Where did you consolidate this into?  It looks like you closed all PRs

jmestwa-coder:
Sorry for the confusion, that was my mistake. I closed them before the consolidated branch was actually pushed, so it looked like everything just disappeared. I've reopened this one (#1158) as the single consolidated PR and pushed the real change.

It now adds `ares_str_parse_uint()` with `ares_parse_port()` reworked on top of it, and routes link-local scope, sortlist netmask (was #1157), `name_match` (was #1162), and the previously-unchecked options-parser `strtoul` through it, plus the `SetServersCSV` overflow test. #1157 and #1162 stay closed in favor of this one.

src/lib/str/ares_str.c
```diff
@@ -127,23 +127,36 @@ ares_bool_t ares_str_isnum(const char *str)
...
-  if (str == NULL || port == NULL || *str == 0) {
+  if (str == NULL || out == NULL || *str == 0) {
     return ARES_FALSE;
   }
 
   errno = 0;
-  val   = strtoul(str, &endptr, 10);
-  if (errno == ERANGE || endptr == str || *endptr != '\0') {
+  val   = strtoul(str, &end, 10);
+  if (errno == ERANGE || end == str || *end != '\0' || val > max) {
```
bradh352: **[Low + Nit] Two small notes on the helper's contract:**

**(Low) `max` is `unsigned long` but `out` is `unsigned int`.** If a future caller ever passes `max > UINT_MAX`, the `val > max` guard passes and `*out = (unsigned int)val` then silently truncates. All three current callers pass `max ≤ UINT_MAX`, so it's safe today — but worth either clamping (`if (max > UINT_MAX) max = UINT_MAX;`) or documenting "`max` must be ≤ `UINT_MAX`".

**(Nit) Leading `+`/`-`/whitespace is accepted.** `strtoul` skips leading whitespace and honors a sign, so `" 5"`/`"+5"` parse, and `"-1"` *wraps* — on ILP32 `strtoul("-1") == UINT_MAX` without `ERANGE`, so it'd be accepted when `max == UINT_MAX`. Not a regression (`atoi`/old `strtoul` behaved identically) and the `ares_str_isnum()`-guarded callers can't reach it — but `name_match()` and `process_option()` have no such guard. Since this is meant to be *the* audited path, consider rejecting a non-digit first char inside the helper (`if (!ares_isdigit(*str)) return ARES_FALSE;`) so callers don't each have to remember the guard. The doc mentions "trailing garbage" but is silent on leading.

bradh352 (commented):
Re-reviewed the consolidated version. The intent is sound and the structural ask from the last round is satisfied: one `ares_str_parse_uint()` helper with `ares_parse_port()` reworked on top, and the link-local scope / sortlist netmask / `name_match` / options-parser sites all routed through it. Build is clean (CMake+Ninja, no new warnings on the changed files) and the relevant offline tests pass (`SetServersCSV`, `EnvInitModernOptions`, sortlist, options). C89-clean, no runtime-panic paths, and all inputs are local config rather than a remote surface — this is hardening.

A few non-blocking items, all left inline:

**Low — one port parser was missed by the consolidation.** `numeric_service_to_port()` at `src/lib/ares_getaddrinfo.c:580` is a near-identical hand-rolled `strtoul` + errno/end/`<=65535` port parser — exactly what the reworked `ares_parse_port(service, port, ARES_TRUE)` now does, and it's already called under `if (service)`. Folding it in would complete the "one audited path" goal and incidentally fix its latent empty-`service` edge (`""` currently yields port 0 rather than an error). Optional follow-up — the current code is correct as-is.

**Low — test coverage** (inline on `ares-test-misc.cc`): the added `%99999999999` case returns `ARES_ENOTFOUND` with both the old and new code, so it doesn't lock in the overflow rejection; a direct `ares_str_parse_uint()` unit test in `ares-test-internal.cc` would.

**Low/Nit — helper contract** (inline on `ares_str.c`): `max` is `unsigned long` while `out` is `unsigned int` (silent truncation if a future caller exceeds `UINT_MAX`), and the helper inherits `strtoul`'s acceptance of leading sign/whitespace, which the two un-guarded callers rely on.

**Nit — doc** (inline on `ares_str.h`): "Must be non-NULL and non-empty" contradicts the defensive NULL/empty handling.

One thing I explicitly checked and am fine with: `process_option()` now rejects a non-numeric/overflowing value with `ARES_EBADSTR` instead of silently truncating — a strictness improvement that's safely contained, since `ares_sysconfig_set_options()` swallows all non-`ENOMEM` returns and simply skips the malformed option. `src/tools/adig.c` still has unchecked `atoi`/`strtoul`, but that's the CLI tool and reasonably out of scope.

src/lib/include/ares_str.h
```diff
@@ -60,6 +60,21 @@ CARES_EXTERN size_t ares_strcpy(char *dest, const char *src, size_t dest_size);
 CARES_EXTERN ares_bool_t    ares_str_isnum(const char *str);
 CARES_EXTERN ares_bool_t    ares_str_isalnum(const char *str);
 
+/*! Parse a base-10 unsigned integer from a NULL-terminated string.  Unlike
+ *  atoi(), this rejects overflow, trailing garbage, and any value that does
+ *  not fit within the caller-provided maximum.
+ *
+ *  \param[in]  str  String to parse.  Must be non-NULL and non-empty.
```
bradh352: **[Nit]** "Must be non-NULL and non-empty" contradicts the implementation, which defensively tolerates both and returns `ARES_FALSE`. Suggest rewording to match, e.g.: `str  String to parse.  May be NULL or empty, in which case ARES_FALSE is returned.`

test/ares-test-misc.cc
```diff
@@ -147,6 +147,12 @@ TEST_F(DefaultChannelTest, SetServersCSV) {
             ares_set_servers_csv(channel_, "1.2.3.4 , [0102:0304:0506:0708:0910:1112:1314:1516]:53, [fe80::1]:53%iface0, 2.3.4.5"));
   EXPECT_EQ(expected, GetNameServers(channel_));
 
+  // Reject an out-of-range numeric link-local scope, keep rest.  Validation
+  // fails before the interface is looked up, so no real interface is needed.
+  EXPECT_EQ(ARES_SUCCESS,
+            ares_set_servers_csv(channel_, "1.2.3.4 , [0102:0304:0506:0708:0910:1112:1314:1516]:53, [fe80::1]:53%99999999999, 2.3.4.5"));
+  EXPECT_EQ(expected, GetNameServers(channel_));
```
bradh352: **[Low — test coverage]** This case passes even against the *unfixed* (`atoi`) code, so it doesn't actually lock in the fix. The default channel resolves the scope through the real `if_indextoname()`, which fails for a non-existent index `99999999999` no matter how `atoi` mangles the overflow — so the server is skipped identically old vs. new. The overflow-rejection in `ares_str_parse_uint()` itself is never what's being observed here.

Since `ares_str_parse_uint()` is `CARES_EXTERN` and `test/ares-test-internal.cc` already unit-tests internal string helpers, a direct test there would genuinely cover it:
```cpp
TEST_F(LibraryTest, StrParseUint) {
  unsigned int v = 0;
  EXPECT_TRUE(ares_str_parse_uint("0", 128, &v));    EXPECT_EQ(0u, v);
  EXPECT_TRUE(ares_str_parse_uint("128", 128, &v));  EXPECT_EQ(128u, v);
  EXPECT_FALSE(ares_str_parse_uint("129", 128, &v));            /* > max      */
  EXPECT_FALSE(ares_str_parse_uint("4294967296", UINT_MAX, &v));/* LP64 fit   */
  EXPECT_FALSE(ares_str_parse_uint("12x", UINT_MAX, &v));       /* trailing   */
  EXPECT_FALSE(ares_str_parse_uint("", UINT_MAX, &v));
  EXPECT_FALSE(ares_str_parse_uint(NULL, UINT_MAX, &v));
}
```
Keeping the `SetServersCSV` case is fine as an integration smoke test — this just adds the coverage that actually exercises the change.

src/lib/str/ares_str.c
jmestwa-coder: done. the helper now rejects a leading non-digit up front (via ares_isdigit) so a sign or whitespace can't slip through name_match or the options parser, and it caps max at UINT_MAX before the compare so the unsigned int assignment can't truncate. dropped the now-dead end == str check since a guaranteed leading digit means strtoul always consumes at least one char.

src/lib/include/ares_str.h
jmestwa-coder: reworded, str now documented as may be NULL or empty (ARES_FALSE returned), and added notes for the leading-garbage rejection and the UINT_MAX cap on max.

test/ares-test-misc.cc
jmestwa-coder: added LibraryTest.StrParseUint in ares-test-internal.cc, close to your suggestion, covering > max, the LP64 4294967296 case, trailing and leading garbage, the sign wrap, and empty/NULL. left the SetServersCSV case in as the integration smoke test.

jmestwa-coder:
Pushed the follow-ups from the last round:

- routed `numeric_service_to_port()` in ares_getaddrinfo.c through `ares_parse_port(service, &port, ARES_TRUE)` and removed the static helper, so that's the last hand-rolled strtoul port parser gone. empty service now returns ARES_ESERVICE instead of yielding port 0, matching the edge you noted.
- hardened `ares_str_parse_uint()`: leading non-digit rejected inside the helper, and max capped at UINT_MAX.
- added the direct `LibraryTest.StrParseUint` unit test.
- fixed the header doc.

Builds clean (CMake+Ninja, no new warnings) and the getaddrinfo/service, SetServersCSV, sortlist, options, and new StrParseUint tests all pass. Left adig.c out as you said, that's the CLI tool.

src/lib/include/ares_str.h
```diff
@@ -60,6 +60,24 @@ CARES_EXTERN size_t ares_strcpy(char *dest, const char *src, size_t dest_size);
...
+ *  \param[in]  max  Largest value considered valid.  Values above UINT_MAX
+ *                   are capped to UINT_MAX since out is an unsigned int.
+ *  \param[out] out  On success, receives the parsed value.  Untouched on
+ *                   failure.
+ *  \return ARES_TRUE if a valid in-range integer was parsed, ARES_FALSE
+ *          otherwise.
+ */
+CARES_EXTERN ares_bool_t    ares_str_parse_uint(const char *str,
+                                                unsigned long max,
+                                                unsigned int *out);
```
bradh352: **[Major — lands with the rebase]** This is the clang-format CI failure: I replicated the workflow's exact `clang-format-diff` run (clang-format 22.x) and it flags just these three lines. The hand-alignment mimics the neighboring `AlignConsecutiveDeclarations` column, but a multi-line declaration splits the alignment group, so the padding must go:

```suggestion
CARES_EXTERN ares_bool_t ares_str_parse_uint(const char *str, unsigned long max,
                                             unsigned int *out);
```

**[Nit]** Separately: `ares_parse_port()` is declared in `ares_private.h` but is now implemented on top of this helper in the same .c file — moving its declaration here would keep the pair together. Fine as a follow-up.

src/lib/ares_getaddrinfo.c
```diff
@@ -625,14 +608,14 @@ static void ares_getaddrinfo_int(ares_channel_t *channel, const char *name,
 
   if (service) {
     if (hints->ai_flags & ARES_AI_NUMERICSERV) {
-      if (!numeric_service_to_port(service, &port)) {
+      if (!ares_parse_port(service, &port, ARES_TRUE)) {
```
bradh352, line 674: **[Nit]** This is the fold-in we discussed and the `""`-to-port-0 fix is intentional, but note it's a user-visible tightening with no coverage: `""`, `" 80"`, and `"+80"` were all accepted by the old helper (strtoul tolerance; `""` proceeded with port 0) and now return `ARES_ESERVICE`. For calibration: musl rejects all three exactly as we now do; glibc treats `""` like NULL and accepts `" 80"`/`"+80"`. POSIX-defensible either way. Two optional follow-ups: a short test (nothing currently exercises any `ARES_AI_NUMERICSERV` path at all), e.g.

```cpp
TEST_F(FileChannelTest, GetAddrInfoEmptyService) {
  TempFile hostsfile("1.2.3.4 example.com");
  EnvValue with_env("CARES_HOSTS", hostsfile.filename());
  struct ares_addrinfo_hints hints = {};
  AddrInfoResult result = {};
  hints.ai_family = AF_INET;
  hints.ai_flags  = ARES_AI_ENVHOSTS | ARES_AI_NOSORT;
  ares_getaddrinfo(channel_, "example.com", "", &hints, AddrInfoCallback, &result);
  Process();
  EXPECT_TRUE(result.done_);
  EXPECT_EQ(ARES_ESERVICE, result.status_);
}
```

(plus a NUMERICSERV loop over `"+80"`, `" 80"`, `"80x"`, `"65536"`), and a one-line mention of the `""` change in the PR description, since that's the case most likely to be noticed by glibc-portable callers.

test/ares-test-misc.cc
```diff
@@ -147,6 +147,12 @@ TEST_F(DefaultChannelTest, SetServersCSV) {
             ares_set_servers_csv(channel_, "1.2.3.4 , [0102:0304:0506:0708:0910:1112:1314:1516]:53, [fe80::1]:53%iface0, 2.3.4.5"));
   EXPECT_EQ(expected, GetNameServers(channel_));
 
+  // Reject an out-of-range numeric link-local scope, keep rest.  Validation
+  // fails before the interface is looked up, so no real interface is needed.
+  EXPECT_EQ(ARES_SUCCESS,
+            ares_set_servers_csv(channel_, "1.2.3.4 , [0102:0304:0506:0708:0910:1112:1314:1516]:53, [fe80::1]:53%99999999999, 2.3.4.5"));
```
bradh352: **[Nit]** Keeping this as a smoke test is fine as discussed, but heads-up that it still passes against the pre-PR code: on LP64, `atoi("99999999999")` wraps to 1215752191 and `if_indextoname()` fails on that just as the new code's rejection does, so the entry is skipped either way. If you want the integration case to actually discriminate, `%4294967297` (2^32+1) wraps to index 1 (loopback) under old atoi — the old code would have resolved the interface and *kept* the server, failing the `expected` comparison, while the new code rejects it everywhere. Optional, since `StrParseUint` now covers the helper directly.

src/lib/ares_sysconfig_files.c
```diff
@@ -409,8 +411,11 @@ static ares_status_t process_option(ares_sysconfig_t *sysconfig,
 
   key = kv[0];
   if (num == 2) {
-    val    = kv[1];
-    valint = (unsigned int)strtoul(val, NULL, 10);
+    val = kv[1];
+    if (!ares_str_parse_uint(val, UINT_MAX, &valint)) {
```
bradh352, line 420: **[Nit]** Because the value is validated before the key is dispatched, a flag option with a bogus value flips behavior: `rotate:on` / `usevc:true` previously set the flag (the garbage `valint` was unused), now the token is dropped. Also `ndots:5x` diverges from glibc, which atoi-accepts it as 5. I verified containment is right (only `ENOMEM` is fatal in `ares_sysconfig_set_options()`, so one bad token never affects the others) and nobody legitimately writes `rotate:on`, so fine as-is — but parsing the value only for the numeric keys would confine the strictness to where it means something. A contained test pinning `ndots:junk` -> default-kept (old code: ndots=0) plus containment would lock in the real fix here:

```cpp
NameContentList badoptresolv = {
  {"/etc/resolv.conf", "nameserver 1.2.3.4\noptions ndots:junk attempts:4\n"}};
CONTAINED_TEST_F(LibraryTest, ContainerBadOptionValueInit,
                 "myhostname", "mydomainname.org", badoptresolv) {
  EXPECT_EQ(ARES_SUCCESS, ares_init(&channel));
  /* invalid ndots value is skipped (previously applied as 0); a bad option
   * must not prevent later options from applying */
  EXPECT_EQ(1, channel->ndots);
  EXPECT_EQ(4, channel->tries);
  ares_destroy(channel);
  return HasFailure();
}
```

test/ares-test-internal.cc
```diff
@@ -85,6 +85,19 @@ TEST_F(LibraryTest, StringLengthWithoutNullTerminator) {
   }
 }
 
+TEST_F(LibraryTest, StrParseUint) {
+  unsigned int v = 0;
```
bradh352: **[Nit]** Coverage gaps worth closing while you're here:
- `"4294967295"` (exactly UINT_MAX) -> TRUE. Only the over-boundary side is tested, and the accept path differs by platform (LP64: `val > max` compare; 32-bit `unsigned long`: parses to exactly ULONG_MAX *without* ERANGE), so the boundary is the interesting case.
- The header promises "untouched on failure" but nothing asserts it — and `v` starts at 0, so the first `EXPECT_EQ(0u, v)` would pass even if `*out` were never written. Start from a nonzero sentinel and assert it survives a failed call:

```cpp
  unsigned int v = 42;
  EXPECT_TRUE(ares_str_parse_uint("0", 128, &v));
  EXPECT_EQ(0u, v);
  v = 42;
  EXPECT_FALSE(ares_str_parse_uint("129", 128, &v));
  EXPECT_EQ(42u, v);
```
- `max == 0` (`"0"` -> TRUE, `"1"` -> FALSE), `"+5"` (the sign case the header doc calls out, distinct from `" 5"`), `"0x10"` (leading digit passes the guard, trailing garbage catches it — a different rejection path than `" 5"`), and NULL `out`.

Style: the two-statements-per-line + aligned `/* */` trailing-comment pattern doesn't appear elsewhere in this file (CI doesn't format-check .cc, so cosmetic only), and `UINT_MAX` currently arrives via transitive include — worth an explicit `#include <climits>`.

src/lib/ares_update_servers.c
```diff
@@ -403,7 +405,13 @@ static ares_status_t ares_sconfig_linklocal(const ares_channel_t *channel,
 
   if (ares_str_isnum(ll_iface)) {
     char ifname[IF_NAMESIZE] = "";
-    ll_scope                 = (unsigned int)atoi(ll_iface);
+
+    /* A numeric interface index can be up to 15 digits; parse with the
+     * range-checked helper so an out-of-range value is rejected rather than
+     * silently wrapped as atoi() would */
```
bradh352: **[Nit]** "up to 15 digits" is the `IF_NAMESIZE == 16` case only — c-ares's own fallback in `ares_ipv6.h` is 32, and Windows defines it as 256. The helper doesn't depend on any length bound, so suggest wording it by the actual reason:

```suggestion
    /* The interface identifier is all digits but may not fit in an
     * unsigned int; parse with the range-checked helper rather than atoi(),
     * whose behavior on overflow is undefined. */
```

src/lib/util/ares_iface_ips.c
```diff
@@ -337,8 +339,11 @@ static ares_bool_t name_match(const char *name, const char *adapter_name,
     return ARES_TRUE;
   }
 
-  if (ares_str_isnum(name) && (unsigned int)atoi(name) == ll_scope) {
-    return ARES_TRUE;
+  {
+    unsigned int scope;
```
bradh352: **[Nit]** C89-legal, but a bare scoping block has essentially no precedent in src/lib — convention is declarations at function top:

```c
static ares_bool_t name_match(const char *name, const char *adapter_name,
                              unsigned int ll_scope)
{
  unsigned int scope;
  ...
  if (ares_str_parse_uint(name, UINT_MAX, &scope) && scope == ll_scope) {
    return ARES_TRUE;
  }
```

src/lib/str/ares_str.c
```diff
@@ -127,23 +128,45 @@ ares_bool_t ares_str_isnum(const char *str)
...
+  /* Require a leading digit so strtoul()'s tolerance of a sign or leading
+   * whitespace (e.g. "-1" wrapping to UINT_MAX) can't slip through here. This
+   * also rejects NULL and empty strings. */
+  if (str == NULL || out == NULL || !ares_isdigit(*str)) {
     return ARES_FALSE;
   }
 
+  /* out is unsigned int, so a max above UINT_MAX would let strtoul's result
+   * truncate silently on assignment.  Cap it. */
+  if (max > UINT_MAX) {
```
bradh352, line 146: **[Nit — FYI only]** On LLP64/ILP32 (all Windows, 32-bit Unix) `unsigned long` is 32-bit, so this comparison is always false. I checked whether that trips any warning gate we actually run: GCC `-Wextra` doesn't include this case (verified 12/14/15 at -m32), MSVC `/W4 /w14296` is clean, and clang only flags it under an explicit `-Wtype-limits`. So: fine as-is. If you ever want it warning-proof for downstream `-Wtype-limits` users, `#if ULONG_MAX > UINT_MAX` around the cap is the portable spelling.

bradh352 (approved):
## Review summary

Re-reviewed the consolidated version in depth (temp checkout @ 3ee58e7; CMake+Ninja build clean under the full strict-warning set; targeted tests plus the full offline suite pass; swept src/lib and confirmed **zero** remaining in-scope unvalidated string-to-int conversions, so the consolidation goal is genuinely met; no API/ABI impact; the strtoul usage is the canonical POSIX pattern and holds on LP64/LLP64/ILP32 including the exact-UINT_MAX boundary on both models).

### Major — rebase needed
main has moved and #1186 conflicts with this branch in `ares_getaddrinfo.c` (it added `ares_append_nulladdr()` immediately after the `numeric_service_to_port()` this PR deletes). Resolution is trivial — drop the helper, keep `ares_append_nulladdr()` — but as-is the PR doesn't merge. While rebasing, note the new declaration in `ares_str.h` fails the clang-format CI gate (exact fix left inline); run `git clang-format` before pushing and both land in one push.

### Everything else is nits (inline)
Behavior changes I verified and am fine with: empty/`" 80"`/`"+80"` getaddrinfo service strings now fail with `ARES_ESERVICE` (musl-aligned, POSIX-defensible, stricter than glibc — worth a line in the PR description since it's user-visible); resolv.conf flag options with bogus values (`rotate:on`) are now dropped rather than applied (contained per-token — verified only `ENOMEM` is fatal in `ares_sysconfig_set_options()`); `dns://` URI ports like `:+53` are now rejected (RFC 3986 `port = *DIGIT`, so more correct).

Interop check against today's glibc/musl masters: the helper's discipline (leading digit, full consumption, range check) exactly matches glibc's own `__inet6_scopeid_pton` numeric branch; `ndots:junk` handling now matches musl (skip, keep default) where it previously accidentally matched glibc (apply 0); no known resolv.conf generator emits anything the new code rejects. RFC 4007 zone-id handling conforms; the numeric-before-name precedence difference vs glibc predates this PR and is unchanged.

Test nits are inline (CSV discriminator value, StrParseUint boundary/contract cases, optional service-string coverage). One more: #1157's sortlist boundary cases (`/129`, IPv4 `/33`, and `/4294967296` — which old code accepted as `/0` on LP64) didn't make it into the consolidation; nice-to-have additions to `SetSortlistFailures` while you're in there.

Maintainer note (no action needed from you): open #1160 is a strict subset of this PR and should be closed in favor of it.

The security framing in the description is accurate and appropriately modest: local config/API inputs only, `atoi()` overflow UB is real but not remotely reachable — hardening plus genuine minor bug fixes.

jmestwa-coder:
Rebased on main and pushed. The conflict was exactly as you called it: dropped `numeric_service_to_port()`, kept `ares_append_nulladdr()`. Ran `git clang-format` before pushing, so the `ares_str.h` declaration now matches the CI formatter.

Folded in the inline nits:

- `ares_str_parse_uint` test: added the exact-`UINT_MAX` accept boundary, a nonzero sentinel to prove `*out` is untouched on failure, and the `+5` / `0x10` / `max == 0` cases, plus an explicit `<climits>`.
- CSV test now uses `%4294967297` instead of `%99999999999` so it actually discriminates old vs new (2^32+1 wraps to index 1 under the old atoi).
- moved the `name_match` scope var to the function top instead of the bare block, and reworded the link-local comment to key off "may not fit an unsigned int" rather than the digit count.
- added a line to the PR description about the user-visible empty/` 80`/`+80` service now returning `ARES_ESERVICE`.

Left the `process_option` value-parsing restructure and moving the `ares_parse_port` declaration as-is since you marked those fine/follow-up. Build is clean and the StrParseUint/SetServersCSV/sortlist/options tests pass.

PR #1159, opened 2026-05-25
skip name compression offsets beyond the 14-bit pointer limit

jmestwa-coder:
ares_nameoffset_create() records every written name as a compression target, but that offset is later emitted as a 14-bit pointer (0xC000 | (idx & 0x3FFF)). Once a message serializes past 16383 bytes, a repeated owner or target name gets encoded as a pointer to idx & 0x3FFF rather than its real position, so ares_dns_write() produces a message that no longer parses. Skip recording offsets that can't fit 14 bits; those names get written uncompressed instead.


jmestwa-coder:
gentle ping

bradh352 (commented):
Thanks for this — the analysis in the description is accurate and the fix is correct and complete. Confirmed:

- `ares_nameoffset_create()` (record site) and the pointer emission `0xC000 | (off->idx & 0x3FFF)` are the **only** record/emit sites, both in this file, so gating the record is sufficient — there is no second path that could still truncate.
- Boundary `idx > 0x3FFF` is exactly right: a name only needs its *start* offset to fit 14 bits, and `0x3FFF` (16383) itself stays addressable.
- Correctly cites RFC 1035 §4.1.4. This is a **write-only** change with no parser changes, so there is no interoperability risk — it only makes emitted messages more compliant, and never rejects a peer's input.
- No memory-safety concern before or after: the old truncated pointer still landed in-range (0…0x3FFF, a valid earlier buffer position), so it produced a corrupt name / parse failure rather than an out-of-bounds access.
- No linked issue, and no overlapping PRs touching the write/compression path.

**Requested before merge:**

1. **(Medium) Add a test.** The new branch is currently uncovered and Coveralls (`.github/workflows/coveralls.yml`) will flag it as a regression. Since this is real logic (not defensive/OOM), a test is preferable to an `LCOV_EXCL_LINE`. See the inline suggestion for a write→parse round-trip sketch.

**Minor / optional:**

2. **(Low)** The `& 0x3FFF` mask in `ares_dns_name_write()` (the `0xC000 | (off->idx & 0x3FFF)` line) is now provably a no-op, since `idx` is capped to 14 bits at record time. Fine to keep as defensive code, but a short comment noting the invariant is enforced in `ares_nameoffset_create()` would help the next reader (I couldn't attach this inline as that line is outside the diff).

3. **(Nit)** See inline note on the early-return semantics of `ares_nameoffset_create()`.

**Out-of-scope FYI (not for this PR):** in `ares_dns_write_buf_tcp()`, the 2-byte length prefix means `pos = ares_buf_len(buf)` is buffer-absolute rather than message-relative (offset by +2), and that same base feeds both the pointer emission and this new `idx > 0x3FFF` check. It is harmless today because the TCP write path only serializes single-name queries where compression never fires, but it's worth keeping in mind if the TCP writer ever serializes multi-name messages.

src/lib/record/ares_dns_name.c
```diff
@@ -52,6 +52,14 @@ static ares_status_t ares_nameoffset_create(ares_llist_t **list,
     return ARES_EFORMERR; /* LCOV_EXCL_LINE: DefensiveCoding */
   }
 
+  /* Don't record a name as a compression target if it can't be referenced by
+   * a pointer.  DNS name compression offsets are only 14 bits (RFC 1035
+   * 4.1.4), so a name positioned beyond 0x3FFF would have its offset
+   * truncated when written, producing a pointer to the wrong location. */
+  if (idx > 0x3FFF) {
```
bradh352, line 71: **Medium: needs test coverage.** This branch isn't exercised by any existing test, and c-ares gates PRs on Coveralls, so it will show as a coverage regression. Since this is real logic (not defensive/OOM), the right fix is a test rather than an `LCOV_EXCL_LINE`. A round-trip test covers the branch and guards against regression:

```cpp
TEST_F(LibraryTest, DNSNameCompressionBeyond14Bit) {
  ares_dns_record_t *dnsrec = NULL;
  /* Build a message that pushes a *repeated* compressible name past
   * offset 0x3FFF (e.g. many large TXT RRs), reusing the same owner
   * name both before and after byte 16383. */
  ...
  unsigned char *buf = NULL; size_t buf_len = 0;
  EXPECT_EQ(ARES_SUCCESS, ares_dns_write(dnsrec, &buf, &buf_len));
  EXPECT_GT(buf_len, 0x3FFFU);

  ares_dns_record_t *parsed = NULL;
  /* Without this fix the >16383 name would emit a truncated pointer and
   * this parse would fail / mis-parse. */
  EXPECT_EQ(ARES_SUCCESS, ares_dns_parse(buf, buf_len, 0, &parsed));
  ares_dns_record_destroy(parsed);
  ares_free_string(buf);
}
```

Key point: the compressible name must appear **twice** — once before and once after byte 16383 — so that without the fix a pointer would actually be emitted. That's what makes it a true regression test.

bradh352, line 71: **Nit:** `ares_nameoffset_create()` now returns `ARES_SUCCESS` without creating anything — a "create" that no-ops. The comment you added explains the *why* well, so this is acceptable. Purely optional alternative: hoist the check into the caller next to where `pos` is computed and the other record-conditions live (in `ares_dns_name_write()`), keeping `create()`'s contract "create or fail." Current placement (single choke point) is also perfectly defensible — your call.

bradh352:
Wanted to check on this... will be cutting a release in a few days.

src/lib/record/ares_dns_name.c
jmestwa-coder: Added a write/parse round-trip test (DNSNameCompression14Bit). It pads past 16383 with TXT RRs, then repeats an owner name that first lands beyond 0x3FFF, so the second occurrence would emit a truncated pointer without the fix. Confirmed it fails (EBADNAME) with the branch reverted and passes with it in.

jmestwa-coder: Kept the single choke point in create() with the explanatory comment, since it also covers the TCP write path without duplicating the check. Can hoist it into the caller if you'd rather keep create()'s contract strict.

jmestwa-coder:
Pushed the test plus the invariant comment on the 0x3FFF mask. Should be good to go for the release.

test/ares-test-internal.cc
```diff
@@ -1398,6 +1398,63 @@ TEST_F(LibraryTest, DNSRecord) {
...
+  for (size_t i = 0; i < 2; i++) {
+    EXPECT_EQ(ARES_SUCCESS,
+      ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER,
+        "repeat.example.com", ARES_REC_TYPE_TXT, ARES_CLASS_IN, 0));
+    EXPECT_EQ(ARES_SUCCESS,
+      ares_dns_rr_add_abin(rr, ARES_RR_TXT_DATA, txt.data(), txt.size()));
+  }
+
+  EXPECT_EQ(ARES_SUCCESS, ares_dns_write(dnsrec, &msg, &msglen));
+  EXPECT_GT(msglen, 0x3FFFU);
```
bradh352, line 1439: **[Nit]** Nice construction — this incidentally also covers "late name pointing to an early target": the ~18 `example.com` owners past offset 16383 are still emitted as 2-byte pointers to offset 12. One small gap: every assertion here would still pass if compression got disabled *entirely* (the fully-uncompressed message is ~22.9KB — also > 0x3FFF, parses fine, same ancount and names). Pinning the size closes that; the layout is deterministic given the API:

```cpp
  /* 12 hdr + 17 question + 80*268 compressed TXT + 2*275 repeat TXT.
   * An exact size also verifies compression remains active for names
   * below the limit. */
  EXPECT_EQ(22019U, msglen);
```

(or `EXPECT_LT(msglen, 22300U);` if you'd rather not hard-pin). The exact `idx == 0x3FFF` boundary isn't pinned either, but that needs byte-exact padding and all four major implementations agree on the boundary — fine to skip.

test/ares-test-internal.cc
```diff
@@ -1398,6 +1398,63 @@ TEST_F(LibraryTest, DNSRecord) {
...
+    EXPECT_EQ(ARES_SUCCESS,
+      ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER, "example.com",
+        ARES_REC_TYPE_TXT, ARES_CLASS_IN, 0));
+    EXPECT_EQ(ARES_SUCCESS,
+      ares_dns_rr_add_abin(rr, ARES_RR_TXT_DATA, txt.data(), txt.size()));
+  }
+
+  /* Past byte 16383, add a repeated owner name.  Its first occurrence records a
+   * compression target at an offset that doesn't fit in 14 bits; the second
+   * occurrence must be written uncompressed rather than as a pointer to
```
bradh352, line 1428: **[Nit]** "must be written uncompressed" isn't quite what happens post-fix: the second occurrence still *suffix-compresses* — it's emitted as `\6repeat` plus a pointer to the in-range `example.com` target at offset 12. Only the full `repeat.example.com` target went unrecorded. Suggest: "...the second occurrence must not be emitted as a pointer to (offset & 0x3FFF); it instead compresses only against the in-range example.com target."

bradh352 (approved):
## Review summary

Deep re-review of the final state (temp checkout; CMake+Ninja build clean; discriminator verified — with the fix reverted the new test fails with exactly `ARES_EBADNAME`, and I traced why: the truncated pointer lands mid-TXT-data where 0x61 bytes parse as runaway label lengths). Everything I asked for in the previous rounds is in and done right. Approving; the two inline notes are nits that can ride along with any future push.

**Note:** I've merged current main into this branch myself (maintainer edit) to resolve the test-file conflict that accumulated since your base — both your `DNSNameCompression14Bit` and main's `DNSRecordNSEC3EmptyFields` are kept, and I verified the merged state builds and the relevant tests pass. Nothing for you to do there.

Additional verification this round:
- **Boundary and approach match industry practice byte-for-byte** — BIND9 (`insert_label`: `coff >= 0x4000` → skip), NSD (`MAX_COMPRESSION_OFFSET` 16383), miekg/dns (`maxCompressionOffset`), and PowerDNS all silently skip recording rather than erroring, with the identical 16383-inclusive boundary and the same skip-don't-error strategy.
- **No round-trip break with #1190**: the writer and parser gate RDATA compression through the same `ares_dns_rec_allow_name_comp()`, so c-ares output always re-parses under the new RFC 3597 strictness.
- **Reachability is stronger than the description says**: `ares_dns_record_duplicate()` is implemented as write→parse, and `ares_qcache_insert()` duplicates every response it caches — so pre-fix, a legitimate 16K–64K TCP response with a repeated name past offset 16383 could be corrupted in-library at cache-insert time, not just via external `ares_dns_write()` users. I'll note that in the merge commit.
- Zero API/ABI surface (file-local statics only), C89-clean, clang-format-clean on changed lines, no new warnings, and `ares_dns_name.c` is confirmed to be the sole compression-pointer emitter in the tree.

**Adjacent FYIs, out of scope, no action needed:** the write path has three sibling u16-truncation sites that survive this PR — the rdlength backfill (`ares_dns_write.c:1083`, reachable with >64KB of TXT abin data in one RR), the OPT/SVCB/HTTPS option-value length (`:667/:756/:807`, writes a wrapped length yet appends all the bytes), and the header counts (`:108-130`, needs >65535 RRs). The TCP length prefix is already guarded (`:1172`). Those are follow-up-issue material — one `> 65535` guard each — possibly coordinated with #1168, which guards the allocation side of the same setters. Also pre-existing and unchanged: on the TCP write path `pos` is buffer-absolute rather than message-relative; your check is conservative in the safe direction there (can only over-skip compression, never under-skip), and that path only writes single-name queries today.

test/ares-test-internal.cc
jmestwa-coder: This merged before I could fold these in, so I've put both nits in #1228. Went with the exact pin since the layout is deterministic; confirmed 22019 locally.

test/ares-test-internal.cc
jmestwa-coder: Right, only the full-name target goes unrecorded. Reworded in #1228.

PR #1161, opened 2026-05-30
discard oversized UDP datagrams instead of truncating the length frame

jmestwa-coder:
Defense-in-depth hardening for the UDP receive path.

read_conn_packets() prefixes each buffered UDP datagram with a 2-byte length so read_answers() can later split the buffer back into messages. The read buffer handed to recvfrom() comes from ares_buf_append_start(), which grows in powers of two and so can be larger than 65535 bytes, while count is written back through (unsigned short). If a datagram larger than 65535 ever landed there, count & 0xFFFF would be stored and the framing of everything buffered after it would desync.

This isn't reachable over standard UDP: max payload is 65507 (IPv4) / 65527 (IPv6), both below 65535. The only vector is IPv6 jumbograms (RFC 2675), which DNS never uses and which would still have to pass the source-address check in ares_conn_read(). So this is purely hardening, not a fix for an observed bug on real networks.

The guard rolls the datagram back to the iteration start and stops reading, matching the existing short-read rollback just above. Only ares_buf_* is used.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -496,7 +496,10 @@ static ares_status_t read_conn_packets(ares_conn_t *conn,
        * return more than the 2-byte length prefix is able to represent.  A
        * datagram that large can't be a valid DNS message, and writing the
        * truncated (unsigned short)count would desync the framing of everything
-       * buffered after it, so discard it. */
+       * buffered after it, so discard it.  Standard UDP can't actually deliver
+       * a payload this large (max is 65507 IPv4 / 65527 IPv6); the only vector
+       * is IPv6 jumbograms (RFC 2675), which DNS never uses.  This is purely
+       * defense-in-depth. */
       if (count > 65535) {
         ares_buf_set_length(conn->in_buf, start_len);
         break;
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -492,6 +492,15 @@ static ares_status_t read_conn_packets(ares_conn_t *conn,
 
     /* If UDP, overwrite length */
     if (!(conn->flags & ARES_CONN_FLAG_TCP)) {
+      /* The read buffer is grown in powers of two, so a single recvfrom() can
+       * return more than the 2-byte length prefix is able to represent.  A
+       * datagram that large can't be a valid DNS message, and writing the
+       * truncated (unsigned short)count would desync the framing of everything
+       * buffered after it, so discard it. */
+      if (count > 65535) {
+        ares_buf_set_length(conn->in_buf, start_len);
+        break;
+      }
       len = ares_buf_len(conn->in_buf);
       ares_buf_set_length(conn->in_buf, start_len);
       ares_buf_append_be16(conn->in_buf, (unsigned short)count);


src/lib/ares_process.c
```diff
@@ -492,6 +492,15 @@ static ares_status_t read_conn_packets(ares_conn_t *conn,
 
     /* If UDP, overwrite length */
     if (!(conn->flags & ARES_CONN_FLAG_TCP)) {
+      /* The read buffer is grown in powers of two, so a single recvfrom() can
+       * return more than the 2-byte length prefix is able to represent.  A
+       * datagram that large can't be a valid DNS message, and writing the
+       * truncated (unsigned short)count would desync the framing of everything
+       * buffered after it, so discard it. */
+      if (count > 65535) {
```
bradh352, line 503: **[Medium — reachability, accepting as hardening]** The logic gap is genuine: `ares_buf_append_start()` reports the real power-of-two capacity, so the `len` handed to `recvfrom()` can exceed 65535, and `count` is written through `(unsigned short)` below — a `count > 65535` would desync framing for everything buffered after it. The rollback to `start_len` + `break` here is correct and mirrors the short-read handling above.

For the record, though: this isn't reachable over standard UDP — max payload is 65507 (IPv4) / 65527 (IPv6), both < 65535. The only vector is IPv6 jumbograms (RFC 2675), which DNS never uses. We're happy to take it as cheap defense-in-depth.

**Nits (optional):**
- Consider noting in the comment that standard UDP can't deliver this and jumbograms are the only vector, so future readers don't go hunting for a live bug.
- `65535` now appears three times in this function (the initial `len = 65535`, this check, and the `(unsigned short)` cast). A shared constant (e.g. `ARES_UDP_MAX_MSG_LEN`) would tie them together. Matches the existing literal style, so take it or leave it.

bradh352 (commented):
Thanks for the patch — accepting this as hardening. The change is correct, minimal, consistent with the existing short-read rollback just above, uses only `ares_buf_*`, and has no panic risk. No duplicate/overlapping PRs and no associated tracking issue.

**One request (description):** please reword the PR description so it doesn't read as an observed framing bug on ordinary networks. The truncation is not reachable over standard UDP — max UDP payload is 65507 (IPv4) / 65527 (IPv6), both below 65535. The only vector is IPv6 jumbograms (RFC 2675), which DNS never uses (and which would still have to pass the source-address check in `ares_conn_read()`). Framing it as defense-in-depth hardening is accurate; the current wording implies everything buffered after gets mis-framed on normal networks, which can't happen.

**Test coverage (Low, optional):** none added. A real oversized datagram can't be produced, but the branch *is* unit-testable by overriding the socket layer with a custom `arecvfrom` that returns `65536` (no real payload needed — `count` alone trips the guard), in `MockUDPChannelTest` (`test/ares-test-mock.cc`), following the `noop_recvfrom` injection pattern in `test/ares-test-mock-et.cc`. Cheap to add and would lock in the behavior, but not a blocker. Existing tests couldn't have caught this — the mock server never returns >65535.

Nothing required for correctness.

src/lib/ares_process.c
jmestwa-coder: Reworded the description to make clear this is defense-in-depth, not a live bug on real networks, and added the jumbogram/RFC 2675 note to the comment so nobody goes hunting for a reachable path. Left the 65535 literals as-is to match the surrounding style. Happy to add the mock arecvfrom test that returns 65536 if you'd like it locked in, though as you note it can't distinguish much since both paths just end in a timeout.

bradh352 (approved):
Approving — both requested items landed: the description now reads clearly as defense-in-depth (not an observed bug on real networks), and the comment documents the standard-UDP limits and the jumbogram/RFC 2675 vector so nobody hunts for a live bug.

Re-verified the guard against the `ares_buf` internals:

- **Rollback is exact.** At the guard the buffer holds `start_len` + 2-byte placeholder + `count` committed bytes; `ares_buf_set_length(conn->in_buf, start_len)` truncates back to the pre-iteration length, discarding only the oversized datagram while preserving everything framed in earlier loop iterations. Mirrors the short-read rollback just above.
- **No panic/UB.** The ignored `ares_buf_set_length` return can only fail on NULL/const buf or out-of-range len, neither possible here (`start_len` is a prior valid data_len) — consistent with the existing overwrite block.
- **`break` is correct.** `err == ARES_CONN_ERR_SUCCESS` on this path, so the connection is preserved and the function returns `ARES_SUCCESS`; the datagram was already drained from the socket by recvfrom, so no busy-loop, and any remaining queued datagrams are picked up on the next event-loop pass.
- **Threshold correct.** `> 65535` (strict) passes `count == 65535` through losslessly (`(unsigned short)0xFFFF`) and drops only unrepresentable values.
- C89-clean, `ares_buf_*` only, no API/ABI impact, no dead code. Leaving the `65535` literal as-is to match surrounding style is fine.
- No overlapping PRs (#1113/#1118 touch the distant read_answers/requeue region only), no tracking issue needed.

**Tests:** none added, and that's the right call here. The only value the guard adds is preventing framing desync of a *following* datagram; a single synthetic oversized datagram can't distinguish guarded from unguarded (both end in a timeout, since the unguarded build writes a 0-length prefix that just fails to parse). A meaningful test would need a stateful `arecvfrom` that returns the oversized count once (with the correct source address, or `ares_conn_read` drops it first) followed by a valid reply — disproportionate scaffolding for a branch that's unreachable over real UDP and whose unguarded failure mode is a graceful parse error, not corruption. Not worth locking in.

PR #1162, opened 2026-06-02
validate numeric interface scope in name_match with strtoul

jmestwa-coder:
name_match() compares interface ids via (unsigned int)atoi(name), so an oversized token like 4294967298 truncates to a valid index and matches the wrong adapter.

### DIFF
--- a/src/lib/util/ares_iface_ips.c
+++ b/src/lib/util/ares_iface_ips.c
@@ -337,8 +337,15 @@ static ares_bool_t name_match(const char *name, const char *adapter_name,
     return ARES_TRUE;
   }
 
-  if (ares_str_isnum(name) && (unsigned int)atoi(name) == ll_scope) {
-    return ARES_TRUE;
+  if (ares_str_isnum(name)) {
+    char         *endptr = NULL;
+    unsigned long idx;
+
+    errno = 0;
+    idx   = strtoul(name, &endptr, 10);
+    if (errno != ERANGE && *endptr == '\0' && idx == (unsigned long)ll_scope) {
+      return ARES_TRUE;
+    }
   }
 
   return ARES_FALSE;


jmestwa-coder:
any update?

src/lib/util/ares_iface_ips.c
```diff
@@ -337,8 +337,15 @@ static ares_bool_t name_match(const char *name, const char *adapter_name,
...
 
-  if (ares_str_isnum(name) && (unsigned int)atoi(name) == ll_scope) {
-    return ARES_TRUE;
+  if (ares_str_isnum(name)) {
+    char         *endptr = NULL;
+    unsigned long idx;
+
+    errno = 0;
+    idx   = strtoul(name, &endptr, 10);
+    if (errno != ERANGE && *endptr == '\0' && idx == (unsigned long)ll_scope) {
```
bradh352, line 346: **Nit:** The `*endptr == '\0'` check (and an implicit empty-parse case) is dead here — the enclosing `ares_str_isnum(name)` guard already guarantees `name` is non-empty and all-digits, so `strtoul` always consumes the whole string and `*endptr` is always `'\0'`. The only behavioral change vs. the old code is the `errno != ERANGE` overflow guard, which is the meaningful part. Harmless to keep for copy-paste safety, just noting it.

bradh352, line 346: **Medium (consistency/robustness):** This is correct here because the block is Windows-only, where `unsigned long` is 32-bit, so any value exceeding a valid `unsigned int` index trips `ERANGE`. But sibling PR #1158 guards the *identical* conversion with an explicit fit-check `scope != (unsigned long)(unsigned int)scope`, which stays correct regardless of `unsigned long` width. Recommend matching that idiom (or better, sharing a helper — see summary) so the two fixes are consistent and this one remains correct if ever reused off-Windows.

bradh352 (commented):
**Overall:** The change is correct, portable, and runtime-safe — it removes real signed-overflow UB from `atoi` and correctly rejects oversized tokens (on Windows `unsigned long` is 32-bit, so anything exceeding a valid `unsigned int` index trips `ERANGE`). I'd accept the *intent*, but recommend considering consolidation before merge. Details below.

**[Major — process, not a code bug] Consolidate the three sibling PRs behind one shared helper.** #1157, #1158, and this PR all address the same `atoi`-truncation bug class, and each open-codes `strtoul` differently: #1157 has no `errno`/`ERANGE`/`endptr` check at all, #1158 has a full check plus an explicit `scope != (unsigned long)(unsigned int)scope` fit-test (and adds `#include <errno.h>`), and this one is in between. That divergence is exactly the fragmentation #1097 ("Replace atoi-based port parsing with validated helper", which added `ares_parse_port`) was meant to end. Suggest a single general helper in `ares_str.c`, e.g.:

```c
ares_bool_t ares_parse_uint(const char *str, unsigned int *out)
{
  char         *endptr = NULL;
  unsigned long val;

  if (str == NULL || out == NULL || *str == 0) {
    return ARES_FALSE;
  }
  errno = 0;
  val   = strtoul(str, &endptr, 10);
  if (errno == ERANGE || endptr == str || *endptr != '\0' || val > UINT_MAX) {
    return ARES_FALSE;
  }
  *out = (unsigned int)val;
  return ARES_TRUE;
}
```

then `name_match` collapses to:

```c
  unsigned int idx;
  if (ares_parse_uint(name, &idx) && idx == ll_scope) {
    return ARES_TRUE;
  }
```

This makes all three fixes uniform, portable, and unit-testable cross-platform (see tests note).

**[Medium — scope] The failure mode described here also exists on the producer side.** The description says an oversized token "truncates to a valid index and matches the wrong adapter." The identical `(unsigned int)atoi()` truncation lives at `src/lib/ares_update_servers.c:406` (`ares_sconfig_linklocal`), which *sets* `ll_scope`. This PR only fixes the enumeration/consumer side. (#1158 covers the producer side separately — reinforcing the consolidation point.)

**[Low — severity/justification] Real-world impact is low, but the fix is legitimate.** The triggering input is config-controlled (a numeric zone-id like `fe80::1%4294967298`), not remote network data, and the worst case is selecting a wrong *local* adapter — no memory-safety issue, and no host actually has interface index 4294967298. However, the old `atoi` path is genuine signed-overflow UB, so replacing it is worthwhile hygiene consistent with the #1097 campaign.

**[Low — tests] No tests, but a targeted test isn't practical as written.** `name_match` is static, Windows-only, and reachable only through un-mockable `GetAdaptersAddresses()` enumeration; the existing `IfaceIPs` test (`test/ares-test-internal.cc:308`) calls with `name == NULL` and never reaches this branch. `ares_parse_port` (#1097) was also merged without a dedicated test, so missing tests shouldn't block. Extracting the shared helper above would make all three fixes unit-testable cross-platform — another point for consolidation.

**[Nit] Minor inconsistency with siblings:** #1158 adds an explicit `#include <errno.h>`; this PR relies on the transitive include via `ares_setup.h` (fine, and matches `ares_parse_port`). A shared helper removes the question entirely.

**RFC/interop:** No concern — this is RFC 4007 scoped-address zone-id parsing, not DNS wire parsing. Tightening only rejects numerically-impossible interface indices, with zero interoperability impact on third-party implementations.

jmestwa-coder:
Consolidating the atoi->strtoul series into #1157 per bradh352's note. The name_match scope change now goes through the shared ares_str_to_num() helper there.

PR #1163, opened 2026-06-03
ci: pin GitHub Actions to full commit SHAs

XananasX7:
## Summary

Several workflow files use mutable version tags/refs for GitHub Actions, which is a supply chain risk.

### Actions pinned

| Action | Was | Now |
|--------|-----|-----|
| `awalsh128/cache-apt-pkgs-action` | `@latest` | `@acb598e5ddbc6f68a970c5da0688d2f3a9f04d05` (v1.6.0) |
| `actions/checkout` | `@v4` | `@34e114876b0b11c390a56381ad16ebd13914f8d5` |
| `actions/upload-artifact` | `@v4` | `@ea165f8d65b6e75b540449e92b4886f43607fa02` |
| `actions/download-artifact` | `@v4` | `@d3f86a106a0bac45b974a628896c90dbdf5c8093` |
| `softprops/action-gh-release` | `@v2` | `@3bb12739c298aeb8a4eeaf626c5b8d85266b0e65` |

### Why

Mutable refs like `@latest` and `@v4` can change without notice. Pinning to a full commit SHA makes the exact version immutable and auditable, preventing supply chain attacks from a compromised or maliciously updated action.

Recommended by [GitHub's security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) and [OpenSSF Scorecard](https://securityscorecards.dev/).

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base"
+  ],
+  "packageRules": [
+    {
+      "matchManagers": ["github-actions"],
+      "pinDigests": true,
+      "automerge": false,
+      "labels": ["ci", "dependencies"]
+    }
+  ]
+}
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -24,12 +24,12 @@ jobs:
     name: "build"
     steps:
       - name: Install packages
-        uses: awalsh128/cache-apt-pkgs-action@latest
+        uses: awalsh128/cache-apt-pkgs-action@acb598e5ddbc6f68a970c5da0688d2f3a9f04d05 # v1.6.0
         with:
           packages: autoconf automake libtool g++ libgmock-dev pkg-config gdb
           version: 1.0
       - name: Checkout c-ares
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Get Tag version
         id: version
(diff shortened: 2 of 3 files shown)


dfandrich:
A one-time pinning isn't very helpful. This is better done by having something
like dependabot or renovate periodically update the hashes, especially since
many of these labels expect to be run on the latest release.

XananasX7:
That's a fair point @dfandrich — a one-time pin without automated updates to follow means the hashes will go stale. I can add a `renovate.json` or `.github/renovate.json` config alongside this PR that enables Renovate's GitHub Actions datasource to keep the pinned SHAs updated automatically. Alternatively, a Dependabot `ecosystem: github-actions` config would do the same. Would either of those work for this project?

XananasX7:
Fair point — added `.github/renovate.json` in the latest commit. It configures Renovate's GitHub Actions datasource with `pinDigests: true`, so any pinned SHAs will be kept current automatically via PRs from the Renovate bot whenever the underlying action releases a new version. That way the pin stays meaningful over time rather than going stale.

XananasX7:
Friendly ping — happy to address any review comments. Thank you!

XananasX7:
Hey @dfandrich — confirming the `renovate.json` was added in the latest commit to address exactly that concern. It enables Renovate's GitHub Actions datasource with `pinDigests: true`, so the SHAs will be kept current automatically whenever the underlying actions release new versions. That way the pin stays meaningful over time rather than going stale after the first few months. Let me know if the config looks good!

bradh352:
dependabot has been deployed to this repository, closing

PR #1164, opened 2026-06-05
enforce 255 octet name limit when parsing dns names

jmestwa-coder:
ares_dns_name_parse had no upper bound on the assembled name length:

- the strictly-decreasing pointer rule stops loops but not unbounded expansion
- it accepted names past the RFC 1035 255 octet limit
- a ~64KB reply with one long name plus answer RRs that are just a pointer back to it re-expands that name per RR (~1.2s to parse a single packet here)
- a name built purely from compression pointers adds no label bytes, so a length cap alone never bounds it: a ~65KB reply (an ~8k descending pointer chain plus RRs whose name points at the top of it) still parses in ~0.9s and returns SUCCESS, and every RR is walked with no early exit

Fixes:

- cap the running presentation length at 255 in the parse loop and return ARES_EBADNAME past it, matching the existing check in ares_split_dns_name on the write side
- cap compression-pointer indirections per name at 128 (a <=255 octet name has <=128 labels), which bounds the pointer-only case the length cap misses
- factor the 255 limit into a shared constant used by the read cap and both write-side checks

Tests (test/ares-test-parse.cc):

- ParseRejectsOverlongName / ParseRejectsOverlongNameViaPointers (literal labels, and length accumulated across a backward pointer)
- ParseRejectsPointerChain (pure pointer chain -> ARES_EBADNAME)
- ParseAcceptsMaxLengthName / ParseRejectsName256 (255 accepted, 256 rejected)

Duplicate #1114 is already closed; keeping this one.

### DIFF
--- a/src/lib/record/ares_dns_name.c
+++ b/src/lib/record/ares_dns_name.c
@@ -543,6 +543,7 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
   ares_status_t status;
   ares_buf_t   *namebuf     = NULL;
   size_t        label_start = ares_buf_get_position(buf);
+  size_t        name_len    = 0;
 
   if (buf == NULL) {
     return ARES_EFORMERR;
@@ -636,6 +637,19 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
     }
 
     /* New label */
+
+    /* RFC 1035 3.1: a name is limited to 255 octets.  Enforce it during
+     * decompression so a chain of pointers can't expand a single name without
+     * bound.  Track the unescaped length (label data plus the separator that
+     * precedes each label after the first), matching ares_split_dns_name(). */
+    if (name_len) {
+      name_len++;
+    }
+    name_len += c;
+    if (name_len > 255) {
+      status = ARES_EBADNAME;
+      goto fail;
+    }
 
     /* Labels are separated by periods */
     if (ares_buf_len(namebuf) != 0 && name != NULL) {
--- a/src/lib/record/ares_dns_name.c
+++ b/src/lib/record/ares_dns_name.c
@@ -25,6 +25,18 @@
  */
 #include "ares_private.h"
 
+/* RFC 1035 3.1 limits a name to 255 octets.  We track presentation length
+ * (label octets plus one separator before each label after the first), which is
+ * up to ~2 octets looser than the strict wire limit and so never rejects a
+ * compliant name.  Shared by the read and write paths. */
+#define ARES_MAX_NAME_PRESENTATION_LEN 255
+
+/* A name of <= 255 octets holds at most 128 labels, so it can never legitimately
+ * require more than that many compression-pointer jumps.  Bounding the number of
+ * indirections stops a name built purely from pointers (which never adds label
+ * bytes, so the length cap never fires) from being walked without limit. */
+#define ARES_MAX_INDIRS 128
+
 typedef struct {
   char  *name;
   size_t name_len;
@@ -48,7 +60,7 @@ static ares_status_t ares_nameoffset_create(ares_llist_t **list,
   ares_nameoffset_t *off = NULL;
 
   if (list == NULL || name == NULL || ares_strlen(name) == 0 ||
-      ares_strlen(name) > 255) {
+      ares_strlen(name) > ARES_MAX_NAME_PRESENTATION_LEN) {
     return ARES_EFORMERR; /* LCOV_EXCL_LINE: DefensiveCoding */
   }
 
@@ -340,7 +352,8 @@ static ares_status_t ares_split_dns_name(ares_array_t *labels,
   }
 
   /* Can't exceed maximum (unescaped) length */
-  if (ares_array_len(labels) && total_len + ares_array_len(labels) - 1 > 255) {
+  if (ares_array_len(labels) &&
+      total_len + ares_array_len(labels) - 1 > ARES_MAX_NAME_PRESENTATION_LEN) {
     status = ARES_EBADNAME;
     goto done;
   }
@@ -544,6 +557,7 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
   ares_buf_t   *namebuf     = NULL;
   size_t        label_start = ares_buf_get_position(buf);
   size_t        name_len    = 0;
+  size_t        indir       = 0;
 
   if (buf == NULL) {
     return ARES_EFORMERR;
@@ -615,6 +629,16 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
         goto fail;
       }
 
+      /* Bound the number of indirections.  A name made purely of pointers never
+       * adds label bytes, so the length cap below can't stop it; the
+       * strictly-decreasing rule alone still allows thousands of jumps per name.
+       * No legitimate <= 255 octet name needs more than 128 labels/jumps. */
+      indir++;
+      if (indir > ARES_MAX_INDIRS) {
+        status = ARES_EBADNAME;
+        goto fail;
+      }
+
       /* First time we make a jump, save the current position */
       if (save_offset == 0) {
         save_offset = ares_buf_get_position(buf);
@@ -638,15 +662,16 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
 
     /* New label */
 
-    /* RFC 1035 3.1: a name is limited to 255 octets.  Enforce it during
-     * decompression so a chain of pointers can't expand a single name without
-     * bound.  Track the unescaped length (label data plus the separator that
-     * precedes each label after the first), matching ares_split_dns_name(). */
+    /* RFC 1035 3.1 limits a name to 255 octets.  Enforce it during
+     * decompression so labels reached through a chain of pointers can't expand a
+     * single name without bound.  Track the presentation length (label data plus
+     * the separator that precedes each label after the first), matching
+     * ares_split_dns_name() on the write side. */
     if (name_len) {
       name_len++;
     }
     name_len += c;
-    if (name_len > 255) {
+    if (name_len > ARES_MAX_NAME_PRESENTATION_LEN) {
       status = ARES_EBADNAME;
       goto fail;
     }
--- a/test/ares-test-parse.cc
+++ b/test/ares-test-parse.cc
@@ -219,5 +219,41 @@ TEST_F(LibraryTest, ParseMalformedRRCount) {
   EXPECT_EQ(nullptr, dnsrec);
 }
 
+TEST_F(LibraryTest, ParseRejectsOverlongName) {
+  // RFC 1035 3.1 limits a name to 255 octets.  A longer name (here built from
+  // six 63-octet labels) must be rejected instead of expanded.
+  std::vector<byte> data = {
+    0x12, 0x34, 0x81, 0x80,
+    0x00, 0x01,  // num questions
+    0x00, 0x01,  // num answer RRs
+    0x00, 0x00,  // num authority RRs
+    0x00, 0x00,  // num additional RRs
+    // Question
+    0x01, 'a', 0x00,
+    0x00, 0x01,  // type A
+    0x00, 0x01,  // class IN
+  };
+  // Answer name: 6 x 63 = 378 octets, then root terminator
+  for (int i = 0; i < 6; i++) {
+    data.push_back(63);
+    for (int j = 0; j < 63; j++) {
+      data.push_back('a');
+    }
+  }
+  data.push_back(0x00);
+  std::vector<byte> tail = {
+    0x00, 0x01,              // type A
+    0x00, 0x01,              // class IN
+    0x00, 0x00, 0x00, 0x00,  // TTL
+    0x00, 0x04,              // rdata length
+    0x01, 0x02, 0x03, 0x04,
+  };
+  data.insert(data.end(), tail.begin(), tail.end());
+
+  ares_dns_record_t *dnsrec = NULL;
+  EXPECT_EQ(ARES_EBADNAME, ares_dns_parse(data.data(), data.size(), 0, &dnsrec));
+  EXPECT_EQ(nullptr, dnsrec);
+}
+
 }  // namespace test
 }  // namespace ares
--- a/test/ares-test-parse.cc
+++ b/test/ares-test-parse.cc
@@ -233,7 +233,8 @@ TEST_F(LibraryTest, ParseRejectsOverlongName) {
     0x00, 0x01,  // type A
     0x00, 0x01,  // class IN
   };
-  // Answer name: 6 x 63 = 378 octets, then root terminator
+  // Answer name: 6 x 63 label octets + 5 separators = 383 presentation octets
+  // (> 255), then root terminator
   for (int i = 0; i < 6; i++) {
     data.push_back(63);
     for (int j = 0; j < 63; j++) {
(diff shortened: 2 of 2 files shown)


jmestwa-coder:
gentle ping

src/lib/record/ares_dns_name.c
```diff
@@ -637,6 +638,19 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
...
 
+    /* RFC 1035 3.1: a name is limited to 255 octets.  Enforce it during
+     * decompression so a chain of pointers can't expand a single name without
+     * bound.  Track the unescaped length (label data plus the separator that
+     * precedes each label after the first), matching ares_split_dns_name(). */
+    if (name_len) {
+      name_len++;
+    }
+    name_len += c;
+    if (name_len > 255) {
```
bradh352: **Major — the pointer-only compression chain is not bounded, so the DoS this PR targets is still open.**

`name_len` is only incremented in the "New label" branch. The pointer/redirect branch (`(c & 0xc0) == 0xc0`, ~L579-628) hits `continue` before ever reaching this check, so a name built **purely from compression pointers** never touches `name_len` and this cap never fires. The only remaining guard on pointer-only chains is the strictly-decreasing rule plus the 14-bit offset ceiling (16383), which bounds a *single* name to ~8191 jumps — not to O(1).

This was verified empirically against this branch: a valid 65,529-byte response (an ~8191-link descending pointer chain + ~4126 minimal RRs whose NAME is a single pointer to the top of the chain) parses in **~889 ms** and returns **`ARES_SUCCESS`**, the same order of magnitude as the ~1.2 s attack this PR set out to eliminate. It is arguably *worse* than the label variant: a label-based overlong name now aborts early with `EBADNAME` on the first name, but the pointer-only name returns SUCCESS, so **every one of the N RRs is fully walked** with no early exit.

Suggested complement — bound the number of indirections. A legitimate <=255-octet name has <=128 labels, so it can never legitimately require more than ~128 pointer jumps:

```c
size_t indir = 0;   /* declare alongside name_len */
...
if ((c & 0xc0) == 0xc0) {
    ...
    if (++indir > 128) {   /* pointer-chain bound; far above any legit name */
      status = ARES_EBADNAME;
      goto fail;
    }
    ...
}
```

This kills the pointer-only chain (8191 >> 128) while leaving all legitimate names untouched, and complements (does not replace) the length cap. Please also add a regression test that feeds a pure pointer-only chain and asserts `ARES_EBADNAME` once the cap is in.

src/lib/record/ares_dns_name.c
```diff
@@ -637,6 +638,19 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
 
     /* New label */
 
+    /* RFC 1035 3.1: a name is limited to 255 octets.  Enforce it during
```
bradh352: **Low — the comment says "255 octets" but the metric counted is presentation length, not wire length.**

This counts `label bytes + separators`, which equals `wire_length - 2`. RFC 1035 3.1's 255 is the *wire* limit (label octets **and** length octets, plus the root octet), i.e. presentation <= 253. So this admits names up to 257 wire octets. That is intentional and consistent with the write side (`ares_split_dns_name`, L343), and it is never *too strict* (no compliant name is ever rejected — good for interop), so keeping the value at 255 is fine. Only the comment wording is imprecise about which length is being counted — worth a word so the next reader doesn't "fix" it to 253 and break parity with the write path.

Also: `255` is now duplicated across the read cap (here) and the two write-side checks (L343, L51). Consider a single named constant (note the `255` at L230 is unrelated — it validates an escape-octet *value*; and `NS_MAXCDNAME`/`NS_MAXDNAME` carry wire semantics, so a dedicated presentation-length constant is clearer).

test/ares-test-parse.cc
```diff
@@ -219,5 +219,41 @@ TEST_F(LibraryTest, ParseMalformedRRCount) {
   EXPECT_EQ(nullptr, dnsrec);
 }
 
+TEST_F(LibraryTest, ParseRejectsOverlongName) {
```
bradh352, line 222: **Medium — the scenario that motivated this PR (compression-pointer amplification) is not tested.**

The production comment says the cap exists "so a chain of pointers can't expand a single name without bound," but this test uses only *literal* labels. Please also cover a name assembled through backward compression pointers so the length accumulates across jumps. (This is distinct from, and in addition to, the pure pointer-only chain that the Major review comment shows is not yet bounded.)

```cpp
TEST_F(LibraryTest, ParseRejectsOverlongNameViaPointers) {
  // 5 chained 63-octet segments -> 63*5 + 4 separators = 319 > 255,
  // assembled purely through backward compression pointers.
  std::vector<byte> data = {
    0x12,0x34,0x81,0x80, 0x00,0x01, 0x00,0x01, 0x00,0x00, 0x00,0x00,
  };
  std::vector<size_t> off;
  auto add_seg = [&](bool terminate, size_t ptr_to){
    off.push_back(data.size());
    data.push_back(63);
    for (int j=0;j<63;j++) data.push_back('a');
    if (terminate) { data.push_back(0x00); }
    else { data.push_back(0xc0 | ((ptr_to>>8)&0x3f)); data.push_back(ptr_to & 0xff); }
  };
  add_seg(true, 0);        // seg5 (deepest) -> root
  add_seg(false, off[0]);  // seg4 -> seg5
  add_seg(false, off[1]);  // seg3 -> seg4
  add_seg(false, off[2]);  // seg2 -> seg3
  add_seg(false, off[3]);  // seg1 -> seg2
  size_t qname = off[4];   // question name = pointer to seg1
  data.push_back(0xc0 | ((qname>>8)&0x3f)); data.push_back(qname & 0xff);
  std::vector<byte> tail = {0x00,0x01, 0x00,0x01, 0xc0,0x0c, 0x00,0x01,
                            0x00,0x01, 0x00,0x00,0x00,0x00, 0x00,0x04, 1,2,3,4};
  data.insert(data.end(), tail.begin(), tail.end());
  ares_dns_record_t *dnsrec = NULL;
  EXPECT_EQ(ARES_EBADNAME, ares_dns_parse(data.data(), data.size(), 0, &dnsrec));
  EXPECT_EQ(nullptr, dnsrec);
}
```

test/ares-test-parse.cc
```diff
@@ -219,5 +219,41 @@ TEST_F(LibraryTest, ParseMalformedRRCount) {
...
+    0x00, 0x01,              // type A
+    0x00, 0x01,              // class IN
+    0x00, 0x00, 0x00, 0x00,  // TTL
+    0x00, 0x04,              // rdata length
+    0x01, 0x02, 0x03, 0x04,
+  };
+  data.insert(data.end(), tail.begin(), tail.end());
+
+  ares_dns_record_t *dnsrec = NULL;
+  EXPECT_EQ(ARES_EBADNAME, ares_dns_parse(data.data(), data.size(), 0, &dnsrec));
```
bradh352, line 255: **Medium — no off-by-one boundary test (255 accept vs 256 reject).**

This test lands at 383 octets — 128 past the limit — so it would still pass even if the bound were mistakenly written `>= 255` or `> 300`. The exact `> 255` edge is the regression-prone line (it is what sank the sibling PR #1173). Add a pair, using the parse counting `sum(labels) + (labels-1)`:

- **255 -> accept:** 4x63 = 252 + 3 separators = 255 (`255 > 255` is false).
- **256 -> reject:** labels 63,63,63,62,1 = 252 + 4 separators = 256 -> `EBADNAME`.

```cpp
TEST_F(LibraryTest, ParseAcceptsMaxLengthName) {   // exactly 255
  std::vector<byte> data = {0x12,0x34,0x81,0x80,0x00,0x01,0x00,0x01,
                            0x00,0x00,0x00,0x00, 0x01,'a',0x00, 0x00,0x01,0x00,0x01};
  for (int i=0;i<4;i++){ data.push_back(63); for(int j=0;j<63;j++) data.push_back('a'); }
  data.push_back(0x00);
  std::vector<byte> tail = {0x00,0x01,0x00,0x01,0x00,0x00,0x00,0x00,0x00,0x04,1,2,3,4};
  data.insert(data.end(),tail.begin(),tail.end());
  ares_dns_record_t *r=NULL;
  EXPECT_EQ(ARES_SUCCESS, ares_dns_parse(data.data(),data.size(),0,&r));
  EXPECT_NE(nullptr, r); ares_dns_record_destroy(r);
}
// 256-reject variant: same but five labels 63,63,63,62,1 -> expect ARES_EBADNAME, nullptr.
```

test/ares-test-parse.cc
```diff
@@ -219,5 +219,41 @@ TEST_F(LibraryTest, ParseMalformedRRCount) {
...
+    0x00, 0x01,  // num questions
+    0x00, 0x01,  // num answer RRs
+    0x00, 0x00,  // num authority RRs
+    0x00, 0x00,  // num additional RRs
+    // Question
+    0x01, 'a', 0x00,
+    0x00, 0x01,  // type A
+    0x00, 0x01,  // class IN
+  };
+  // Answer name: 6 x 63 = 378 octets, then root terminator
```
bradh352: **Nit — the octet count in this comment ignores separators.** "6 x 63 = 378" is neither the presentation length (378 + 5 separators = 383) nor the wire length (385). Suggest: `// 6 x 63 label octets + 5 separators = 383 presentation octets (> 255)`.

bradh352 (changes requested):
## Review summary

Thanks for this — the direction is right and the change is small and clean (counting matches the write side, `sum(labels) + (N-1)`, correctly avoiding the off-by-one that closed #1173). But as written it **does not actually close the DoS it targets**, so I'm marking this as request-changes.

### Blocking
- **Major (inline on `ares_dns_name.c`):** the cap only bounds *label-bearing* expansion. A name made purely of compression pointers never increments `name_len`, so the cap never fires. A crafted 64 KB reply still parses in ~889 ms (measured on this branch) and returns `ARES_SUCCESS` — same order as the attack this PR set out to stop, and worse in that it doesn't abort early. Fix: also cap pointer indirections per name (~128). See the inline comment for the diff and reproduction.

### Non-blocking
- **Medium (test):** the compression-pointer amplification (the PR's actual motivation) isn't tested — only literal labels. Add a pointer-chain test (snippet inline).
- **Medium (test):** no off-by-one boundary test (255 accept / 256 reject); the current test lands 128 octets past the limit and wouldn't catch a `>=`-vs-`>` mistake (snippet inline).
- **Low (RFC/consistency):** the counted metric is presentation length (`wire - 2`), ~2 octets looser than strict RFC 1035 3.1; intentional and interop-safe (never rejects a compliant name), but the "255 octets" comment is imprecise, and `255` is now duplicated across read/write paths — consider a named constant. (inline)
- **Low (coverage):** only the *answer* RR name path is tested; the question-name and rdata target-name paths (CNAME/NS/MX/SOA/SRV) run through the same function but aren't exercised.
- **Low (process):** this duplicates your own still-open #1114 and the closed #1173 — worth closing one to consolidate, and filing/linking a tracking issue for the CPU-DoS since a concrete scenario exists and none is referenced.
- **FYI (not a defect):** the cap bounds the *unescaped* length to 255, but the returned string is the *escaped* form and can reach ~1020 bytes (each non-printable octet -> `\\DDD`). No downstream code assumes <=255 today, so no active bug — noting it because it interacts with the truncating `name_copy[512]` on re-serialization.

### Confirmed correct (for the record)
`name_len` persists across pointer jumps without reset; no integer overflow (`size_t`, label <=63, capped at 255); the cap runs on the skip path (`name==NULL`) too; `ARES_EBADNAME` is the consistent error code; this is the single decompression choke point on the read path; and the write-side checks are independent and correctly left untouched. Full offline suite (1063 tests) passes.

src/lib/record/ares_dns_name.c
jmestwa-coder: good catch. a pointer-only name adds no label bytes, so the length cap never fires on it. added an indirection counter capped at 128 (a <=255 octet name is <=128 labels) that returns EBADNAME, next to the existing length cap. also added ParseRejectsPointerChain that feeds a pure pointer chain (stashed in a raw RR rdata so the pointers arent parsed as names) and asserts EBADNAME.

jmestwa-coder: good catch. a pointer-only name adds no label bytes, so the length cap never fires on it. added an indirection counter capped at 128 (a <=255 octet name is <=128 labels) that returns EBADNAME, next to the existing length cap. also added ParseRejectsPointerChain that feeds a pure pointer chain (stashed in a raw RR rdata so the pointers arent parsed as names) and asserts EBADNAME.

src/lib/record/ares_dns_name.c
jmestwa-coder: reworded the comment to say presentation length, and pulled 255 into a shared ARES_MAX_NAME_PRESENTATION_LEN used by the read cap and both write-side checks. kept the value at 255 to stay in parity with the write side.

test/ares-test-parse.cc
jmestwa-coder: added ParseRejectsOverlongNameViaPointers. one thing on the snippet: it lays the segment data at offset 12, which is where the question name starts, so the parser reads the segments as the question name and everything after misaligns (i was getting EFORMERR). reworked it so the answer name is a 63 octet label plus a pointer back to the question name, which accumulates 5x63 across the jump and gets rejected.

test/ares-test-parse.cc
jmestwa-coder: added both: ParseAcceptsMaxLengthName (4x63 = 255, accepted) and ParseRejectsName256 (labels 63,63,63,62,1 = 256, rejected).

test/ares-test-parse.cc
jmestwa-coder: fixed, now reads 6 x 63 label octets + 5 separators = 383 presentation octets (> 255).

jmestwa-coder:
pushed. summary:

- added a per-name compression-pointer indirection cap (128) so a pointer-only name gets rejected instead of walked, since the length cap only counts label bytes. this closes the ~0.9s pointer-only case you measured
- pulled 255 into a shared constant and reworded the length comment to say presentation length
- tests: ParseRejectsPointerChain (pure pointer chain), ParseRejectsOverlongNameViaPointers (length across a backward pointer), and ParseAcceptsMaxLengthName/ParseRejectsName256 for the 255/256 edge

the duplicate #1114 is already closed. full offline suite passes here.

bradh352 (approved):
## Review v2 — approve

Thanks for the quick turnaround. All five round-1 items are resolved and I verified the fix end-to-end on `83165278`.

**Verification performed:** built clean with CMake+Ninja (Debug), zero warnings on the changed files; all 5 new tests pass. Confirmed `ParseRejectsPointerChain` is a real guard — removing only the `indir` cap block makes it return `ARES_SUCCESS`/non-null and fail; restored, it's green. Re-derived the worst case: label work is bounded by the 255 cap (`name_len` accumulates across jumps and never resets, so a label can't be re-walked more than ~4×), jump work by the 128 cap → per-name work is a constant and total packet work is now linear. The pointer-only chain (~889 ms) now aborts with `EBADNAME`. **DoS closed.**

**Round-1 items (all confirmed):**
- ✅ **Major** (pointer-only chain unbounded) → 128-indirection cap; counting matches the write side and persists across jumps.
- ✅ **Medium** (compression amplification untested) → `ParseRejectsOverlongNameViaPointers` (319 accumulated across a backward pointer). Your rework of my snippet was correct — the original offset-12 layout collided with the question name.
- ✅ **Medium** (off-by-one boundary) → `ParseAcceptsMaxLengthName` (255 accept) / `ParseRejectsName256` (256 reject); the `> 255` edge is pinned.
- ✅ **Low** (presentation-length wording + duplicated `255`) → `ARES_MAX_NAME_PRESENTATION_LEN` shared across read/write; comment reworded.
- ✅ **Nit** (separator math) → fixed.
- Also confirmed: exact write-side parity (`Σlabel + (N-1)`), no integer overflow, C89-clean, cap runs on the skip path too, and this remains the single decompression choke point covering question / owner / rdata-target names.

**Non-blocking follow-ups (none block merge):**
- **Low (pre-existing, out of scope):** the cap bounds *unescaped* length to 255, but the returned string is escaped and can reach ~1 KB; re-serialization copies through `name_copy[512]` in `ares_dns_name_write`, which `ares_strcpy` truncates *safely* (no overflow) but silently — potentially splitting a `\DDD` escape. This PR *improves* the situation (previously unbounded). Worth a separate tracking issue now that the bound is concrete, not a change here.
- **Low (test coverage):** rdata target-name paths (CNAME/NS/MX/SOA/SRV) run through the same function but still aren't exercised directly; the question-name path now is. Purely additive — identical code path.
- **FYI:** #1114 and #1173 are both closed, so the duplicate concern is resolved. Heads-up that open #1159 also edits `ares_nameoffset_create` — expect a trivial merge conflict there (independent write-side change).

PR #1165, opened 2026-06-07
ares_array: reset offset when array becomes empty

dimbleby:
ares_array_claim_at() removes the first element in O(1) by advancing arr->offset instead of shifting. If every element is claimed from the front, offset can climb to arr->alloc_cnt. Reusing the array afterwards (appending while draining from the front) then made ares_array_insert_at() compact via ares_array_move() with an out-of-range source index, failing with ARES_EFORMERR.

Reset offset to 0 once the array is empty. There is nothing to preserve, and this is provably the only case ares_array_move() could fail: the invariant offset + cnt <= alloc_cnt holds everywhere, so offset can only reach alloc_cnt when cnt == 0.

Add a regression test exercising interleaved front-claim and append.

### DIFF
--- a/src/lib/dsa/ares_array.c
+++ b/src/lib/dsa/ares_array.c
@@ -378,6 +378,13 @@ ares_status_t ares_array_claim_at(void *dest, size_t dest_size,
   }
 
   arr->cnt--;
+
+  /* When empty, reset offset so a later insert doesn't compact from an
+   * out-of-range index (offset can reach alloc_cnt after front claims). */
+  if (arr->cnt == 0) {
+    arr->offset = 0;
+  }
+
   return ARES_SUCCESS;
 }
 
--- a/src/lib/dsa/ares_array.c
+++ b/src/lib/dsa/ares_array.c
@@ -380,7 +380,10 @@ ares_status_t ares_array_claim_at(void *dest, size_t dest_size,
   arr->cnt--;
 
   /* When empty, reset offset so a later insert doesn't compact from an
-   * out-of-range index (offset can reach alloc_cnt after front claims). */
+   * out-of-range index (offset can reach alloc_cnt after front claims).
+   * This also protects ares_array_finish(), which has the same
+   * move(0, offset) and would otherwise fail on an offset == alloc_cnt
+   * array. */
   if (arr->cnt == 0) {
     arr->offset = 0;
   }
--- a/test/ares-test-internal.cc
+++ b/test/ares-test-internal.cc
@@ -1488,22 +1488,32 @@ TEST_F(LibraryTest, ArrayClaimFrontThenReuse) {
   ares_array_t *a = ares_array_create(sizeof(size_t), NULL);
   EXPECT_NE(nullptr, a);
 
-  /* Interleave front-claim and append many times. */
-  for (size_t i = 0; i < 1000; i++) {
-    size_t  val = i;
-    size_t *ptr = NULL;
+  for (size_t iter = 0; iter < 4; iter++) {
+    /* Fill exactly to the minimum allocation (ARES__ARRAY_MIN == 4) so a
+     * full front-drain later pushes offset to precisely alloc_cnt. */
+    for (size_t i = 0; i < 4; i++) {
+      size_t val = iter * 100 + i;
+      EXPECT_EQ(ARES_SUCCESS, ares_array_insertdata_last(a, &val));
+    }
+    EXPECT_EQ((size_t)4, ares_array_len(a));
 
-    EXPECT_EQ(ARES_SUCCESS, ares_array_insertdata_last(a, &val));
-    EXPECT_EQ((size_t)1, ares_array_len(a));
-
-    EXPECT_EQ(ARES_SUCCESS, ares_array_claim_at(NULL, 0, a, 0));
+    /* Drain ALL elements from the front so offset climbs to alloc_cnt (the
+     * condition the fix guards); verify copy-out value and FIFO order. */
+    for (size_t i = 0; i < 4; i++) {
+      size_t out = 0;
+      EXPECT_EQ(ARES_SUCCESS, ares_array_claim_at(&out, sizeof(out), a, 0));
+      EXPECT_EQ(iter * 100 + i, out);
+    }
     EXPECT_EQ((size_t)0, ares_array_len(a));
 
-    /* Array is now empty; appending must continue to work. */
+    /* Array is now empty with offset == alloc_cnt; appending must still
+     * work, and the data must be readable back out. */
+    size_t *ptr = NULL;
     EXPECT_EQ(ARES_SUCCESS, ares_array_insert_last((void **)&ptr, a));
     EXPECT_NE(nullptr, ptr);
-    *ptr = i;
-    EXPECT_EQ(ARES_SUCCESS, ares_array_claim_at(NULL, 0, a, 0));
+    *ptr = 424242;
+    EXPECT_EQ((size_t)424242, *(size_t *)ares_array_at(a, 0));
+    EXPECT_EQ(ARES_SUCCESS, ares_array_remove_first(a));
   }
 
   EXPECT_EQ((size_t)0, ares_array_len(a));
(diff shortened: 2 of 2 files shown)


src/lib/dsa/ares_array.c
```diff
@@ -378,6 +378,13 @@ ares_status_t ares_array_claim_at(void *dest, size_t dest_size,
   }
 
   arr->cnt--;
+
+  /* When empty, reset offset so a later insert doesn't compact from an
+   * out-of-range index (offset can reach alloc_cnt after front claims). */
+  if (arr->cnt == 0) {
```
bradh352, line 387: **Nit (positive/optional).** Correct and minimal — this is the single decrement point for `cnt`, so it's the right home for the reset. Optional: the comment could note this also protects `ares_array_finish()`, which has the same `move(0, offset)` and would otherwise fail (return `NULL`, silently dropping the buffer) on an `offset == alloc_cnt` array.

test/ares-test-internal.cc
```diff
@@ -1480,6 +1480,36 @@ TEST_F(LibraryTest, Array) {
...
+/* Regression: repeatedly claiming the first element drives the internal
+ * offset up to the allocation size.  Re-inserting afterwards must still
+ * succeed (previously ares_array_insert_at() failed with ARES_EFORMERR
+ * because compaction tried to move from an out-of-range source index). */
+TEST_F(LibraryTest, ArrayClaimFrontThenReuse) {
+  ares_array_t *a = ares_array_create(sizeof(size_t), NULL);
+  EXPECT_NE(nullptr, a);
+
+  /* Interleave front-claim and append many times. */
+  for (size_t i = 0; i < 1000; i++) {
```
bradh352: **Medium.** With the fix applied, `offset` resets to 0 on every iteration, so this loop never lets `offset` climb above 1 — it never reproduces the multi-front-claim state (`offset -> alloc_cnt`) that the fix actually guards. Every iteration is byte-identical, so the 1000 iterations are redundant (it fails on iteration 3 without the fix; ~3 give identical coverage). Also `*ptr = i` is written but never read back, so a fix that reset `offset` but corrupted the payload/order would still pass.

Suggest a version that fills past `ARES__ARRAY_MIN`, drains **all** elements from the front (so `offset` genuinely climbs to `alloc_cnt`), and verifies value + FIFO order and copy-out:

```c
TEST_F(LibraryTest, ArrayClaimFrontThenReuse) {
  ares_array_t *a = ares_array_create(sizeof(size_t), NULL);
  EXPECT_NE(nullptr, a);

  for (size_t iter = 0; iter < 4; iter++) {
    /* Fill past the minimum allocation so offset can actually climb. */
    for (size_t i = 0; i < 6; i++) {
      size_t val = iter * 100 + i;
      EXPECT_EQ(ARES_SUCCESS, ares_array_insertdata_last(a, &val));
    }
    EXPECT_EQ((size_t)6, ares_array_len(a));

    /* Drain ALL from the front -> offset climbs to alloc_cnt (the trigger);
     * verify copy-out value and FIFO order via non-NULL dest. */
    for (size_t i = 0; i < 6; i++) {
      size_t out = 0;
      EXPECT_EQ(ARES_SUCCESS, ares_array_claim_at(&out, sizeof(out), a, 0));
      EXPECT_EQ(iter * 100 + i, out);
    }
    EXPECT_EQ((size_t)0, ares_array_len(a));

    /* Reuse after empty: without the offset reset this returns EFORMERR. */
    size_t *ptr = NULL;
    EXPECT_EQ(ARES_SUCCESS, ares_array_insert_last((void **)&ptr, a));
    EXPECT_NE(nullptr, ptr);
    *ptr = 424242;
    EXPECT_EQ(*(size_t *)ares_array_at(a, 0), (size_t)424242);
    EXPECT_EQ(ARES_SUCCESS, ares_array_remove_first(a));  /* claim/destructor path */
  }

  EXPECT_EQ((size_t)0, ares_array_len(a));
  ares_array_destroy(a);
}
```

bradh352 (changes requested):
## Summary

The fix itself is **correct, minimal, and complete** — I traced it and verified it independently, and confirmed empirically that the new test fails on the 3rd iteration with `ARES_EFORMERR` without the fix and passes with it. I'm requesting changes only for the **regression test**, which as written doesn't actually exercise the condition the fix guards (see inline). Happy to flip to approve once the test is strengthened.

### Correctness (verified)
The invariant `offset + cnt <= alloc_cnt` holds after every mutator (`insert_at`, `claim_at`, `set_size`, `finish`), so `offset == alloc_cnt` is reachable **only** at `cnt == 0`. `offset` is incremented in exactly one place and `cnt` decremented in exactly one place, so resetting `offset` when `cnt` hits 0 closes the bug with no remaining path. All four `ares_array_move()` call sites are provably in-range post-fix. The fix also incidentally protects `ares_array_finish()`, which has the same `move(0, offset)` that would otherwise return `NULL` on an `offset == alloc_cnt` array.

### Security / safety
No security concern. Pre-fix, the out-of-range access was **caught** by the guard at `ares_array.c:126` and failed safe with `ARES_EFORMERR` — there was never an out-of-bounds access. This is a spurious-error/correctness fix, not a memory-safety fix.

### Scope / real-world reachability
The bug requires an array to be **fully front-drained and then reused**. Auditing every front-removal call site, **no current c-ares code path triggers it**: the only array drained from index 0 in a loop is `requeue` (`ares_process.c:631-660`), and it's drained-then-destroyed (`:665`), never reused. `ares_array_remove_first` has no callers at all. So this is effectively **hardening / defense-in-depth**. Consider reframing the description accordingly — and could you note how it was found (inspection, fuzzing, or a real/planned caller that reuses after front-drain)? That determines whether "observed failure" framing is warranted.

### No dead code introduced
The `insert_at` compaction shift, the `finish` relocation, and the O(1) front-claim optimization all remain necessary for the partial-drain case (`cnt>0, offset>0`).

### Pre-existing nits (not introduced here, optional)
- `ares_array_move()` (`ares_array.c:144`): `nmembers = arr->cnt - (src_idx - arr->offset)` underflows if a future caller passes `src_idx < offset`. Safe for all current call sites, but the helper is unguarded — worth an `assert`/comment.
- `ares_array_remove_at()` (`ares_array.c:386-387`): calls `ares_array_at(arr, idx)` before the `arr == NULL` check. Null-safe, so no crash, but reads backwards.

dimbleby:
The AI proposed test script fix was good in principle but needed a small tweak actually to hit the original bug (6 is not a power of two, rounds up to 8, so offset only reaches 6).  That aside, markups pushed.

bradh352 (approved):

x4m:
I came here after AI-assisted analysis of [proclaimed UAF](https://github.com/bikini/exploitarium/tree/main/c-ares-tcp-uaf-calc-poc).
The agent confirmed the double-free on c93e50f3 via trace instrumentation
(REQUEUE_ENDQUERY appeared twice for the same qid in the requeue flush loop),
and identified this PR as an accidental fix.

Commit 9c11ea4b ("ares_array: reset offset when array becomes empty")
accidentally prevented a use-after-free in the REQUEUE_ENDQUERY path.
The array fix stopped duplicate entries from appearing in the requeue
queue, which was the trigger for the double free. However the root cause
in ares_process.c was not addressed.

The unsafe pattern in ares_process.c (REQUEUE_ENDQUERY branch):

    query->callback(query->arg, entry.status, query->timeouts, entry.dnsrec);
    ares_free_query(query);

The callback may call ares_close_connection(), which calls
ares_requeue_queries(), which appends another REQUEUE_ENDQUERY for the
same query. If the array code ever changes again this double free can
reappear.

Suggested fix: call ares_query_remove_from_conn(query) before the
callback so the query is no longer visible to ares_requeue_queries()
when the connection is closed:

    ares_query_remove_from_conn(query);
    query->callback(query->arg, entry.status, query->timeouts, entry.dnsrec);
    ares_free_query(query);

This makes the fix explicit and independent of ares_array internals.

Sorry, if this is just AI-noise. I'm completely new to the area.

dimbleby:
@x4m comments on closed pull requests are unlikely to attract any attention.  If you think you have a bug - (ask the AI to) write a testcase reproducing it, and report a new issue.

PR #1166, opened 2026-06-10
Add IBM i PASE platform support for DNS configuration

sravani1510:
Implement native DNS configuration retrieval for IBM i PASE using the 
`QtocRtvTCPA` ILE API. This fixes DNS resolution failures on IBM i where
the system uses `CHGTCPDMN` instead of `/etc/resolv.conf`.

Tested on IBM i with Node.js 24.16.0, resolving ETIMEDOUT errors.

```bash
bash-5.2$ /QOpenSys/pkgs/lib/nodejs24/bin/node --version
v24.14.1
bash-5.2$ /QOpenSys/pkgs/lib/nodejs24/bin/node  -e 'require("dns").resolve("smtp.office365.com", function() { console.log(arguments); })'
[Arguments] {
  '0': Error: queryA ECONNREFUSED smtp.office365.com
      at QueryReqWrap.onresolve [as oncomplete] (node:internal/dns/callback_resolver:45:19) {
    errno: undefined,
    code: 'ECONNREFUSED',
    syscall: 'queryA',
    hostname: 'smtp.office365.com'
  }
}
bash-5.2$ yum install nodejs24-24.16.0-9999.ibmi7.3.ppc64.rpm 
Loaded plugins: changelog, downloadonly, security
Setting up Install Process
Examining nodejs24-24.16.0-9999.ibmi7.3.ppc64.rpm: nodejs24-24.16.0-9999.ppc64
Marking nodejs24-24.16.0-9999.ibmi7.3.ppc64.rpm as an update to nodejs24-24.14.1-1.ppc64
Resolving Dependencies
There are unfinished transactions remaining. You might consider running yum-complete-transaction first to finish them.
--> Running transaction check
---> Package nodejs24.ppc64 0:24.14.1-1 will be updated
---> Package nodejs24.ppc64 0:24.16.0-9999 will be an update
--> Finished Dependency Resolution

Dependencies Resolved

===========================================================================================================================================================================
 Package                          Arch                          Version                                 Repository                                                    Size
===========================================================================================================================================================================
Updating:
 nodejs24                         ppc64                         24.16.0-9999                            /nodejs24-24.16.0-9999.ibmi7.3.ppc64                         206 M

Transaction Summary
===========================================================================================================================================================================
Upgrade       1 Package

Total size: 206 M
Is this ok [y/N]: y
Downloading Packages:
Running Transaction Check
Running Transaction Test
Transaction Test Succeeded
Running Transaction
  Updating   : nodejs24-24.16.0-9999.ppc64                                                                                                                             1/2 
  Cleanup    : nodejs24-24.14.1-1.ppc64                                                                                                                                2/2 

Updated:
  nodejs24.ppc64 0:24.16.0-9999                                                                                                                                            

Complete!
bash-5.2$ /QOpenSys/pkgs/lib/nodejs24/bin/node --version
v24.16.0
bash-5.2$ /QOpenSys/pkgs/lib/nodejs24/bin/node  -e 'require("dns").resolve("smtp.office365.com", function() { console.log(arguments); })'
[Arguments] {
  '0': null,
  '1': [ '52.96.121.162', '52.96.57.114', '52.96.121.98', '40.99.169.130' ]
}

```


sravani1510:
Fixes https://github.com/nodejs/node/issues/53948

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,250 @@
...
+        if (status != ARES_SUCCESS) {
+          ares_free(buffer);
+          return status;
+        }
+      }
+    }
+  }
+
+  /* Extract domain search list (convert from EBCDIC) */
+  if (header2->search_list[0] != '\0' && header2->search_list[0] != ' ') {
```
kadler:
```suggestion
  if (header2->search_list[0] != '\0' && header2->search_list[0] != '\x40') {
```

search_list will be in EBCDIC so you need an EBCDIC blank.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,250 @@
...
+    char search_ascii[256];
+    int  j;
+    for (j = 0; j < 256 && header2->search_list[j] != '\0' &&
+                header2->search_list[j] != ' ';
```
kadler:
```suggestion
                header2->search_list[j] != '\x40';
```

Same here. Might be worth a `#define`.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,250 @@
...
+        }
+      }
+    }
+  }
+
+  /* Extract domain search list (convert from EBCDIC) */
+  if (header2->search_list[0] != '\0' && header2->search_list[0] != ' ') {
+    char search_ascii[256];
+    int  j;
+    for (j = 0; j < 256 && header2->search_list[j] != '\0' &&
```
kadler:
```suggestion
    for (j = 0; j < sizeof(header2->search_list) && header2->search_list[j] != '\0' &&
```

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,250 @@
...
+    return 'J' + (c - 0xD1);
+  }
+  if (c >= 0xE2 && c <= 0xE9) {
+    return 'S' + (c - 0xE2);
+  }
+
+  /* Numbers */
+  if (c >= 0xF0 && c <= 0xF9) {
+    return '0' + (c - 0xF0);
+  }
```
kadler: I've usually done this with a lookup table, which is much faster at the expense of 256 bytes.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,250 @@
...
+  return c; /* Keep as-is for unmapped characters */
+}
+
+/* Thread-safe initialization of ILE API */
+static ares_status_t load_ile_api(void)
+{
+  ares_status_t      status = ARES_SUCCESS;
+  unsigned long long actmark;
+
+  pthread_mutex_lock(&ile_mutex);
```
kadler, line 177: [pthread_once](https://www.man7.org/linux/man-pages/man3/pthread_once.3p.html) would normally be a better mechanism for this, but you will need to re-init if the parent forks since ILE pointers are invalidated in the child after a fork. You need to cache the pid and re-init if it changes.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,250 @@
...
+      (dns_list_item_t *)(buffer + header2->dns_list_offset +
+                          (i * header2->dns_list_entry_size));
+
+    char ip[INET6_ADDRSTRLEN];
+    int  af = (item->version == 1 ? AF_INET : AF_INET6);
+
+    if (inet_ntop(af, item->ip_address, ip, sizeof(ip)) != NULL) {
+      /* Skip empty or invalid entries */
+      if (strlen(ip) > 0 && strcmp(ip, "0.0.0.0") != 0 &&
+          strcmp(ip, "::") != 0) {
```
kadler: Does the API generally return these for some reason or is it just if a user has set it? Because if a user has set these, then it seems like that's their problem.

kadler (commented):
I left a few comments. You are also missing configuration of various options that are available from the API: https://github.com/c-ares/c-ares/blob/main/src/lib/ares_sysconfig_files.c#L526-L533

eg.
- lookup -> search order
- timeout -> time interval
- attempts -> retries
- rotate -> initial domain name server
- usevc -> DNS protocol

src/lib/ares_sysconfig_pase.c
sravani1510: Replaced with the EBCDIC blank

src/lib/ares_sysconfig_pase.c
sravani1510: Added #define EBCDIC_SPACE '\x40' and replaced with the same

src/lib/ares_sysconfig_pase.c
sravani1510: modified this instead of hard coding

src/lib/ares_sysconfig_pase.c
sravani1510: Implemented 256-byte EBCDIC-to-ASCII lookup table

src/lib/ares_sysconfig_pase.c
sravani1510: Added PID caching to detect fork and re-initialize ILE pointers when process ID changes.

src/lib/ares_sysconfig_pase.c
sravani1510: The `QtocRtvTCPA` API returns 0.0.0.0 (or ::) for unused DNS server slots as a design choice. The API always returns a fixed number of slots (typically 10), filling empty ones with 0.0.0.0/:: rather than reducing dns_list_count. This is API behavior, not user configuration.

sravani1510:
@bagder @daviddrysdale @bradh352 
Could you please take a look at this PR

src/lib/ares_sysconfig_pase.c
kadler: I cannot replicate this behavior. On all the systems I have tested this on, dns_list_count is always less than 10.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
+#include "ares_private.h"
```
bradh352, line 27: **[Major] Missing MIT/SPDX license header.**

Every other c-ares source file opens with the standard 23-line MIT block ending in `SPDX-License-Identifier: MIT` (see `ares_sysconfig_mac.c:1-24`). This file jumps straight to `#include`. Please prepend the standard header (updating the copyright line) — this is a legal/process blocker for merge.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+        }
+      }
+    }
+  }
+
+  /* Extract domain search list (convert from EBCDIC) */
+  if (header2->search_list[0] != '\0' && header2->search_list[0] != EBCDIC_SPACE) {
+    char search_ascii[256];
+    int  j;
+    for (j = 0; j < sizeof(header2->search_list) && header2->search_list[j] != '\0' &&
```
bradh352: **[Major] Multi-domain search list is truncated to the first domain (+ a 1-byte stack overflow).**

The copy loop terminates at the **first** EBCDIC space (`!= EBCDIC_SPACE`), but IBM's TCPA1400 "Domain search list" is *space-separated* (up to six domains, blank-padded). So `"corp.example.com eng.example.com"` gets cut to `"corp.example.com"`, and the later `ares_strsplit(search_ascii, ", ")` has no separator left — only one domain ever survives.

The fix is to convert the **whole** field first and let `ares_strsplit` do the splitting: the table already maps EBCDIC space `0x40 -> ' '`, and `ares_strsplit` calls `ares_buf_split_str` *without* `ARES_BUF_SPLIT_ALLOW_BLANK`, so empty tokens (from the trailing blank padding) are dropped automatically (`ares_buf.c:956`).

Separately, this same loop is a **1-byte stack overflow**: `search_ascii[256]` with bound `j < sizeof(search_list)` (==256) lets `j` reach 256, then `search_ascii[256] = '\0'` writes one past the buffer. Sizing the buffer `+1` fixes it.

Suggest extracting a reusable buffer-level helper (also usable for `domain_name`, see the struct comment below) and rewriting the block:

```c
/* Convert an EBCDIC buffer (up to len bytes) into a NUL-terminated ASCII
 * string in out, which must be at least len+1 bytes.  Stops at the first
 * NUL; trailing EBCDIC blank (0x40) padding becomes ASCII spaces. */
static void ebcdic_to_ascii_str(const char *ebcdic, size_t len, char *out)
{
  size_t i;
  for (i = 0; i < len && ebcdic[i] != '\0'; i++) {
    out[i] = ebcdic_to_ascii((unsigned char)ebcdic[i]);
  }
  out[i] = '\0';
}
```

```c
  /* Domain search list: EBCDIC, space-separated, blank-padded.  Convert the
   * whole field, then split; ares_strsplit() splits on space/comma and drops
   * empty tokens, so the blank padding is handled automatically. */
  {
    char search_ascii[sizeof(header2->search_list) + 1];
    ebcdic_to_ascii_str(header2->search_list, sizeof(header2->search_list),
                        search_ascii);
    sysconfig->domains = ares_strsplit(search_ascii, ", ", &sysconfig->ndomains);
    /* NULL with ndomains==0 just means "no domains", not an allocation error */
    if (sysconfig->domains == NULL && sysconfig->ndomains > 0) {
      ares_free(buffer);
      return ARES_ENOMEM;
    }
  }
```

This also resolves the signed/unsigned comparison warning (`int j < sizeof(...)`).

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+
+  /* Configure timeout (time_interval is in seconds) */
+  if (header2->time_interval > 0) {
+    sysconfig->timeout_ms = header2->time_interval * 1000;
+  }
+
+  /* Configure retries (attempts) */
+  if (header2->retries > 0) {
+    sysconfig->tries = header2->retries;
+  }
```
bradh352: **[Major] `retries` -> `tries` is off by one, and the `> 0` guard drops a valid value.**

Per IBM's TCPA1400 docs, "Retries" is the number of **additional** attempts *after the first* (default 2), whereas c-ares `tries` is the **total** attempts. So the IBM default `retries=2` should yield `tries=3`, and `retries=0` ("one attempt, no retries") should yield `tries=1` — but the `> 0` guard drops it and leaves the c-ares default of 3. Since the API call succeeded, the value is always a valid 0–99, so set it unconditionally:

```suggestion
  /* IBM "retries" excludes the first attempt; c-ares "tries" is total attempts */
  sysconfig->tries = (size_t)header2->retries + 1;
```

bradh352 (commented):
Thanks for this — the hard parts are well done. I verified the receiver structs (`tcpa1100_t`/`tcpa1400_t`/`dns_list_item_t`) against IBM's documented TCPA1100/TCPA1400 formats and they match **field-for-field, offset-for-offset**; the EBCDIC `"TCPA1400"` format bytes, the `Qus_EC_t` error overlay, the `QTOCNETSTS`/`QtocRtvTCPA` names, the `dns_protocol`/`initial_server`/`search_order`/`time_interval` mappings, and the fork -> `getpid()` re-resolution (ILE pointers are invalidated across fork — IBM-doc confirmed) are all correct. Build wiring (Makefile.inc -> CMake) and dispatch order are also correct. Nice work, and thanks for addressing all of the first-round comments.

A number of things to resolve before merge (details inline):

**Major**
- Missing MIT/SPDX license header.
- Search-list copy loop stops at the first EBCDIC space, truncating multi-domain lists to one entry — plus a 1-byte stack overflow on the terminator. Suggested a reusable `ebcdic_to_ascii_str()` helper + whole-field conversion inline.
- `retries` -> `tries` is off by one (IBM "retries" excludes the first attempt) and drops `retries=0`.
- C89 declaration-after-statement will fail the PASE `-std=c90 -Werror` build.
- **No tests.** The EBCDIC table and the search-list parser are pure logic that can and should be unit-tested off-IBM-i. There's no existing platform-sysconfig parse test to follow, so this would establish the pattern — at minimum a table-driven test of the DNS charset and the space/comma-separated search-list split (which would have caught both the truncation and the `0x6B` comma issue below).

**Medium**
- Unvalidated API offsets (`additional_info_offset`, `dns_list_offset`, `dns_list_entry_size`) used for pointer arithmetic; also consider validating the header's `bytes_returned`.
- `dns_listening_port` ignored — non-53 configs silently query 53.
- Primary `domain_name` ignored as a fallback search domain.
- EBCDIC `0x6B` (comma) maps to `'k'`; the table is presented as complete but several non-hostname punctuation slots are wrong.

**Low / Nit**
- `lookups` set unconditionally + `ares_strdup` return unchecked (inline).
- `ARES_ELOADIPHLPAPI` reused for ILE-load failure — misleading though inert (inline).
- `time_interval * 1000` computed in `int` before the `size_t` store (inline).
- `qtocrtvtcpa_ptr` is read by `_ILECALL` outside `ile_mutex` — not realistically reachable (child is single-threaded post-fork, pointer stable once initialized), but formally an unsynchronized read.
- On the empty-servers fallback to `ares_init_sysconfig_files`, API-derived `timeout`/`tries`/`lookups`/`rotate` linger and mix with resolv.conf values (no leak — the files parser frees-before-set). Consider zeroing resolver fields before falling back, or documenting the intent.
- Nits: the `10` is duplicated (buffer `*10` vs loop `i < 10`) with no shared constant, and the `+ 512` slack is unexplained — a `#define ARES_PASE_MAX_DNS_SERVERS 10` + a comment would help; `buflen` is `unsigned int` where IBM's length parameter is a signed `BINARY(4)`; the binary-layout structs and `version == 1` (IPv4) are uncommented — a `static_assert` on the layout would guard against silent corruption. (Don't remove the "unused" struct fields — they define the wire layout.)

On the open `0.0.0.0`/`::` filter thread: I dug into the docs and I think @kadler is right; I'll follow up in that thread.

One process note: the description references `nodejs/node#53948` but doesn't GitHub-link it, so it won't auto-close — add a `Fixes:`/`Closes:` line if that's intended.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+
+  buffer = ares_malloc(buflen);
+  if (buffer == NULL) {
+    return ARES_ENOMEM;
+  }
+
+  memset(buffer, 0, buflen);
+
+  /* Prepare error code structure */
+  struct {
```
bradh352, line 224: **[Major] C89 declaration-after-statement — will fail the IBM i build.**

`err_code` (here), `signature[]`, `arglist`, and `format[]` are all declared *after* executable statements (`ares_malloc`, `memset`). c-ares mandates C90 and compiles with `-Wdeclaration-after-statement` (`DEVELOPER-NOTES.md`, `configure.ac`), typically under `-Werror`. Non-PASE CI won't catch this since the body is `#ifdef __PASE__`, but the actual PASE toolchain build will. Please hoist these declarations to the top of the function with the other locals, assigning the `.addr`/value fields afterward.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+
+  if (rc != ILECALL_NOERROR || err_code.bytes_available) {
+    ares_free(buffer);
+    /* Fall back to file-based configuration */
+    return ares_init_sysconfig_files(channel, sysconfig, ARES_TRUE);
+  }
+
+  /* Parse the returned data */
+  header  = (tcpa1100_t *)buffer;
+  header2 = (tcpa1400_t *)(buffer + header->additional_info_offset);
```
bradh352: **[Medium] Unvalidated API offsets used for pointer arithmetic.**

`header->additional_info_offset` (here), and `header2->dns_list_offset` / `dns_list_entry_size` below, are read straight from the API-filled buffer and used as offsets/stride into a fixed-size `buflen` allocation with no bounds validation (only `dns_list_count` is capped at 10). A short/truncated/malformed response then yields out-of-bounds reads. Note the code checks `err_code.bytes_available` but never the header's own `bytes_returned`/`bytes_available`, so it can't detect a short response before trusting these offsets. Consistent with the recent overflow-hardening (dbb3d2be, b73bb616), add guards, e.g.:

```c
if (header->additional_info_offset < 0 ||
    (size_t)header->additional_info_offset + sizeof(tcpa1400_t) > buflen) {
  ares_free(buffer);
  return ares_init_sysconfig_files(channel, sysconfig, ARES_TRUE);
}
...
if (header2->dns_list_entry_size < (int)sizeof(dns_list_item_t)) { /* bail */ }
size_t off = (size_t)header2->dns_list_offset +
             (size_t)i * (size_t)header2->dns_list_entry_size;
if (off + sizeof(dns_list_item_t) > buflen) break;
```
Real-world risk is bounded (trusted OS API into a buffer we sized), but it's cheap defense-in-depth.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+    char ip[INET6_ADDRSTRLEN];
+    int  af = (item->version == 1 ? AF_INET : AF_INET6);
+
+    if (inet_ntop(af, item->ip_address, ip, sizeof(ip)) != NULL) {
+      /* Skip empty or invalid entries that the API returns when slots are unused.
+       * The QtocRtvTCPA API returns 0.0.0.0 or :: for unused DNS server slots
+       * rather than reducing dns_list_count, so we filter them here. */
+      if (strlen(ip) > 0 && strcmp(ip, "0.0.0.0") != 0 &&
+          strcmp(ip, "::") != 0) {
+        status = ares_sconfig_append_fromstr(channel, &sysconfig->sconfig, ip,
```
bradh352: **[Medium] Non-standard DNS port is ignored.**

`header2->dns_listening_port` (parsed but unused) can be 1–65532 per IBM docs, and applies to all configured servers. Here the servers are appended bare, so a `CHGTCPDMN` config using a non-53 port will silently query port 53 — no resolution. `ares_sconfig_append_fromstr` accepts `[ip]:port` (see the libresolv path in `ares_sysconfig.c`). When `dns_listening_port` is non-zero and != 53, format `"[%s]:%d"` before appending.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+  int  dns_list_count;
+  int  dns_list_entry_size;
+  int  dns_protocol;
+  int  retries;
+  int  time_interval;
+  int  search_order;
+  int  initial_server;
+  int  dns_listening_port;
+  char hostname[64];
+  char domain_name[255];
```
bradh352, line 149: **[Medium] Primary `domain_name` is parsed but never used.**

On IBM i, `CHGTCPDMN DMNNAME(...)` sets the primary domain independently of the search list. If a system has a primary domain but an empty `search_list`, `sysconfig->domains` ends up empty — a partial re-run of the very problem this PR fixes. When the search list yields no domains (`ndomains == 0`), fall back to appending the EBCDIC-converted `domain_name` as a single search domain (mirrors resolv.conf `domain` semantics). The `ebcdic_to_ascii_str` helper proposed for the search list can be reused here.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+  0x38, 0x39, 0x3A, 0x3B, 0x3C, 0x3D, 0x3E, 0x3F,
+  /* 0x40 = EBCDIC space */
+  ' ',  0x41, 0x42, 0x43, 0x44, 0x45, 0x46, 0x47,
+  0x48, 0x49, 0x4A, '.',  0x4C, 0x4D, 0x4E, 0x4F,
+  /* 0x50-0x5F */
+  '&',  0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57,
+  0x58, 0x59, 0x5A, '$',  0x5C, 0x5D, 0x5E, 0x5F,
+  /* 0x60 = EBCDIC minus/hyphen */
+  '-',  0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67,
+  0x68, 0x69, 0x6A, 0x6B, '%',  0x6D, 0x6E, 0x6F,
```
bradh352: **[Medium] EBCDIC `0x6B` (comma) maps to `'k'`, not `','`.**

In CCSID 37, `0x6B` is a comma. The table passes it through as `0x6B` = `'k'`, so a comma in the source would corrupt the string — and the search-list splitter uses `", "` as its delimiter set yet the table can never *produce* a comma. Lower impact once the search list splits on spaces, but the table is presented as a full EBCDIC->ASCII map and isn't. Either map `0x6B -> ','` (and fix the other non-hostname punctuation slots: `0x4A 0x4C 0x4D 0x4E 0x5C 0x61 0x7A 0x7C` are also identity/wrong), or add a comment that only the DNS/hostname charset is guaranteed correct.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+    sysconfig->usevc = ARES_TRUE;
+  }
+
+  /* Configure lookup order (search_order: 1=local first, 2=remote first)
+   * This maps to the "lookup" option in resolv.conf */
+  if (header2->search_order == 1) {
+    sysconfig->lookups = ares_strdup("fb");  /* files, bind */
+  } else {
+    sysconfig->lookups = ares_strdup("bf");  /* bind, files */
+  }
```
bradh352, line 403: **[Low] `lookups` is set unconditionally (unlike every other field) and the `ares_strdup` return isn't checked.**

The mapping direction is correct (1=Local -> `"fb"`, 2=Remote -> `"bf"`, verified against IBM docs). But unlike `timeout_ms`/`tries`/`rotate`/`usevc`, this writes on every path via the `else`, so a `search_order` of 0/unset silently flips c-ares's `"fb"` default to `"bf"`. Prefer explicit branches, and NULL-check the allocation for consistency with the `domains` path:

```c
  if (header2->search_order == 1) {
    sysconfig->lookups = ares_strdup("fb");  /* files, bind */
  } else if (header2->search_order == 2) {
    sysconfig->lookups = ares_strdup("bf");  /* bind, files */
  }
```

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+  /* Re-initialize if we're in a different process (after fork) */
+  if (qtocrtvtcpa_initialized && cached_pid == current_pid) {
+    pthread_mutex_unlock(&ile_mutex);
+    return ARES_SUCCESS;
+  }
+
+  /* Load or reload the ILE service program */
+  actmark = _ILELOADX("QSYS/QTOCNETSTS", ILELOAD_LIBOBJ);
+  if (actmark == (unsigned long long)-1) {
+    status = ARES_ELOADIPHLPAPI;
```
bradh352: **[Low] Reusing `ARES_ELOADIPHLPAPI` (a Windows "load iphlpapi.dll" error) for an ILE load failure is misleading.**

It's functionally inert — this status never reaches a caller, since a failed `load_ile_api()` is immediately converted into the `ares_init_sysconfig_files()` fallback — but it reads oddly on IBM i. Since the value is discarded, consider having `load_ile_api()` return a plain `ares_bool_t`, or use a neutral code.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,303 @@
...
+      if (sysconfig->domains == NULL) {
+        ares_free(buffer);
+        return ARES_ENOMEM;
+      }
+    }
+  }
+
+  /* Configure timeout (time_interval is in seconds) */
+  if (header2->time_interval > 0) {
+    sysconfig->timeout_ms = header2->time_interval * 1000;
```
bradh352: **[Low/Nit] `time_interval * 1000` is computed in `int` before the `size_t` store.**

A large `time_interval` would overflow (signed overflow = UB). The API caps it at 0–99 so it's not reachable in practice, but cast first for cleanliness: `sysconfig->timeout_ms = (size_t)header2->time_interval * 1000;`.

src/lib/ares_sysconfig_pase.c
bradh352: I dug into the IBM `QtocRtvTCPA` TCPA1400 documentation on this. The DNS list is **count-driven** — "Number of DNS Internet addresses" (`dns_list_count`) is authoritative — and I couldn't find anything documenting a fixed-10-slot list padded with `0.0.0.0`/`::`. That matches @kadler's observation that `dns_list_count` is always < 10 in practice. Since the loop is already bounded by `dns_list_count`, unused padding slots (if they existed) would never be read anyway, so the filter is effectively dead but harmless.

Two options: drop the `0.0.0.0`/`::` filter entirely and rely on `dns_list_count`, or keep it purely as defensive belt-and-suspenders — but in that case please reword the comment, since claiming the API pads to 10 slots isn't supported by the docs and could mislead future maintainers. My preference is to remove it and trust the count. Either way it shouldn't block the substantive fixes above.

sravani1510: Removed the `0.0.0.0/::` filter and the `&& i < 10` cap entirely. The loop now simply relies on dns_list_count as the authoritative count:
`for (i = 0; i < header2->dns_list_count; i++)`

src/lib/ares_sysconfig_pase.c
sravani1510: Added the standard MIT license block with `SPDX-License-Identifier: MIT` at the top of the file, matching the format used by all other c-ares source files

src/lib/ares_sysconfig_pase.c
sravani1510: Replaced the truncating copy loop with `ebcdic_to_ascii_str()` which converts the whole `search_list` field at once into `char search_ascii[sizeof(header2->search_list) + 1]` (the `+1` fixes the stack overflow). Since EBCDIC `0x40` maps to ASCII space in the table, `ares_strsplit(search_ascii, ", ", …)` correctly splits all space-separated domains and automatically drops the trailing blank padding as empty tokens. The `ebcdic_to_ascii_str()` helper is also reused for the `domain_name` fallback.

src/lib/ares_sysconfig_pase.c
sravani1510: Modified as per the suggested change

src/lib/ares_sysconfig_pase.c
sravani1510: All declarations (`err_code, signature[], arglist, format[]`) are now hoisted to the very top of the function before any executable statements, in compliance with C89/C90.

src/lib/ares_sysconfig_pase.c
sravani1510: 1. `additional_info_offset` is validated (non-negative and fits within `buflen`) before `header2` is derived from it.
2. `dns_list_entry_size` is checked against `sizeof(dns_list_item_t)` before entering the loop.
3. Per-entry offset `off` is computed with full `size_t` casts and checked against `buflen` before each item is accessed, breaking out of the loop if it would read past the buffer.

src/lib/ares_sysconfig_pase.c
sravani1510: After `inet_ntop`, if `dns_listening_port` is non-zero and not 53, the server is formatted as `[ip]:port` using `snprintf` before being passed to `ares_sconfig_append_fromstr`. For port 0 or 53 the IP is passed as-is. The `ip_str` buffer is sized `INET6_ADDRSTRLEN + 10` to safely hold the brackets and port suffix.

src/lib/ares_sysconfig_pase.c
sravani1510: After the search list block, if `ndomains == 0` and `domain_name` is non-empty and non-blank, `ebcdic_to_ascii_str()` converts it to ASCII and `ares_strsplit` produces a single-entry domain list from it, mirroring resolv.conf domain semantics. The `ebcdic_to_ascii_str` helper is reused here exactly as suggested. A NULL-check on the allocation is also included.

src/lib/ares_sysconfig_pase.c
sravani1510: `0x6B` is now correctly mapped to `','` in the table. A comment has also been added making clear that only the DNS/hostname charset (letters, digits, hyphen, dot, space, comma) is guaranteed fully correct, other slots map identity or to a safe fallback. The comment explicitly calls out `0x6B -> ','` as a key correction vs a naive identity table.

src/lib/ares_sysconfig_pase.c
sravani1510: Replaced the bare `else` with an explicit `else if` (`search_order == 2`) so `search_order` of 0 or any unrecognised value leaves `lookups` unchanged preserving c-ares's default. Both branches now also NULL-check `ares_strdup` and return `ARES_ENOMEM` on failure, consistent with the `domains` path.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+
+#  include <sys/socket.h>
+#  include <netinet/in.h>
+#  include <arpa/inet.h>
+#  include <as400_protos.h>
+#  include <pthread.h>
+#  include <unistd.h>
+
+/* EBCDIC space character (CCSID 37: 0x40 = space) */
+#define EBCDIC_SPACE '\x40'
```
bradh352: **[Nit] clang-format:** inside the `#ifdef __PASE__` this needs to be `#  define` (the file sets `IndentPPDirectives: AfterHash` — see the `#  include` block above and `ares_sysconfig_win.c`). CI checks the whole new file, so expect it to also reflow the EBCDIC table's manual 8-per-row alignment. Just run `git clang-format` before pushing.

```suggestion
#  define EBCDIC_SPACE '\x40'
```

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+  int  dns_list_entry_size;
+  int  dns_protocol;
+  int  retries;
+  int  time_interval;
+  int  search_order;
+  int  initial_server;
+  int  dns_listening_port;
+  char hostname[64];
+  char domain_name[255];
+  char reserved;
```
bradh352: **[Nit] Document the alignment-dependent overlay.** This struct is byte-exact against IBM's TCPA1400 format only because `domain_name[255]` + this `reserved[1]` sum to 256, which lands `request_dnssec` at offset 612 (4-aligned) with **no compiler-inserted padding**. A short comment here would stop a future cleanup from "tidying" `char reserved;` and silently breaking the layout.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+  0x38, 0x39, 0x3A, 0x3B, 0x3C, 0x3D, 0x3E, 0x3F,
+  /* 0x40 = EBCDIC space -> ASCII space */
+  ' ',  0x41, 0x42, 0x43, 0x44, 0x45, 0x46, 0x47,
+  0x48, 0x49, 0x4A, '.',  0x4C, 0x4D, 0x4E, 0x4F,
+  /* 0x50-0x5F */
+  '&',  0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57,
+  0x58, 0x59, 0x5A, '$',  0x5C, 0x5D, 0x5E, 0x5F,
+  /* 0x60 = EBCDIC hyphen/minus -> ASCII hyphen */
+  '-',  0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67,
+  0x68, 0x69, 0x6A, ',',  '%',  0x6D, 0x6E, 0x6F,
```
bradh352: **[Nit] EBCDIC `0x6D` (underscore) maps to identity `'m'`, should be `'_'`.** In CCSID 37 `0x6D` is `_`; passing it through as `0x6D`=`'m'` would corrupt an underscore in a search/primary domain to `m`. The header comment scopes the guarantee to letters/digits/hyphen/dot/space/comma so it's technically documented, but it's a one-character fix (and `_`, like the rest of that set, is in the EBCDIC *invariant* range, so it's correct regardless of the job CCSID):

```suggestion
  0x68, 0x69, 0x6A, ',',  '%',  '_',  0x6E, 0x6F,
```

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+  pthread_mutex_unlock(&ile_mutex);
+  return ok;
+}
+
+ares_status_t ares_init_sysconfig_pase(const ares_channel_t *channel,
+                                       ares_sysconfig_t     *sysconfig)
+{
+  /* --- All declarations at top of function (C89/C90 compliance) --- */
+  ares_status_t status;
+  unsigned int  buflen;
```
bradh352: **[Nit]** `buflen` is `unsigned int`, but IBM's length parameter is a signed `BINARY(4)`. Use `int` (or add a note) to match the API contract.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+  char format[9] = "\xe3\xc3\xd7\xc1\xf1\xf4\xf0\xf0";
+
+  /* Load ILE API; fall back to file-based config if unavailable */
+  if (!load_ile_api()) {
+    return ares_init_sysconfig_files(channel, sysconfig, ARES_TRUE);
+  }
+
+  /* Allocate buffer for the API response */
+  buflen = sizeof(tcpa1100_t) + sizeof(tcpa1400_t) +
+           sizeof(dns_list_item_t) * 10 + 512;
```
bradh352: **[Nit] Magic `10` and unexplained `+ 512`.** The per-entry loop is already bounds-checked so over/under-sizing is safe — this is purely for clarity:

```c
#define ARES_PASE_MAX_DNS_SERVERS 10   /* IBM i allows at most this many */
...
  buflen = sizeof(tcpa1100_t) + sizeof(tcpa1400_t) +
           sizeof(dns_list_item_t) * ARES_PASE_MAX_DNS_SERVERS + 512; /* header slack */
```
A word on why `+512` would help too.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+
+  buffer = ares_malloc(buflen);
```
bradh352: **[Low] Prefer `ares_malloc_zero()`** over `ares_malloc()` + `memset()`, per the house memory convention (AGENTS.md):

```suggestion
  buffer = ares_malloc_zero(buflen);
  if (buffer == NULL) {
    return ARES_ENOMEM;
  }
```
(then drop the separate `memset(buffer, 0, buflen);` below.)

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+  err_code.bytes_provided = sizeof(err_code);
+
+  /* Wire up argument pointers for the ILE call */
+  arglist.buffer.s.addr  = (address64_t)(intptr_t)&buffer[0];
+  arglist.buflen.s.addr  = (address64_t)(intptr_t)&buflen;
+  arglist.format.s.addr  = (address64_t)(intptr_t)&format[0];
+  arglist.errcode.s.addr = (address64_t)(intptr_t)&err_code;
+
+  /* Call QtocRtvTCPA ILE API */
+  rc = _ILECALL(&qtocrtvtcpa_ptr, &arglist.base, signature, RESULT_VOID);
```
bradh352, line 279: **[Low]** `qtocrtvtcpa_ptr` is read here by `_ILECALL` outside `ile_mutex` — formally an unsynchronized read of the global. Inert in practice (the child is single-threaded post-fork and the pointer is stable once initialized), but worth a note.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+    if (inet_ntop(af, item->ip_address, ip, sizeof(ip)) == NULL) {
+      continue;
+    }
+
+    /* Apply non-standard port if configured (B6) */
+    if (header2->dns_listening_port != 0 && header2->dns_listening_port != 53) {
+      snprintf(ip_str, sizeof(ip_str), "[%s]:%d", ip,
+               header2->dns_listening_port);
+    } else {
+      snprintf(ip_str, sizeof(ip_str), "%s", ip);
```
bradh352: **[Nit]** For the no-port case this is a plain copy; `ares_strcpy(ip_str, ip, sizeof(ip_str))` is the idiom here (minor).

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,390 @@
...
+    if (sysconfig->lookups == NULL) {
+      ares_free(buffer);
+      return ARES_ENOMEM;
+    }
+  }
+
+  ares_free(buffer);
+
+  /* If no DNS servers were found, fall back to file-based config */
+  if (ares_llist_len(sysconfig->sconfig) == 0) {
```
bradh352, line 413: **[Low] Empty-servers fallback lets API-derived resolver settings linger.** By the time we reach this fallback, `timeout_ms`/`tries`/`rotate`/`usevc`/`lookups`/`domains` set from the API remain and mix with whatever `ares_init_sysconfig_files()` then reads from resolv.conf. There's **no leak** (the files parser frees-before-set), but the blend is surprising — e.g. a resolv.conf with no `options timeout` keeps the API timeout. Suggest either zeroing those resolver fields before falling back, or a one-line comment documenting that the mix is intended.

bradh352 (commented):
Thanks for turning this around so thoroughly — I re-reviewed the current head (`44a2190`) end-to-end and **all 17 of the previous inline comments are correctly addressed**. I also re-verified the substantive machinery against primary sources and it holds up well:

- **Structs vs IBM docs:** re-checked `tcpa1100_t`/`tcpa1400_t`/`dns_list_item_t` against IBM's official 7.4 `QtocRtvTCPA` (QTOCNETSTS) TCPA1400/TCPA1100 format tables — they match **field-for-field, offset-for-offset**, including the single `reserved` byte before `search_list` and the trailing `request_dnssec`. The overlay is byte-exact with no compiler padding.
- **Semantics:** `retries` is *additional* attempts (0–99, default 2) so `tries = retries + 1` is correct; `search_order` 1=Local/2=Remote, `version` 1=IPv4/2=IPv6, port range 1–65532, and the "up to six, space-separated, blank-padded" search list are all confirmed against the docs.
- **c-ares plumbing:** `[ipv4]:port` bracket form is accepted by `ares_sconfig_append_fromstr`; `ares_strsplit` drops the blank padding (no `ALLOW_BLANK`); no leaks on the fallback paths; no NULL-deref on the empty-servers path; dispatch order in `ares_sysconfig.c` is correct.
- **AGENTS.md rule 3 (no manual parsing):** I considered whether the offset/pointer walk should go through `ares_buf`. I'm satisfied it's fine here — this is a byte-exact overlay of a *trusted, fixed-layout OS API result*, exactly how `ares_sysconfig_win.c`/`ares_sysconfig_mac.c` consume their OS structs; `ares_buf` targets untrusted wire/text input. **No change requested.**

The remaining items are inline (all Low/Nit) plus one thing I'd still like resolved before merge:

### [Major] No tests, and the file is invisible to CI
The whole translation unit is under `#ifdef __PASE__` and there's no PASE CI job, so **nothing compiles it** — a field-name/type/symbol typo would only surface on a real IBM i build, and our contribution policy requires tests (>90% coverage). The good news: the pure logic — the EBCDIC table / `ebcdic_to_ascii_str()` and the space-separated search-list split — has zero PASE dependencies. Lifting it out of the `#ifdef` (or into a small always-compiled helper) and unit-testing it in `test/ares-test-internal.cc` (wrapped in `#ifndef CARES_SYMBOL_HIDING`) would both satisfy the test requirement *and* give CI a real compile check of the parsing logic. That's the main thing I'd want before we merge; the table-driven test would also have caught the `0x6D` underscore nit below.

### Process
The description points at `nodejs/node#53948` but has no `Fixes:`/`Closes:` line — and since it's a cross-repo issue it won't auto-close anything here regardless. Fine to leave; just noting.

Really nice work overall — this is close.

src/lib/ares_sysconfig_pase.c
sravani1510: `#define EBCDIC_SPACE` is now `# define EBCDIC_SPACE` (matching `IndentPPDirectives: AfterHash`), consistent with the `# include` block above it.

src/lib/ares_sysconfig_pase.c
sravani1510: Added a block comment above the struct explaining why the `reserved` byte must not be removed, `domain_name[255] + reserved[1] = 256` keeps `request_dnssec` at offset 612 (4-byte aligned, zero compiler padding), plus an inline comment on the field itself.

src/lib/ares_sysconfig_pase.c
sravani1510: `0x6D` now maps to '\_' instead of identity `0x6D`. The header comment has also been updated to include `0x6D` -> '_' (underscore) in the key corrections list.

src/lib/ares_sysconfig_pase.c
sravani1510: `buflen` is now declared as `int` (signed) to match IBM's BINARY(4) API contract, and the size calculation is cast accordingly: `buflen = (int)(sizeof(...) + ...)`

src/lib/ares_sysconfig_pase.c
sravani1510: extracted `ARES_PASE_MAX_DNS_SERVERS 10` as a named constant with a comment explaining it is an upper bound for buffer sizing (IBM docs don't document a hard cap; `dns_list_count` is trusted as authoritative by the loop). Also added a comment explaining the +512: the buffer layout is `tcpa1100_t` at offset 0 followed by `tcpa1400_t` at `additional_info_offset` (API-reported, may exceed `sizeof(tcpa1100_t)` due to alignment or future header growth) — the `+512` covers that gap. The bounds checks below ensure we never read past `buflen` regardless.

src/lib/ares_sysconfig_pase.c
sravani1510: Replaced `ares_malloc() + memset(buffer, 0, buflen)` with `ares_malloc_zero((size_t)buflen)`
and the separate `memset` is removed.

src/lib/ares_sysconfig_pase.c
sravani1510: Added a comment above the `_ILECALL` noting that `qtocrtvtcpa_ptr` is read outside `ile_mutex`

src/lib/ares_sysconfig_pase.c
sravani1510: Replaced `snprintf(ip_str, sizeof(ip_str), "%s", ip)` with `ares_strcpy(ip_str, ip, sizeof(ip_str))` for the no-port case.

src/lib/ares_sysconfig_pase.c
sravani1510: Added a comment documenting that the blend is intentional: API-configured settings (timeout_ms, tries, rotate, usevc, lookups, domains) take precedence and the file-based config fills in any gaps. The files parser frees-before-set so there is no leak.

sravani1510:
hey @bradh352 could you please re-review this, thanks

bradh352 (commented):
Third round. I built and tested this end to end; the substance is in good shape and the two remaining items are both about the tests and naming rather than the IBM i logic.

**All 9 open review threads from the previous round are correctly addressed** — the `#  define` indentation, the struct-overlay comment, `0x6D` -> `_`, `buflen` as a signed `int`, the `ARES_PASE_MAX_DNS_SERVERS` constant and the `+ 512` explanation, `ares_malloc_zero()`, the note about `qtocrtvtcpa_ptr` being read outside the mutex, `ares_strcpy()` for the no-port case, and the comment on the empty-servers fallback. I can resolve those threads.

### What I verified

- **The character table is correct where it claims to be.** I dumped `ares__ebcdic_to_ascii_table` out of the built library and compared all 256 entries against a reference coded character set identifier 37 (CCSID 37) codec. Every character the file's comment scopes as guaranteed — letters, digits, hyphen, period, space, comma, underscore — matches exactly, with zero mismatches. The 38 slots that differ are all outside that documented scope, which is the arrangement we settled on last round.
- **The `#ifdef __PASE__` body compiles clean.** I built it on this machine with `__PASE__` defined and stub Integrated Language Environment headers, under `-std=gnu90` with the exact warning list c-ares uses for `src/lib`. No `-Wconversion`, `-Wsign-conversion`, or `-Wdeclaration-after-statement` problems. Only `-Wcast-align` fires, which is the Low finding inline.
- **Builds and tests.** Debug with `-DCARES_WERROR=ON`, and a symbol-hiding shared build (`-DCARES_SYMBOL_HIDING=ON -DCARES_STATIC=OFF`) — both clean, so the `#ifndef CARES_SYMBOL_HIDING` guard around the tests is right. All five new tests pass, including under AddressSanitizer and UndefinedBehaviorSanitizer.

### Findings

2 Medium, 1 Low, 2 Nit — all inline.

The one I would like resolved before merge is the test comment: `EbcdicSearchListSplit` and `EbcdicSearchListAllBlanks` build ASCII string literals and call `ares_strsplit()`, so they test pre-existing c-ares code rather than anything this pull request adds. I confirmed by re-introducing last round's truncation bug into `ares__ebcdic_to_ascii_str()` — all five tests still passed. There is a drop-in replacement in that comment that fails on the mutation and passes on the current code.

### Please rebase onto `main` before this goes further

The branch is 41 commits behind. I checked and **it rebases cleanly — zero conflicts** — so this should be cheap. It matters more than usual here:

- **The `clang-format` continuous integration job landed on `main` after your branch point, so it has never run against this pull request.** It checks `**.c` and `**.h` on the lines a pull request changes; `ares_sysconfig_pase.c` is entirely new, so all 420 lines count as changed. I confirmed `ares_private.h` and `ares_sysconfig.c` are still not whole-file clean on `main` either, so their changed lines get checked too. Commit `a13b89cc` also bulk-reformatted `src/` and `include/` in the meantime.
- **`AGENTS.md` landed after your branch point.** It is the short written-down version of the rules this review keeps citing — C89, the memory conventions, tests required, clang-format. Worth reading once you have it.
- **`ares_sysconfig_domains_idna()` now post-processes search domains for every platform.** It came in with `380be0d3` (punycode and Internationalized Domain Names support) and is called in `ares_init_by_sysconfig()` *after* the platform dispatch, so the domains this file sets from `search_list` and `domain_name` would get IDNA-encoded automatically once rebased. That is almost certainly what you want on IBM i, but it is new behavior your branch has never run, so please sanity-check it.

Rebasing also means the current continuous integration job set actually runs against the change, rather than the set that existed at your branch point.

### One thing that cannot be anchored to a line

**The rest of the file is still compiled by no continuous integration job.** Lifting the table and the string helper out of the `#ifdef` did what I asked last round and gives us a real compile check plus tests for the parsing logic — thank you. But roughly 300 lines inside `#ifdef __PASE__` (the offset validation, the port formatting, the `retries` and `search_order` mapping) are still built only on a real IBM i. That is inherent to platform code and I am not asking for more here; I am noting it so the record is clear about what the tests do and do not cover. It is also why the `-Wcast-align` item is worth taking even though it is only a warning.

On process: the description still points at `nodejs/node#53948` — the Node.js issue reporting that DNS resolution fails on IBM i — with no `Fixes:` or `Closes:` line. It is a cross-repository issue so it would not auto-close anything here in any case; leaving it as-is is fine.

src/lib/ares_private.h
```diff
@@ -399,6 +399,15 @@ ares_status_t ares_init_sysconfig_files(const ares_channel_t *channel,
...
                                         ares_sysconfig_t     *sysconfig);
 #endif
+#ifdef __PASE__
+ares_status_t ares_init_sysconfig_pase(const ares_channel_t *channel,
+                                       ares_sysconfig_t     *sysconfig);
+#endif
+
+/* EBCDIC <-> ASCII helpers compiled on all platforms (used in unit tests) */
+extern const unsigned char ares__ebcdic_to_ascii_table[256];
+void ares__ebcdic_to_ascii_str(const char *ebcdic, size_t len, char *out);
```
bradh352: **[Medium] `ares__` double-underscore prefix — the project removed it everywhere.**

These two are the only `ares__` symbols left in the tree. I checked: `grep -rn "ares__" src include test` matches nothing else. The prefix was deliberately removed project-wide in commit `f09f0a02` ("Remove `ares__` prefix, just use `ares_`", PR #872), so re-introducing it here would be the one exception.

Please rename to a single underscore. The call sites are `src/lib/ares_sysconfig_pase.c` lines 47, 102, 106, 342 and 359, plus the five uses in `test/ares-test-internal.cc`.

Also, the comment says "EBCDIC <-> ASCII helpers" but only the EBCDIC-to-ASCII direction exists.

```suggestion
/* Extended Binary Coded Decimal Interchange Code (EBCDIC) to ASCII helpers.
 * Compiled on all platforms so the conversion logic is unit-testable off
 * IBM i. */
extern const unsigned char ares_ebcdic_to_ascii_table[256];
void ares_ebcdic_to_ascii_str(const char *ebcdic, size_t len, char *out);
```

test/ares-test-internal.cc
```diff
@@ -1928,5 +1928,94 @@ const struct ares_socket_functions VirtualizeIO::default_functions = {
...
+}
+
+/* Verify that an all-blank (empty) search list produces NULL / 0 domains */
+TEST_F(LibraryTest, EbcdicSearchListAllBlanks) {
+  const char blank_field[] = "              "; /* all spaces */
+  size_t n = 99;
+  char **parts = ares_strsplit(blank_field, ", ", &n);
+  EXPECT_EQ(nullptr, parts);
+  EXPECT_EQ((size_t)0, n);
+}
```
bradh352: **[Medium] These two tests do not exercise any code this pull request adds.**

Both build a plain ASCII string literal and hand it to `ares_strsplit()`. `ares_strsplit()` is pre-existing c-ares code — neither test ever calls `ares__ebcdic_to_ascii_str()` or reads the conversion table, so neither one can fail because of a bug in this pull request. The comment on line 1997 says as much ("we test with a simulated already-converted ASCII string").

I verified this is not theoretical. I re-introduced the exact truncation bug from the previous review round — stopping the conversion loop at the first EBCDIC blank (`0x40`) instead of converting the whole field, which is what silently reduced a multi-domain search list to one domain — and rebuilt. All five tests still passed. With the replacement below, `EbcdicSearchListSplit` and `EbcdicSearchListAllBlanks` both fail on that same mutation and both pass on the current code.

The fix is to feed the helper a real Extended Binary Coded Decimal Interchange Code (EBCDIC) blank-padded field, exactly as `ares_init_sysconfig_pase()` does, and then split the result.

Two smaller things folded into the same replacement: `using std::string;` / `using std::vector;` on lines 1993-1994 are unused, and the comment on line 1996 says the literal is "blank-padded to 32 bytes" while line 2000 says "48 chars".

I built and ran this under a plain debug build and under AddressSanitizer plus UndefinedBehaviorSanitizer; clean in both.

```suggestion
/* Convert a whole EBCDIC, space-separated, blank-padded search_list field and
 * split it, exactly as ares_init_sysconfig_pase() does.  Exercises the table's
 * 0x40 -> ' ' mapping and the whole-field (non-truncating) conversion. */
TEST_F(LibraryTest, EbcdicSearchListSplit) {
  /* "corp.example.com eng.example.com" in CCSID 37, blank(0x40)-padded */
  const char ebcdic_field[48] = {
    '\x83', '\x96', '\x99', '\x97', '\x4b', '\x85', '\xa7', '\x81',
    '\x94', '\x97', '\x93', '\x85', '\x4b', '\x83', '\x96', '\x94',
    '\x40', '\x85', '\x95', '\x87', '\x4b', '\x85', '\xa7', '\x81',
    '\x94', '\x97', '\x93', '\x85', '\x4b', '\x83', '\x96', '\x94',
    '\x40', '\x40', '\x40', '\x40', '\x40', '\x40', '\x40', '\x40',
    '\x40', '\x40', '\x40', '\x40', '\x40', '\x40', '\x40', '\x40'
  };
  char   ascii[sizeof(ebcdic_field) + 1];
  size_t n     = 0;
  char **parts = nullptr;

  ares__ebcdic_to_ascii_str(ebcdic_field, sizeof(ebcdic_field), ascii);
  EXPECT_STREQ("corp.example.com eng.example.com                ", ascii);

  parts = ares_strsplit(ascii, ", ", &n);
  ASSERT_NE(nullptr, parts);
  ASSERT_EQ((size_t)2, n);
  EXPECT_STREQ("corp.example.com", parts[0]);
  EXPECT_STREQ("eng.example.com", parts[1]);
  ares_strsplit_free(parts, n);
}

/* An all-blank (EBCDIC 0x40) search_list field must yield no domains. */
TEST_F(LibraryTest, EbcdicSearchListAllBlanks) {
  char   ebcdic_field[16];
  char   ascii[sizeof(ebcdic_field) + 1];
  size_t n = 99;
  char **parts;

  memset(ebcdic_field, 0x40, sizeof(ebcdic_field));
  ares__ebcdic_to_ascii_str(ebcdic_field, sizeof(ebcdic_field), ascii);
  EXPECT_STREQ("                ", ascii);

  parts = ares_strsplit(ascii, ", ", &n);
  EXPECT_EQ(nullptr, parts);
  EXPECT_EQ((size_t)0, n);
}
```

(If you take the `ares__` -> `ares_` rename from the other comment, adjust the two helper calls above to match.)

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,420 @@
...
+   * unsynchronized read of the global. */
+  rc = _ILECALL(&qtocrtvtcpa_ptr, &arglist.base, signature, RESULT_VOID);
+
+  if (rc != ILECALL_NOERROR || err_code.bytes_available) {
+    ares_free(buffer);
+    return ares_init_sysconfig_files(channel, sysconfig, ARES_TRUE);
+  }
+
+  /* --- Validate outer header offset before use (B5) --- */
+  header = (tcpa1100_t *)buffer;
```
bradh352: **[Low] Three casts trip `-Wcast-align`, which is in the project's standard warning set.**

I compiled this file with `__PASE__` defined against stub Integrated Language Environment (ILE) headers, using the exact flag list c-ares builds `src/lib` with (taken from the generated build rules) plus `-Werror`. It is otherwise clean — no `-Wconversion`, `-Wsign-conversion` or `-Wdeclaration-after-statement` complaints under `-std=gnu90` — but these three casts fail:

```
ares_sysconfig_pase.c:287: error: cast from 'char *' to 'tcpa1100_t *' increases required alignment from 1 to 4 [-Werror,-Wcast-align]
ares_sysconfig_pase.c:294: error: cast from 'char *' to 'tcpa1400_t *' increases required alignment from 1 to 4 [-Werror,-Wcast-align]
ares_sysconfig_pase.c:312: error: cast from 'char *' to 'dns_list_item_t *' increases required alignment from 1 to 4 [-Werror,-Wcast-align]
```

Whether this actually bites on IBM i depends on the toolchain: gcc only issues `-Wcast-align` on targets it considers strict-alignment, and PowerPC is not one, so a gcc build there is likely silent today. It is worth fixing anyway — no continuous integration job compiles this block, so nothing else will tell you if that changes.

The house idiom for exactly this is an intermediate `void *` cast, used at `src/lib/ares_conn.c:98` and `:109`. I confirmed all three warnings go away with it and the file then compiles clean under the full flag set with `-Werror`.

```suggestion
  header = (tcpa1100_t *)(void *)buffer;
```

Line 294 becomes `header2 = (tcpa1400_t *)(void *)(buffer + header->additional_info_offset);` and line 312 becomes `item = (dns_list_item_t *)(void *)(buffer + off);`.

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,420 @@
...
+/* Thread-safe cache for ILE procedure pointer with fork detection */
+static ILEpointer      qtocrtvtcpa_ptr __attribute__((aligned(16)));
+static int             qtocrtvtcpa_initialized = 0;
+static pid_t           cached_pid              = 0;
+static pthread_mutex_t ile_mutex               = PTHREAD_MUTEX_INITIALIZER;
+
+/* Thread-safe initialization of ILE API with fork detection.
+ * ILE pointers are invalidated in a child process after fork(), so we
+ * cache the PID and re-initialize whenever it changes. */
+static ares_bool_t load_ile_api(void)
```
bradh352: **[Nit] clang-format wants an extra space run here.**

This is the only line in the file that is not clang-format clean. Running `clang-format` (I used 22.1.8) over the whole file produces exactly this one change — the preceding `static` variable block anchors the declaration-alignment column.

Worth fixing because the branch predates the `clang-format` continuous integration job, which landed on `main` after your branch point and has therefore never run against this pull request. It checks `**.c` and `**.h` on the lines a pull request changes; since this whole file is new, every line counts as changed.

```suggestion
static ares_bool_t     load_ile_api(void)
```

src/lib/ares_sysconfig_pase.c
```diff
@@ -0,0 +1,420 @@
...
+  int bytes_available;
+  int ipv6_status;
+  int additional_info_offset;
+  int additional_info_length;
+  int ipv4_status;
+} tcpa1100_t;
+
+/* Byte-exact overlay of IBM TCPA1400 (QtocRtvTCPA format TCPA1400).
+ * domain_name[255] + reserved[1] = 256 bytes; this lands request_dnssec at
+ * offset 612 (4-byte aligned) with zero compiler padding — do not reorder or
```
bradh352: **[Nit] Non-ASCII characters in the comments.**

Lines 136, 150, 276, 277, 338 and 375 contain a UTF-8 em dash (U+2014). Excluding third-party code and the generated `ares_idnamap.c`, this is the only file under `src/lib` with any byte outside plain ASCII — I checked the whole directory.

Several of these came in from my own earlier review comments, so this is partly my doing. Still worth swapping for a plain `-` or `--`, particularly in this file: it is the one that gets carried onto a system where source may pass through a coded character set conversion.

src/lib/ares_private.h
sravani1510: Renamed both symbols to single underscore everywhere:

- `ares__ebcdic_to_ascii_table` → `ares_ebcdic_to_ascii_table`
- `ares__ebcdic_to_ascii_str` → `ares_ebcdic_to_ascii_str`

Updated all call sites across all three files (`ares_private.h`, `ares_sysconfig_pase.c`, `ares-test-internal.cc`). Also corrected the comment from `"EBCDIC <-> ASCII helpers"` to `"EBCDIC to ASCII helpers"` since only the one direction exists.

Verified: `grep -rn "ares__" src include test` returns no matches.

test/ares-test-internal.cc
sravani1510: Both tests now feed real EBCDIC byte arrays through `ares_ebcdic_to_ascii_str()` before splitting, exactly as `ares_init_sysconfig_pase()` does. Also applied the `ares__` → `ares_` rename to the two helper calls as noted.

Also removed the unused `using std::string` / `using std::vector` and fixed the "32 bytes" vs "48 chars" comment inconsistency.

src/lib/ares_sysconfig_pase.c
sravani1510: Added `(void *)` intermediate cast to all three sites:

```c
header  = (tcpa1100_t *)(void *)buffer;
header2 = (tcpa1400_t *)(void *)(buffer + header->additional_info_offset);
item    = (dns_list_item_t *)(void *)(buffer + off);
```

src/lib/ares_sysconfig_pase.c
sravani1510: Added the extra spaces to align with the preceding `static` variable block.

src/lib/ares_sysconfig_pase.c
sravani1510: Replaced all 6 em dashes with `--` on lines 136, 150, 276, 277, 338, and 375.

sravani1510:
hey @bradh352, Thanks for all the suggestions and I Addressed all third-round review findings and rebased onto `main`. Could you please review this, thanks

PR #1168, opened 2026-06-11
Guard DNS record binary length overflow

Alb3e3:
## Summary
- validate null-terminated DNS record binary allocation sizes before adding the trailing NUL byte
- reject nonzero copy lengths with NULL input pointers in the public binary/option setters
- add regression coverage for SIZE_MAX lengths across TXT ABINP, SIG BINP, and OPT option paths

## Testing
- `/tmp/cmake-3.29.6-linux-x86_64/bin/cmake -S . -B build -G "Unix Makefiles" -DCARES_BUILD_TESTS=OFF -DCARES_STATIC=ON -DCARES_SHARED=OFF`
- `/tmp/cmake-3.29.6-linux-x86_64/bin/cmake --build build -j2`
- `/tmp/cmake-3.29.6-linux-x86_64/bin/cmake -S . -B build-test -G "Unix Makefiles" -DCARES_BUILD_TESTS=ON -DCARES_STATIC=ON -DCARES_SHARED=OFF -DCMAKE_PREFIX_PATH=/tmp/googletest-cares-install`
- `/tmp/cmake-3.29.6-linux-x86_64/bin/cmake --build build-test -j2 --target arestest`
- `./build-test/bin/arestest --gtest_filter=LibraryTest.DNSRecordRejectsInvalidBinaryInput`
- `./build-test/bin/arestest --gtest_filter=LibraryTest.DNSRecord*:LibraryTest.DNSParseFlags`
- `./build-test/bin/arestest --gtest_filter=-*.Live*:*Live*` (1063 tests passed)

### DIFF
--- a/src/lib/record/ares_dns_record.c
+++ b/src/lib/record/ares_dns_record.c
@@ -1102,13 +1102,22 @@ ares_status_t ares_dns_rr_add_abin(ares_dns_rr_t *dns_rr, ares_dns_rr_key_t key,
   ares_dns_datatype_t datatype = ares_dns_rr_key_datatype(key);
   ares_bool_t         is_nullterm =
     (datatype == ARES_DATATYPE_ABINP) ? ARES_TRUE : ARES_FALSE;
-  size_t                   alloclen = is_nullterm ? len + 1 : len;
+  size_t                   alloclen;
   unsigned char           *temp;
   ares_dns_multistring_t **strs;
 
   if (ares_dns_rr_key_datatype(key) != ARES_DATATYPE_ABINP) {
     return ARES_EFORMERR;
   }
+
+  if (val == NULL && len != 0) {
+    return ARES_EFORMERR;
+  }
+
+  if (is_nullterm && len == SIZE_MAX) {
+    return ARES_ENOMEM;
+  }
+  alloclen = is_nullterm ? len + 1 : len;
 
   strs = ares_dns_rr_data_ptr(dns_rr, key, NULL);
   if (strs == NULL) {
@@ -1127,7 +1136,9 @@ ares_status_t ares_dns_rr_add_abin(ares_dns_rr_t *dns_rr, ares_dns_rr_key_t key,
     return ARES_ENOMEM;
   }
 
-  memcpy(temp, val, len);
+  if (len != 0) {
+    memcpy(temp, val, len);
+  }
 
   /* NULL-term ABINP */
   if (is_nullterm) {
@@ -1410,14 +1421,32 @@ ares_status_t ares_dns_rr_set_bin(ares_dns_rr_t *dns_rr, ares_dns_rr_key_t key,
     (datatype == ARES_DATATYPE_BINP || datatype == ARES_DATATYPE_ABINP)
               ? ARES_TRUE
               : ARES_FALSE;
-  size_t         alloclen = is_nullterm ? len + 1 : len;
-  unsigned char *temp     = ares_malloc(alloclen);
+  size_t         alloclen;
+  unsigned char *temp;
+
+  if (datatype != ARES_DATATYPE_BIN && datatype != ARES_DATATYPE_BINP &&
+      datatype != ARES_DATATYPE_ABINP) {
+    return ARES_EFORMERR;
+  }
+
+  if (val == NULL && len != 0) {
+    return ARES_EFORMERR;
+  }
+
+  if (is_nullterm && len == SIZE_MAX) {
+    return ARES_ENOMEM;
+  }
+  alloclen = is_nullterm ? len + 1 : len;
+
+  temp = ares_malloc(alloclen);
 
   if (temp == NULL) {
     return ARES_ENOMEM;
   }
 
-  memcpy(temp, val, len);
+  if (len != 0) {
+    memcpy(temp, val, len);
+  }
 
   /* NULL-term BINP */
   if (is_nullterm) {
@@ -1573,12 +1602,23 @@ ares_status_t ares_dns_rr_set_opt(ares_dns_rr_t *dns_rr, ares_dns_rr_key_t key,
   ares_status_t  status;
 
   if (val != NULL) {
-    temp = ares_malloc(val_len + 1);
+    size_t alloclen;
+
+    if (val_len == SIZE_MAX) {
+      return ARES_ENOMEM;
+    }
+    alloclen = val_len + 1;
+
+    temp = ares_malloc(alloclen);
     if (temp == NULL) {
       return ARES_ENOMEM;
     }
-    memcpy(temp, val, val_len);
+    if (val_len != 0) {
+      memcpy(temp, val, val_len);
+    }
     temp[val_len] = 0;
+  } else if (val_len != 0) {
+    return ARES_EFORMERR;
   }
 
   status = ares_dns_rr_set_opt_own(dns_rr, key, opt, temp, val_len);
--- a/test/ares-test-internal.cc
+++ b/test/ares-test-internal.cc
@@ -1550,6 +1550,44 @@ TEST_F(LibraryTest, DNSRecordNSEC3EmptyFields) {
 }
 #endif /* !CARES_SYMBOL_HIDING */
 
+TEST_F(LibraryTest, DNSRecordRejectsInvalidBinaryInput) {
+  ares_dns_record_t *dnsrec = NULL;
+  ares_dns_rr_t     *rr     = NULL;
+  unsigned char      data[] = { 'x' };
+
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_record_create(&dnsrec, 0x1234, ARES_FLAG_RD,
+      ARES_OPCODE_QUERY, ARES_RCODE_NOERROR));
+
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER, "example.com",
+      ARES_REC_TYPE_TXT, ARES_CLASS_IN, 60));
+  EXPECT_EQ(ARES_ENOMEM,
+    ares_dns_rr_add_abin(rr, ARES_RR_TXT_DATA, data, SIZE_MAX));
+  EXPECT_EQ(ARES_EFORMERR,
+    ares_dns_rr_add_abin(rr, ARES_RR_TXT_DATA, NULL, 1));
+
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER, "example.com",
+      ARES_REC_TYPE_CAA, ARES_CLASS_IN, 60));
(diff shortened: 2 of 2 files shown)


bradh352 (commented):
## Review: Guard DNS record binary length overflow

Thanks for the patch. This is correct, minimal, and C89-clean. I'm good to take it as low-cost robustness hardening for the public setters, **provided the test is fixed** (see the Medium inline comment) — as written, the `set_bin` case doesn't actually exercise the new guard and is fragile under ASAN.

### Verified
- Builds clean under CMake+Ninja (zero warnings in the modified file); `DNSRecordRejectsOverflowingBinaryLengths` + `DNSRecord` + `DNSParseFlags` pass 3/3; ASAN-clean (with `allocator_may_return_null=1`).
- **C89-compliant** — all declarations at block start, no mid-block declarations.
- **Completeness** — these three (`ares_dns_rr_add_abin`, `ares_dns_rr_set_bin`, `ares_dns_rr_set_opt`) are the complete set of public setters with the `len + 1`-on-caller-length pattern. Every peer allocation site (`ares_strdup`, `ares_buf_*`, `set_addr`/`set_addr6`, multistring) is already SIZE_MAX-guarded or provably bounded. No dead code introduced; the new datatype guard in `set_bin` correctly mirrors `set_bin_own` (whose own check must stay — it has 4 other callers).
- The `(val == NULL, len != 0)` rejection and the `if (len != 0) memcpy(...)` guards fix genuine latent UB (`memcpy` with a NULL src is UB even at length 0) and are worth keeping regardless.

### Scope / reachability (for the record)
The `SIZE_MAX` overflow is **not reachable from untrusted DNS data** — the parsers use the internal `_own` variants (unaffected here), and wire lengths derive from a u16-bounded `rdlength`. It's only reachable via direct public-API misuse (a caller passing `SIZE_MAX`), which `SECURITY.md` treats as out of scope. No linked issue or fuzzer report, and no conflict/duplication with the other in-flight overflow work (#1116/#1117 touch different files). I'm accepting it anyway as cheap hardening of the public surface, but flagging the framing so expectations are clear: this is robustness, not a security fix.

### Notes that couldn't be inline
- One consolidation thought (also in the Low inline comment): a dedicated helper for a single addition-overflow check is more indirection than the one-liner needs — cf. the shared-overflow-helper direction on #1141/#1147. Non-blocking.
- Please keep the NULL-src / zero-length `memcpy` guards even if the `SIZE_MAX` helper is trimmed — those are the parts with real value.

Inline comments below.

test/ares-test-internal.cc
```diff
@@ -1105,6 +1105,36 @@ TEST_F(LibraryTest, DNSRecord) {
...
+    ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER, "example.com",
+      ARES_REC_TYPE_TXT, ARES_CLASS_IN, 60));
+  EXPECT_EQ(ARES_ENOMEM,
+    ares_dns_rr_add_abin(rr, ARES_RR_TXT_DATA, data, SIZE_MAX));
+
+  EXPECT_EQ(ARES_SUCCESS,
+    ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER, "example.com",
+      ARES_REC_TYPE_SIG, ARES_CLASS_ANY, 60));
+  EXPECT_EQ(ARES_ENOMEM,
+    ares_dns_rr_set_bin(rr, ARES_RR_SIG_SIGNATURE, data, SIZE_MAX));
```
bradh352: **[Medium] This case does not exercise the `set_bin` overflow guard, and is fragile under ASAN.** `ARES_RR_SIG_SIGNATURE` has datatype `ARES_DATATYPE_BIN` (`is_nullterm == FALSE`), so `ares_dns_rr_set_bin()` never computes `len + 1` here — it calls `ares_malloc(SIZE_MAX)` directly and the `ARES_ENOMEM` comes from that allocation *failing*, not from the new guard. The test would still pass if the `set_bin` guard were deleted. And relying on `malloc(SIZE_MAX)` returning `NULL` isn't portable: under AddressSanitizer's default `allocator_may_return_null=0` (and some hardened allocators) the process **aborts** with `allocation-size-too-big` instead of returning NULL, which would crash this test in those configs.

Use a null-terminated binary datatype key so the guard is actually hit — and note it short-circuits *before* `ares_malloc`, so no huge allocation is attempted at all. `ARES_RR_CAA_VALUE` is `ARES_DATATYPE_BINP`:

```c
  EXPECT_EQ(ARES_SUCCESS,
    ares_dns_record_rr_add(&rr, dnsrec, ARES_SECTION_ANSWER, "example.com",
      ARES_REC_TYPE_CAA, ARES_CLASS_IN, 60));
  EXPECT_EQ(ARES_ENOMEM,
    ares_dns_rr_set_bin(rr, ARES_RR_CAA_VALUE, data, SIZE_MAX));
```

(The `add_abin`/TXT and `set_opt`/OPT cases are fine — both are null-terminated paths that hit the guard and never attempt the allocation.)

src/lib/record/ares_dns_record.c
```diff
@@ -49,6 +49,21 @@ static void ares_dns_rr_free_cb(void *arg)
   ares_dns_rr_free(rr);
 }
 
+static ares_status_t ares_dns_rr_bin_alloc_len(ares_bool_t is_nullterm,
+                                               size_t len, size_t *alloclen)
+{
+  if (alloclen == NULL) {
```
bradh352: **[Nit]** All three callers pass a non-NULL `&alloclen`, so this branch is unreachable. The codebase marks such defensive branches so they don't show up as coverage gaps (e.g. `ares_str.c:76`):

```c
  if (alloclen == NULL) {
    return ARES_EFORMERR; /* LCOV_EXCL_LINE: DefensiveCoding */
  }
```

src/lib/record/ares_dns_record.c
```diff
@@ -49,6 +49,21 @@ static void ares_dns_rr_free_cb(void *arg)
   ares_dns_rr_free(rr);
 }
 
+static ares_status_t ares_dns_rr_bin_alloc_len(ares_bool_t is_nullterm,
```
bradh352: **[Low]** For a single addition-overflow check (`is_nullterm && len == SIZE_MAX` — the `+1` only overflows at exactly `SIZE_MAX`), a dedicated helper is arguably more indirection than the one-liner warrants. Given the consolidation direction on the parallel array-overflow PRs (#1141/#1147 — shared overflow helpers rather than one-offs), consider either inlining this at the three call sites, or keeping the helper but scoping/naming it clearly as the DNS-RR NUL-term allocation sizer. Non-blocking.

test/ares-test-internal.cc
```diff
@@ -1105,6 +1105,36 @@ TEST_F(LibraryTest, DNSRecord) {
   EXPECT_EQ(ARES_FALSE, ares_dns_rr_get_opt_byid(NULL, ARES_RR_A_ADDR, 1, NULL, NULL));
 }
 
+TEST_F(LibraryTest, DNSRecordRejectsOverflowingBinaryLengths) {
```
bradh352: **[Low]** Beyond the `SIZE_MAX` guard, the other new branches are untested: the `(val == NULL && len != 0) -> ARES_EFORMERR` rejection, the early datatype guard in `set_bin`, and the `len == 0` memcpy-skip (a legitimate empty-value set). A couple of extra assertions would lock these in, e.g. `EXPECT_EQ(ARES_EFORMERR, ares_dns_rr_set_bin(rr, ARES_RR_CAA_VALUE, NULL, 5));` and a `len == 0` round-trip.

Alb3e3:
Updated in 3fdcfbf:

- changed the `set_bin(SIZE_MAX)` regression to use `ARES_RR_CAA_VALUE` (`BINP`), so it exercises the null-terminated allocation path without the ASAN-fragile `BIN` allocation attempt;
- added public-API coverage for nonzero length with NULL input, the `set_bin` datatype guard, and `set_opt` overflow/null cases;
- removed the helper and inlined the single `len + 1` overflow checks at the three call sites.

Local checks: `git diff --check`, a fresh CMake build without tests, and a small ASAN public-API harness covering the added cases. Full local GTest configuration is blocked here because GTest is not installed. GitHub CI is passing on the completed build/test jobs so far; Coveralls failed during upload with a coveralls.io 502, and CLANG64/OpenBSD/Ubuntu latest/AppVeyor are still pending as of this comment.

bradh352 (approved):
Approving — this is clean, minimal public-API hardening and all prior review items are addressed:

- The `set_bin` regression now uses `ARES_RR_CAA_VALUE` (BINP), so it actually exercises the `is_nullterm && len == SIZE_MAX` guard and short-circuits *before* `ares_malloc`. Verified with a real ASAN build (default `allocator_may_return_null=0`) — no abort, and the new test passes under both normal and ASAN builds.
- Helper removed and the checks inlined at all three call sites.
- Added public-API coverage for the NULL-input rejection and the `set_bin` datatype guard (`CAA_TAG`).

Re-verified on the current HEAD:

- **Complete set.** `ares_dns_rr_add_abin`, `ares_dns_rr_set_bin`, and `ares_dns_rr_set_opt` are the full set of public setters using the `ares_malloc(len + 1)` + `memcpy(caller_buf)` pattern. The `_own` variants take ownership and don't copy; `set_addr`/`set_addr6`/`set_str` are already NULL-guarded or fixed-size. Nothing missed.
- **Correctness.** `temp` is initialized on every path (incl. the `val == NULL`/`val_len == 0` OPT path passing NULL to `set_opt_own`, which accepts it). No leak or double-free on the `_own`-failure `ares_free(temp)` paths. The `len == SIZE_MAX` check precisely catches the only value that overflows `len + 1`, and `if (len != 0) memcpy(...)` closes the NULL-src memcpy UB. C89-clean (all declarations at block start, including the nested-block `size_t alloclen;` in `set_opt`).
- **Builds clean** under CMake+Ninja, zero warnings in `ares_dns_record.c`.

Framing note (for the record): this is robustness hardening, not a security fix — the `SIZE_MAX` overflow isn't reachable from untrusted wire data (parsers use the `_own` variants with u16-bounded `rdlength`; the only in-tree caller of the copying `set_opt`, `ares_cookie.c`, passes fixed small lengths), so it's reachable only via direct public-API misuse, which SECURITY.md treats as out of scope. No linked issue, and no conflict/duplication with the parallel overflow work (#1116/#1117 merged in different files; #1141 only shares the test file with a distinct test block).

Two non-blocking observations, neither introduced by this PR:
- Empty `BIN` sets (`SIG_SIGNATURE`/`TLSA_DATA`/`RAW_RR_DATA`, `len == 0`) still return `ARES_ENOMEM` via `ares_malloc(0)` — pre-existing, and the new `len != 0` memcpy guard implies an intent to support empty values that isn't fully realized for the `BIN` datatype. Out of scope here.
- The PR description's Testing section references the old test name `DNSRecordRejectsOverflowingBinaryLengths`; the committed test is `DNSRecordRejectsInvalidBinaryInput` — worth updating so the reproduction commands work.

PR #1170, opened 2026-06-23
Add ARES_OPT_MINTIMEOUTMS channel option

naclander:
Add a new channel option `ARES_OPT_MINTIMEOUTMS` that lets callers override the minimum per-query timeout used by the adaptive timeout algorithm in ares_metrics_server_timeout(). This mirrors the existing `ARES_OPT_MAXTIMEOUTMS` option which controls the upper bound.

When a caller sets tries=1 (`ARES_OPT_TRIES`), the adaptive algorithm often computes a timeout near the 250ms floor because local DNS servers return cached responses quickly. Without retries and their exponential backoff, queries that require upstream recursion can time out. This new option lets callers raise the floor to account for that.

A value of 0 (the default) preserves existing behavior (250ms floor).

We noticed this while configuring `ARES_OPT_TRIES` to 1 to limit the number of DNS queries c-ares sends per resolution. With the default retry count, the 250ms floor is rarely an issue because retries double the timeout on each round, giving slower lookups time to complete. But with a single try, any lookup that needs upstream recursion exceeds the ~250ms window and fails.

### DIFF
--- a/include/ares.h
+++ b/include/ares.h
@@ -265,6 +265,7 @@ typedef enum {
 #define ARES_OPT_QUERY_CACHE     (1 << 21)
 #define ARES_OPT_EVENT_THREAD    (1 << 22)
 #define ARES_OPT_SERVER_FAILOVER (1 << 23)
+#define ARES_OPT_MINTIMEOUTMS    (1 << 24)
 
 /* Nameinfo flag values */
 #define ARES_NI_NOFQDN        (1 << 0)
@@ -391,6 +392,7 @@ struct ares_options {
   char              *hosts_path;
   int                udp_max_queries;
   int                maxtimeout; /* in milliseconds */
+  int                mintimeout; /* in milliseconds */
   unsigned int qcache_max_ttl;   /* Maximum TTL for query cache, 0=disabled */
   ares_evsys_t evsys;
   struct ares_server_failover_options server_failover_opts;
--- a/include/ares.h
+++ b/include/ares.h
@@ -392,10 +392,10 @@ struct ares_options {
   char              *hosts_path;
   int                udp_max_queries;
   int                maxtimeout; /* in milliseconds */
-  int                mintimeout; /* in milliseconds */
   unsigned int qcache_max_ttl;   /* Maximum TTL for query cache, 0=disabled */
   ares_evsys_t evsys;
   struct ares_server_failover_options server_failover_opts;
+  int          mintimeout; /* in milliseconds */
 };
 
 struct hostent;
--- a/src/lib/ares_metrics.c
+++ b/src/lib/ares_metrics.c
@@ -247,8 +247,12 @@ size_t ares_metrics_server_timeout(const ares_server_t  *server,
   }
 
   /* don't go below lower bounds */
-  if (timeout_ms < MIN_TIMEOUT_MS) {
-    timeout_ms = MIN_TIMEOUT_MS;
+  {
+    size_t min_timeout_ms =
+      channel->mintimeout ? channel->mintimeout : MIN_TIMEOUT_MS;
+    if (timeout_ms < min_timeout_ms) {
+      timeout_ms = min_timeout_ms;
+    }
   }
 
   /* don't go above upper bounds */
--- a/src/lib/ares_metrics.c
+++ b/src/lib/ares_metrics.c
@@ -37,7 +37,8 @@
  * rails.
  *
  * Values:
- * - Minimum Timeout: 250ms (approximate RTT half-way around the globe)
+ * - Minimum Timeout: 250ms (approximate RTT half-way around the globe), can be
+ *   raised by ARES_OPT_MINTIMEOUTMS.
  * - Maximum Timeout: 5000ms (Recommended timeout in RFC 1123), can be reduced
  *   by ARES_OPT_MAXTIMEOUTMS, but otherwise the bound specified by the option
  *   caps the retry timeout.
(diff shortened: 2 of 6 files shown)


include/ares.h
```diff
@@ -391,6 +392,7 @@ struct ares_options {
   char              *hosts_path;
   int                udp_max_queries;
   int                maxtimeout; /* in milliseconds */
+  int                mintimeout; /* in milliseconds */
```
bradh352: **Major (ABI break).** Inserting `mintimeout` here shifts the offsets of every following member (`qcache_max_ttl`, `evsys`, `server_failover_opts`), which breaks binary compatibility for any already-compiled consumer of `struct ares_options`. This is the same issue fixed in `4acd5759` ("the maxtimeout must come at the end of the structure"). New fields must be appended at the tail. Please remove the insertion here and add it after `server_failover_opts`:

```c
  ares_evsys_t evsys;
  struct ares_server_failover_options server_failover_opts;
  int          mintimeout; /* in milliseconds */
};
```

src/lib/ares_metrics.c
```diff
@@ -247,8 +247,12 @@ size_t ares_metrics_server_timeout(const ares_server_t  *server,
...
   /* don't go below lower bounds */
-  if (timeout_ms < MIN_TIMEOUT_MS) {
-    timeout_ms = MIN_TIMEOUT_MS;
+  {
+    size_t min_timeout_ms =
+      channel->mintimeout ? channel->mintimeout : MIN_TIMEOUT_MS;
+    if (timeout_ms < min_timeout_ms) {
+      timeout_ms = min_timeout_ms;
+    }
   }
```
bradh352: **Nit (style):** For symmetry with the `max_timeout_ms` handling just below — declared at the top of the function and computed inline — consider dropping the extra block scope:

```c
  size_t                min_timeout_ms;
  ...
  /* don't go below lower bounds */
  min_timeout_ms = channel->mintimeout ? channel->mintimeout : MIN_TIMEOUT_MS;
  if (timeout_ms < min_timeout_ms) {
    timeout_ms = min_timeout_ms;
  }
```

**Low (min vs. max interaction):** The min floor is applied before the max cap, so a `mintimeout` set above the effective max (`maxtimeout`, or the 5000 ms `MAX_TIMEOUT_MS` default when unset) is silently reduced back down to the max. Given the feature's purpose (raising the floor for slow upstream recursion under `tries=1`), a caller could reasonably set `mintimeout > 5000` and be surprised it's capped unless they also raise `maxtimeout`. At minimum the documentation should call this out; optionally clamp/validate explicitly.

bradh352 (changes requested):
**Summary review.** Thanks for this — the feature is reasonable and the implementation is a clean mirror of `ARES_OPT_MAXTIMEOUTMS`: the intent (raise the adaptive-timeout floor when `tries=1`, since a single try with no exponential backoff can't absorb upstream recursion past the 250 ms floor) matches its effect. Build is clean with no new warnings, existing option tests pass, the `ares_dup()` save/init wiring is correct and complete (the option survives a dup), the optmask bit `(1 << 24)` is the next free bit, and a default of `0` correctly preserves current behavior. No RFC/interoperability/security/manual-parsing concerns — this is purely client-side timeout tuning.

Requesting changes on one blocker plus some missing docs/tests:

**Major (ABI) — inline on `include/ares.h`.** `mintimeout` is inserted mid-struct, shifting the offsets of `qcache_max_ttl`/`evsys`/`server_failover_opts` and breaking binary compatibility. It's the same issue fixed in `4acd5759` ("the maxtimeout must come at the end of the structure"). Move it to the tail of the struct.

**Medium — Missing documentation.** `docs/ares_init_options.3` documents `maxtimeout` (the struct-field list near line 41 and an `ARES_OPT_MAXTIMEOUTMS` section near line 177) but has no `mintimeout` / `ARES_OPT_MINTIMEOUTMS` entries. Please mirror the maxtimeout docs.

**Medium — Missing test coverage.** No test exercises the new option (`grep mintimeout test/` returns nothing), while `maxtimeout` is covered in `LibraryTest.OptionsChannelInit` (`test/ares-test-init.cc`). Mirror it:
```cpp
optmask |= ARES_OPT_MINTIMEOUTMS;
opts.mintimeout = 500;
// ...after the ares_dup + ares_save_options round-trip:
EXPECT_EQ(opts.mintimeout, opts2.mintimeout);
```

**Low + Nit — inline on `src/lib/ares_metrics.c`** (min/max ordering interaction, and a style-symmetry suggestion).

**Nit — Algorithm comment.** `src/lib/ares_metrics.c:39-43` notes the max "can be reduced by `ARES_OPT_MAXTIMEOUTMS`"; add the symmetric note that the minimum can be raised by `ARES_OPT_MINTIMEOUTMS`.

**Policy note (your call, non-blocking).** The comment above `struct ares_options` states new options should ideally be exposed via `ares_set_*()` setters rather than the struct. Recent options (`maxtimeout`, `qcache_max_ttl`, `evsys`, `server_failover_opts`) were nonetheless appended to the struct, so this mechanism follows real convention — noting only in case a setter is preferred. No overlapping PRs/issues exist, and the PR links no issue.

naclander:
Thanks for the review @bradh352, I've gone ahead and addressed your comments. Let me know if there is anything else.

naclander:
@bradh352 May I please get another review?

naclander:
@bradh352 Another ping please

naclander:
@bradh352 Another ping please.

markhankins:
@bradh352 Sorry to bother you, but any chance of a re-review please? 

This fix is for an issue which has been impacting our project for over a year and is making another issue potentially. 

Thanks.

naclander:
@bradh352 Can you please take a look at this PR again please?

naclander:
@bradh352 can you take a look at this please?

PR #1173, opened 2026-06-29
enforce rfc1035 255-octet name limit when parsing dns names

Arawoof06:
ares_dns_name_parse decodes a name without any cap on its length, while the write side in ares_split_dns_name already rejects anything past the rfc1035 255-octet limit. a response can therefore point every resource record's owner name at one long label chain through a 2-byte compression pointer, and each record re-expands the whole chain from scratch, so a 62kb message spends ~240ms in this parser versus well under a millisecond once the limit is enforced. track the running label length in the parse loop and bail with ares_ebadname once it crosses 255, which bounds the work per name and matches the write path. added a regression case alongside parsemalformedrrcount.

### DIFF
--- a/src/lib/record/ares_dns_name.c
+++ b/src/lib/record/ares_dns_name.c
@@ -543,6 +543,7 @@ ares_status_t ares_dns_name_parse(ares_buf_t *buf, char **name,
   ares_status_t status;
   ares_buf_t   *namebuf     = NULL;
   size_t        label_start = ares_buf_get_position(buf);
+  size_t        name_len    = 0;
 
   if (buf == NULL) {
     return ARES_EFORMERR;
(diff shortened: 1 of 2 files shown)


dimbleby:
You and your bot forgot to check for already-open pull requests.  This looks like #1164, but one of you has an out-by-one error.

Arawoof06:
fair, I missed #1164 before opening this. and the out-by-one is mine: I add the length octet per label (c + 1), so my running total is sum(labels) + N, whereas ares_split_dns_name on the write side counts the separators, sum(labels) + (N-1). that makes mine a byte too strict and it'd reject a legal 255-octet name (e.g. four 63-octet labels = 252 + 3 dots = 255). #1164 matches the write path, so closing this in favor of it.

PR #1179, opened 2026-07-02
ci(qnx): retry SDP download/licensing steps to survive transient timeouts

bradh352:
## Problem

The QNX job regularly fails in the **"Download SDP 8.0"** setup step with:

```
curl: (28) Failed to connect to support7.qnx.com port 443 after 136498 ms: Connection timed out
##[error]Process completed with exit code 28.
```

QNX's download/licensing servers are intermittently unreachable from the runner. Because this happens during setup, an ~11-minute run is thrown away before any c-ares code is even built. (This is separate from the test-level QNX failure fixed in #1178.)

## Change

Harden only the network-facing setup steps so a transient blip retries instead of failing the whole job — no behavioral change on success:

- A small shell `retry()` helper (4 attempts, 20s apart) wraps the two `qnxsoftwarecenter_clt` invocations (license sync, SDP mirror install).
- Both `curl` calls get `--connect-timeout 60` (the observed hangs ran ~136s per attempt, so an unbounded connect wastes time) plus `--retry 5 --retry-delay 10 --retry-connrefused`.

The actual c-ares CMake/autotools builds are deliberately **not** retried — only the flaky third-party downloads are.

## Notes

- Caching the SDP was considered and rejected: it's proprietary, license-gated content, and an `actions/cache` entry on a public repo can be read by fork-PR workflows (exfiltration risk). Retry-hardening is the safe lever; a self-hosted runner with the SDP pre-installed (as qnx-ports uses) would be the longer-term fix.

## Validation

- YAML parses; the embedded script passes `bash -n`; the `retry()` helper was unit-tested (retries a flaky command to success, gives up non-zero after max so `set -e` fails the step, passes through immediate success).
- QNX runs on `push`, so the branch push triggers a real QNX run that exercises this path.

### DIFF
--- a/.github/workflows/qnx.yml
+++ b/.github/workflows/qnx.yml
@@ -33,17 +33,39 @@ jobs:
           QNX_PASSWORD: ${{ secrets.QNX_PASSWORD }}
         run: |
           [ -z "${QNX_USER}" -o -z "${QNX_PASSWORD}" -o -z "${QNX_LICENSE_KEY}" ] && echo "Must set QNX_USER, QNX_PASSWORD, and QNX_LICENSE_KEY" && /bin/false
+
+          # QNX's download/licensing servers intermittently time out (curl exit
+          # 28), which fails this ~11 minute job during setup before any c-ares
+          # code is even built.  Retry the network-facing steps a few times so a
+          # transient blip doesn't fail the whole run.  curl gets its own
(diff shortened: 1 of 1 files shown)


bradh352:
Superseded by #1180, which combines this download hardening with the MockUDPMaxQueriesTest fix (#1178) into a single PR.

PR #1185, opened 2026-07-03
ci: validate clang-format on PR-changed lines

bradh352:
## Summary

Adds a `clang-format` workflow that validates formatting of **only the lines actually changed by the PR**, using `clang-format-diff` fed by `git diff -U0` against the PR base. Pre-existing style drift elsewhere in a touched file never blocks an unrelated change.

Also updates `.clang-format` to set `AlignConsecutiveDeclarations.AlignFunctionDeclarations: false` (added in clang-format 20 by llvm/llvm-project#108241), so consecutive function prototypes are no longer padded into column alignment. Variable declaration alignment is unchanged — the previous `AcrossEmptyLinesAndComments` shorthand is expanded to its equivalent explicit form with the new sub-option added.

## clang-format version

The workflow discovers the **latest stable LLVM release at runtime** by parsing `CURRENT_LLVM_STABLE` from apt.llvm.org's `llvm.sh` (currently 22), then installs just `clang-format-NN` from the matching apt repo — no toolchain bloat, and the check tracks new LLVM releases automatically. Since only changed lines are checked, a formatting-behavior change in a new release has minimal blast radius.

## Details

- Scope matches DEVELOPER-NOTES.md: `*.c` / `*.h`, excluding `src/lib/thirdparty`.
- On failure the suggested fix is printed to the log, published as a ```diff``` block in the job step summary, and surfaced via a `::error::` annotation.
- `paths` filter skips the workflow entirely for PRs not touching C sources or the format config (skips count as passing for the **All Checks** aggregator from #1182).
- `clang-format-diff` exits non-zero when it emits a diff; that's neutralized (`|| true`) so the explicit patch-size check controls failure and the diagnostics always print.
- Verified locally with clang-format 22.1.8: the updated `.clang-format` parses, and a synthetic bad-format change is flagged while untouched lines in the same file are left alone.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -26,7 +26,12 @@ AlignConsecutiveAssignments:
   AlignCompound: true
   PadOperators: true
 AlignConsecutiveBitFields: Consecutive
-AlignConsecutiveDeclarations: AcrossEmptyLinesAndComments
+# AlignFunctionDeclarations requires clang-format >= 20
+AlignConsecutiveDeclarations:
+  Enabled: true
+  AcrossEmptyLines: true
+  AcrossComments: true
+  AlignFunctionDeclarations: false
 AlignConsecutiveMacros: Consecutive
 AlignEscapedNewlines: Left
 AlignOperands: Align
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,71 @@
+# Copyright (C) The c-ares project and its contributors
+#
+# SPDX-License-Identifier: MIT
+
+name: clang-format
+
+# Validates formatting of lines changed by the PR (and only those lines,
+# via clang-format-diff) against the repository .clang-format rules.
+#
+# Always uses the latest stable clang-format from apt.llvm.org, discovered
+# at runtime from the llvm.sh installer script.  Our configuration requires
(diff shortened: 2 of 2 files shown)


dimbleby:
fyi I considered something like this at https://github.com/c-ares/c-ares/pull/617.  

It's a long time ago now but IIRC I abandoned it because my local clang-format did not match the CI clang-format - and I figured that if it was turning out to be troublesome for me, then it was likely also to be a pain in the neck for other contributors.

bradh352:
> fyi I considered something like this at #617.
> 
> It's a long time ago now but IIRC I abandoned it because my local clang-format did not match the CI clang-format - and I figured that if it was turning out to be troublesome for me, then it was likely also to be a pain in the neck for other contributors.

There's so little churn to c-ares I don't think its a show stopper (and yes, a lot of that is my fault since I was pulled out of retirement and into a startup).  I wouldn't have wanted to do it a year or so ago because the requirement is llvm v20 or higher due to the settings, but I think its perfectly reasonable now to require it.  And it only impacts the lines that are actually modified so they won't trigger on changes to a file with pre-existing issues.  

Then of course a lot of people are using AI agents these days so they'll just kick the bot to fix trivial stuff like that.

PR #1192, opened 2026-07-04
Fix sticky server recovery when all servers have failures

bradh352:
## Problem

In the review discussion on #992, @DavidMcNallyAlianza and @Flynnzaa identified a failure mode they named **"sticky recovery"**: when every configured server is failing, a server that recovers from an extended outage can go unnoticed for an unbounded amount of time. Credit to them for the analysis and the timeline diagram that made this concrete.

Their scenario (2 servers, 1 qps): both servers fail for 10 minutes (both counters reach 600). Server A briefly recovers — resetting its counter to 0 — then starts failing again, while server B's counter stays frozen at 600 because no traffic reaches it. When B actually recovers moments later, nothing notices: all queries keep going to A (count ~60 and climbing) until A's counter climbs past B's frozen 600. That's ~9 minutes of avoidable downtime in their example, and in general the blindness window scales with the length of the original outage, at any query rate.

Two root causes in the current code:

1. **Probing is gated off exactly when it's needed most.** `ares_send_query()` only considers spawning a probe when the triggering query was sent to a server with *no* failures. During a total outage every query necessarily goes to a failed server, so no probes are ever sent, and the "auto-rotate on each attempt" property of the sorted server list only holds while failure counts stay within ±1 of each other — a recovery blip breaks that symmetry permanently.
2. **The consecutive-failure count is unbounded and freezes when traffic stops.** Only the *relative order* of the counts is used for selection, so magnitude beyond "clearly down" carries no signal — but a huge frozen count means an equally huge number of failures elsewhere before the server is ever consulted again.

## Fix (three small changes, no new state, no new options)

1. **Relax the probe trigger** (`ares_process.c`): a fresh query sent to a failed server can now spawn a probe of a *different* failed server; the probe-selection scan skips the triggering server since the query itself already exercises it. All existing throttles are unchanged (1-in-`retry_chance` dice, per-server `next_retry_time` cooldown, `probe_pending` single-flight), so probe load remains bounded during an outage. In the scenario above, B now gets probed within ~`retry_chance` queries of recovering (~10s at 1 qps with defaults) instead of ~9 minutes.
2. **Cap `consec_failures` at 16** (`SERVER_CONSEC_FAILURES_CAP`): bounds how far a stale counter can diverge, as a backstop for configurations where probing is disabled.
3. **Break sort ties by `next_retry_time`** (`ares_init.c`): among servers with equal failure counts, the least-recently-failed sorts first. This keeps retries rotating across servers once counts saturate at the cap (previously the `idx` tiebreak would pin all retries to the lowest-index server), and it means a stale failed server — which may have silently recovered — is retried before one that just failed. This resolves the sticky-recovery scenario even with probing disabled (`retry_chance = 0`). Healthy servers all have a zeroed retry time and continue to be used in configuration order, so behavior with no failures is unchanged.

`server_increment_failures()`/`server_set_good()` now update `next_retry_time` *before* reinserting into the sorted list, since the sort depends on it.

## Tests

Two new mock tests, both verified to **fail against the previous code** and pass with this change:

- `ServerRecoveryMultiMockTest.StickyRecoveryProbe` — all servers fail; a query to a still-failed server must spawn a probe of the other failed server, which is how its later recovery is detected.
- `ServerRecoveryNoProbeMultiMockTest.StickyRecoveryTieBreak` — probing disabled; after a recovery blip on server 0, a query's retries must reach the stale-failed (recovered) server 1 once the counts tie, rather than burning every attempt on server 0.

The existing `ServerFailoverOpts` and `ProbeAfterProbeTimeout` tests pass unchanged, as do the broader mock suites (120 tests run locally, CMake+Ninja, no new warnings, clang-format clean on changed lines).

## Relation to #992

This addresses the "sticky recovery" concern (concern 1 of 3) from that discussion in isolation, since it's a bug in current behavior independent of the rise/fall threshold feature. The rise/fall thresholds (#987) and the per-request retry-distribution question (concerns 2/3) remain under discussion in #992 — notably, once rise thresholds exist, the probe path here is what delivers the N consecutive successes a recovered server needs, so the two changes compose.

### DIFF
--- a/src/lib/ares_init.c
+++ b/src/lib/ares_init.c
@@ -102,6 +102,20 @@ static int server_sort_cb(const void *data1, const void *data2)
   if (s1->consec_failures > s2->consec_failures) {
     return 1;
   }
+
+  /* Among servers with the same failure count, prefer the one that failed
+   * least recently so that a server that went down long ago (and may have
+   * since recovered) is tried before one that just failed.  This also keeps
+   * retries rotating across servers once the failure counts saturate at
+   * SERVER_CONSEC_FAILURES_CAP.  Healthy servers all have a zeroed retry
+   * time and fall through to configuration order. */
+  if (s1->next_retry_time.sec != s2->next_retry_time.sec) {
+    return s1->next_retry_time.sec < s2->next_retry_time.sec ? -1 : 1;
+  }
+  if (s1->next_retry_time.usec != s2->next_retry_time.usec) {
+    return s1->next_retry_time.usec < s2->next_retry_time.usec ? -1 : 1;
+  }
+
   if (s1->idx < s2->idx) {
     return -1;
   }
--- a/src/lib/ares_private.h
+++ b/src/lib/ares_private.h
@@ -131,6 +131,13 @@ W32_FUNC const char *_w32_GetHostsFile(void);
 #define DEFAULT_SERVER_RETRY_CHANCE 10
 #define DEFAULT_SERVER_RETRY_DELAY  5000
 
+/* Upper bound on the consecutive failure count tracked per server.  Only the
+ * relative order of the counts is used for server selection, so magnitude
+ * beyond "clearly down" carries no additional signal.  Capping it bounds how
+ * long a server that failed for an extended period sorts behind other failed
+ * servers once it comes back online. */
+#define SERVER_CONSEC_FAILURES_CAP 16
+
 struct ares_query;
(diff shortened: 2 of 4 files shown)


Kevin-Molina:
Hey @bradh352 - Thank you for the fix here. When is this change expected to be released?

ericsmalling:
+1 to @Kevin-Molina 's question: is there a schedule for when this will be released?

PR #1195, opened 2026-07-04
Fix recursive retry dispatch in ares_requeue_query()

### LINKED ISSUE #1145: Unbounded Recursion in `ares_requeue_query` / `ares_send_query` Causes Stack Overflow
# Unbounded Recursion in `ares_requeue_query` / `ares_send_query` Causes Stack Overflow

## Summary
When `ares_open_connection()` fails synchronously with a retryable error (e.g. `ARES_ECONNREFUSED`, `ARES_EBADFAMILY`) while `ares_send_query()` is called with `requeue == NULL`, `ares_requeue_query()` re-invokes `ares_send_query()` directly instead of enqueuing the retry. Every retry therefore adds a real C stack frame, and the recursion depth is `num_servers * channel->tries`. With a sufficiently large `ARES_OPT_TRIES` (or just enough servers) the process exhausts its stack and crashes. This is a reliability / availability defect (CWE-674: Uncontrolled Recursion → Stack Overflow).

## Version
```
$ git describe
v1.31.0-357-gbd7d871e (the main branch)
```

## Description
`ares_send_query()` in `src/lib/ares_process.c` calls `ares_open_connection()` (line 1307). When that returns a retryable error it calls:

```c
return ares_requeue_query(query, now, status, ARES_TRUE, NULL, NULL);
```

Note the last argument (`requeue`) is `NULL`. `ares_requeue_query()` then does:

```c
/* src/lib/ares_process.c:994-1000 */
if (query->try_count < max_tries && !query->no_retries) {
    if (requeue != NULL) {
        return ares_append_requeue(requeue, query, NULL);   /* iterative path */
    }
    return ares_send_query(NULL, query, now);               /* direct recursion */
}
```

where

```c
size_t max_tries = ares_slist_len(channel->servers) * channel->tries;
```

Because `requeue == NULL`, the retry is implemented as a direct tail call into `ares_send_query()`. C has no guaranteed tail-call optimisation, so each retry consumes a stack frame. If `ares_open_connection()` keeps failing synchronously (which is a supported outcome — any user-provided `asocket` callback, `sock_config_cb`, `sock_create_cb`, or real-world conditions like `EMFILE`/`EAFNOSUPPORT` can cause it), the two functions recurse into each other `num_servers * channel->tries` times before the normal termination condition fires.

With default configuration (`tries = 3`, one or two servers) this recursion is shallow and harmless. But:

- `ARES_OPT_TRIES` currently only rejects `tries <= 0` (`src/lib/ares_options.c:314-319`); any positive `int` up to `INT_MAX` is accepted.
- `options attempts:N` in `/etc/resolv.conf` is parsed into `sysconfig->tries` (`src/lib/ares_sysconfig.c:482`).
- The same sync-failure path is hit for every retry, so depth grows linearly with the configured retry count.

In this configuration, the stack is exhausted before retries are ever exhausted, and the process is terminated by `SIGSEGV` on the guard page (ASan reports it as `stack-overflow`).

### Call pattern observed

```
ares_send_query
  └─ ares_open_connection  → ARES_ECONNREFUSED (sync failure)
       └─ ares_requeue_query         (requeue == NULL)
            └─ ares_send_query       (recursion, +1 stack frame)
                 └─ ares_open_connection → ARES_ECONNREFUSED
                      └─ ares_requeue_query
                           └─ ares_send_query
                                … repeats until stack is exhausted …
```

## PoC Code
The PoC uses a user-installed `asocket` that always returns `ARES_SOCKET_BAD`. This is a well-defined way to signal a socket-open failure through the public `ares_set_socket_functions()` API and reliably drives `ares_open_connection()` down the synchronous-failure path. `tries` is set to a large value so the stack is exhausted quickly under ASan; the underlying defect is the recursive retry, not the size of `tries`.

```cpp
#include <stddef.h>
#include <stdint.h>
#include <string.h>
#include <ares.h>
static void dummy_query_callback(void *arg, int status, int timeouts, const unsigned char *abuf, int alen) {
    (void)arg; (void)status; (void)timeouts; (void)abuf; (void)alen;
}
static ares_socket_t custom_asocket(int domain, int type, int protocol, void *user_data) {
    (void)domain; (void)type; (void)protocol; (void)user_data;
    return ARES_SOCKET_BAD;
}
extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
    if (size < 1) return 0;
    if (ares_library_init(ARES_LIB_INIT_ALL) != ARES_SUCCESS) {
        return 0;
    }
    struct ares_options options;
    memset(&options, 0, sizeof(options));
    options.tries = 0x7FFFFFFF; 
    int optmask = ARES_OPT_TRIES;
    if (ares_init_options(&channel, &options, optmask) != ARES_SUCCESS) {
        ares_library_cleanup();
        return 0;
    }
    struct ares_socket_functions socket_funcs;
    memset(&socket_funcs, 0, sizeof(socket_funcs));
    socket_funcs.asocket = custom_asocket;
    
    ares_set_socket_functions(channel, &socket_funcs, nullptr);
    ares_query(channel, "example.com", ARES_CLASS_IN, ARES_REC_TYPE_A, dummy_query_callback, nullptr);
    ares_destroy(channel);
    ares_library_cleanup();

    return 0;
}
```

## Reproduction Step
```
clang++ -g -O0 -fno-omit-frame-pointer -fsanitize=fuzzer,address,undefined \
    -I<path-to-c-ares>/include \
    poc.cpp -o poc \
    <path-to-libcares.a> -lpthread
./poc
```

## Stack Trace
```
==51902==ERROR: AddressSanitizer: stack-overflow on address 0x7ffec4115fb8 (pc 0x62a42e385f47 bp 0x7ffec41167f0 sp 0x7ffec4115fc0 T0)
    #0 0x62a42e385f47 in malloc (/root/FuzzAgent/output/c-ares/crash_reports/crash_005/poc+0x16df47) (BuildId: 4448ab4c4f458631f2e8bb713a7f5f2fd91d5857)
    #1 0x62a42e3d5afd in default_malloc /root/src/c-ares/src/lib/ares_library_init.c:52:10
    #2 0x62a42e3d5989 in ares_malloc /root/src/c-ares/src/lib/ares_library_init.c:71:10
    #3 0x62a42e441d2f in ares_open_connection /root/src/c-ares/src/lib/ares_conn.c:355:10
    #4 0x62a42e3dff90 in ares_send_query /root/src/c-ares/src/lib/ares_process.c:1294:14
    #5 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12
    #6 0x62a42e3e046a in ares_send_query /root/src/c-ares/src/lib/ares_process.c
    #7 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12
    #8 0x62a42e3e046a in ares_send_query /root/src/c-ares/src/lib/ares_process.c
    #9 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12
    #10 0x62a42e3e046a in ares_send_query /root/src/c-ares/src/lib/ares_process.c
    #11 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12
    #12 0x62a42e3e046a in ares_send_query /root/src/c-ares/src/lib/ares_process.c
    #13 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12
    #14 0x62a42e3e046a in ares_send_query /root/src/c-ares/src/lib/ares_process.c
    #15 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12
    #16 0x62a42e3e046a in ares_send_query /root/src/c-ares/src/lib/ares_process.c
    #17 0x62a42e3df7e3 in ares_requeue_query /root/src/c-ares/src/lib/ares_process.c:986:12      
    ... (frames #8–#N repeat until the stack is exhausted) ...
```

## Security Impact
This is an Uncontrolled Recursion defect (CWE-674) that manifests as a stack overflow and terminates the calling process. The crash lands on a stack guard page, and lead to availability / denial of service.

---
> Signed-off-by: FuzzAnything fuzzanything@gmail.com

Alearner12:
# Fix recursive retry dispatch in ares_requeue_query()

## Summary

Fixes c-ares/c-ares#1145.
Related to c-ares/c-ares#1043.

`ares_requeue_query()` can currently call `ares_send_query()` directly when it needs to retry a query and no deferred requeue list is provided. When socket creation or connection setup fails synchronously with a retryable error, this creates mutual recursion:

```text
ares_send_query()
  -> ares_requeue_query()
     -> ares_send_query()
        -> ...
```

The recursion depth is bounded by `num_servers * channel->tries`, but `ARES_OPT_TRIES` accepts large positive values. With a sufficiently large retry budget, the process exhausts its stack before the retry limit is reached.

This change routes retry dispatch through the existing requeue mechanism and drains retries iteratively, avoiding one C stack frame per retry.

## Reachability

The issue is reachable through public c-ares APIs:

- `ares_init_options()` accepts `ARES_OPT_TRIES`.
- `ares_set_socket_functions()` or `ares_set_socket_functions_ex()` can provide a socket implementation whose `asocket` fails synchronously.
- `ares_query()` / `ares_query_dnsrec()` dispatches the query.
- `ares_send_query()` treats synchronous `ARES_ECONNREFUSED` / `ARES_EBADFAMILY` from `ares_open_connection()` as retryable.
- The old `ares_requeue_query(..., requeue == NULL)` path immediately re-entered `ares_send_query()`.

The same control-flow shape can also occur when real socket setup fails synchronously in a retryable way.

## Impact

The confirmed impact is uncontrolled recursion leading to stack exhaustion and process termination. This is best described conservatively as an availability issue / denial of service in callers that allow a large retry budget or many configured servers.

## Fix

- Add an internal `ares_send_query_int()` helper that accepts an optional requeue list.
- Reuse the existing `ares_requeue_t` array machinery for retry dispatch.
- Add `ares_flush_requeue()` so top-level callers without an existing queue can still process retries iteratively.
- Update retryable send/open failure paths to append to the active requeue list instead of recursing directly.
- Add a regression test that configures a high retry count and forces synchronous socket-open failures.

## Verification

Performed locally:

```text
cmake -S c-ares -B c-ares/build-notests \
  -DCARES_BUILD_TESTS=OFF -DCARES_STATIC=ON -DCARES_SHARED=OFF
cmake --build c-ares/build-notests --parallel 2
```

Result: patched production build completed successfully.

Standalone ASan reproducer against an unpatched worktree at the same base commit:

```text
env ASAN_OPTIONS=abort_on_error=1:detect_leaks=0 \
  c-ares-unpatched/build-notests/repro_requeue_recursion 100000
```

Result: reproduced `AddressSanitizer: stack-overflow`; stack trace alternates `ares_send_query()` and `ares_requeue_query()`.

Same reproducer against the patched build:

```text
env ASAN_OPTIONS=abort_on_error=1:detect_leaks=0 \
  c-ares/build-notests/repro_requeue_recursion 100000
```

Result:

```text
callback_done=1 callback_status=11 socket_calls=100000
```

Also performed:

```text
git diff --check
```

Result: clean.

Not performed:

Full c-ares test suite. The local machine is missing GMock (`LIBGMOCK` not found), so `CARES_BUILD_TESTS=ON` configure does not complete here. The regression test is included for upstream CI.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -65,6 +65,10 @@ static void        end_query(ares_channel_t *channel, ares_server_t *server,
                              ares_query_t *query, ares_status_t status,
                              ares_dns_record_t *dnsrec,
                              ares_array_t **requeue);
+static ares_status_t ares_send_query_int(ares_server_t        *requested_server,
+                                         ares_query_t         *query,
+                                         const ares_timeval_t *now,
+                                         ares_array_t        **requeue);
 
 static void        ares_query_remove_from_conn(ares_query_t *query)
 {
@@ -586,6 +590,54 @@ static ares_status_t ares_append_endqueue(ares_array_t     **requeue,
     dnsrec);
 }
 
+static ares_status_t ares_flush_requeue(ares_channel_t       *channel,
+                                        ares_array_t        **requeue,
+                                        const ares_timeval_t *now,
+                                        ares_status_t         status)
+{
+  if (requeue == NULL || *requeue == NULL) {
+    return status;
+  }
+
+  while (ares_array_len(*requeue) > 0) {
+    ares_query_t  *query;
+    ares_requeue_t entry;
+    ares_status_t  internal_status;
+
+    internal_status = ares_array_claim_at(&entry, sizeof(entry), *requeue, 0);
+    if (internal_status != ARES_SUCCESS) {
+      break;
+    }
+
+    query = ares_htable_szvp_get_direct(channel->queries_by_qid, entry.qid);
+
+    if (entry.type == REQUEUE_REQUEUE) {
+      /* query disappeared */
+      if (query == NULL) {
+        continue;
+      }
+      internal_status = ares_send_query_int(entry.server, query, now, requeue);
+      /* We only care about ARES_ENOMEM */
+      if (internal_status == ARES_ENOMEM) {
+        status = ARES_ENOMEM;
+      }
+    } else { /* REQUEUE_ENDQUERY */
+      if (query != NULL) {
+        query->callback(query->arg, entry.status, query->timeouts,
+                        entry.dnsrec);
+        ares_free_query(query);
+      }
+      ares_dns_record_destroy(entry.dnsrec);
+    }
+  }
+  /* Don't forget to send notification if queue emptied */
+  ares_queue_notify_empty(channel);
+  ares_array_destroy(*requeue);
+  *requeue = NULL;
+
+  return status;
+}
+
 static ares_status_t read_answers(ares_conn_t *conn, const ares_timeval_t *now)
 {
   ares_status_t   status;
@@ -638,45 +690,7 @@ static ares_status_t read_answers(ares_conn_t *conn, const ares_timeval_t *now)
   }
 
 cleanup:
-
-  /* Flush requeue */
-  while (ares_array_len(requeue) > 0) {
-    ares_query_t  *query;
-    ares_requeue_t entry;
-    ares_status_t  internal_status;
-
-    internal_status = ares_array_claim_at(&entry, sizeof(entry), requeue, 0);
-    if (internal_status != ARES_SUCCESS) {
-      break;
-    }
-
-    query = ares_htable_szvp_get_direct(channel->queries_by_qid, entry.qid);
-
-    if (entry.type == REQUEUE_REQUEUE) {
-      /* query disappeared */
-      if (query == NULL) {
-        continue;
-      }
-      internal_status = ares_send_query(entry.server, query, now);
-      /* We only care about ARES_ENOMEM */
-      if (internal_status == ARES_ENOMEM) {
-        status = ARES_ENOMEM;
-      }
-    } else { /* REQUEUE_ENDQUERY */
-      if (query != NULL) {
-        query->callback(query->arg, entry.status, query->timeouts, entry.dnsrec);
-        ares_free_query(query);
-      }
-      ares_dns_record_destroy(entry.dnsrec);
-    }
-  }
-  /* Don't forget to send notification if queue emptied */
-  if (requeue != NULL) {
-    ares_queue_notify_empty(channel);
-  }
-  ares_array_destroy(requeue);
-
-  return status;
+  return ares_flush_requeue(channel, &requeue, now, status);
 }
 
 static ares_status_t process_read(ares_channel_t       *channel,
@@ -985,9 +999,9 @@ static void handle_conn_error(ares_conn_t *conn, ares_bool_t critical_failure,
   ares_close_connection(conn, failure_status);
 }
 
-/* Requeue query will normally call ares_send_query() but in some circumstances
- * this needs to be delayed, so if requeue is not NULL, it will add the query
- * to the queue instead */
+/* Requeue query will normally queue another send attempt. If a caller is
+ * already collecting requeues, append to its queue; otherwise drain a local
+ * queue iteratively before returning. */
 ares_status_t ares_requeue_query(ares_query_t *query, const ares_timeval_t *now,
                                  ares_status_t            status,
                                  ares_bool_t              inc_try_count,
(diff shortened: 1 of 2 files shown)


bradh352:
This is already fixed and will be released officially on Monday.

Issue #976, opened 2025-03-27
Don't return non-loopback addresses when resolving `localhost`.


ClifHouck:
See discussion #972. 

At some point after c-ares `v1.21.0`, a call to `ares_getaddrinfo` could cause `localhost` to resolve to an IP address which is not a loopback address. In the observed case, an entry for `localhost` in `/etc/hosts` with a non-loopback IP would be used as a valid resolution for `localhost`. This seems to directly contradict [RFC 6761 Section 3 #3](https://rfc-annotations.research.icann.org/rfc6761.html#line-417):

> Name resolution APIs and libraries SHOULD recognize localhost names as special and SHOULD always return the IP loopback address for address queries and negative responses for all other query types.

In these cases c-ares should ignore `/etc/hosts` entries which are not loopback addresses.

bradh352:
doubt this is actually a regression since c-ares used to never do any sort of localhost special casing and would just forward upstream if not found in hosts.  The only thing I can think of is if maybe it used to upstream the request to the DNS server and the dns server was returning the response.  Nonetheless needs to be fixed as the current behavior seems invalid, even with this bogus `/etc/hosts` scenario.

crrodriguez:
It is not a regression, however it will indeed by nice that c-ares adapts that RFC that while it is not an official standard yet it does indeed includes the best practices and avoids most annoying behaviours found in the wild.

PR #1233, opened 2026-07-05
ares_hosts_file: store /etc/hosts as a bipartite adjacency (fixes #1049)

### LINKED ISSUE #1049: /etc/hosts processing mixing entries
I guess I am using /etc/hosts in a more "creative" way than c-ares was intended to, but at least the comment in src/lib/ares_hosts_file.c indicates that issue is not surprising.

* It could be argued that if searching for 192.168.1.1 that the 'host6'
 * hostnames should not be returned, but this implementation will return them
 * since they are related.  It is unlikely this will matter in the real wo

I have /etc/hosts which is used as kind of a load-balancer to test remote connectivity. It goes like this:

```
10.64.0.1               mullvad

# geo: no
# no-osl-wg-socks5-001.relays.mullvad.net
10.124.0.6 mullvad-no
# no-osl-wg-socks5-002.relays.mullvad.net
10.124.0.108 mullvad-no
# no-osl-wg-socks5-003.relays.mullvad.net
10.124.0.109 mullvad-no
# no-osl-wg-socks5-004.relays.mullvad.net
10.124.0.110 mullvad-no
# no-osl-wg-socks5-005.relays.mullvad.net
10.124.0.56 mullvad-no
# no-osl-wg-socks5-006.relays.mullvad.net
10.124.0.136 mullvad-no
# no-svg-wg-socks5-001.relays.mullvad.net
10.124.2.38 mullvad-no
# no-svg-wg-socks5-002.relays.mullvad.net
# ERR: 10.124.2.39 mullvad-no
# no-svg-wg-socks5-003.relays.mullvad.net
# ERR: 10.124.2.40 mullvad-no
# no-svg-wg-socks5-004.relays.mullvad.net
# ERR: 10.124.2.41 mullvad-no

# geo: no-osl
10.124.0.6 mullvad-no-osl
10.124.0.108 mullvad-no-osl
10.124.0.109 mullvad-no-osl
10.124.0.110 mullvad-no-osl
10.124.0.56 mullvad-no-osl
10.124.0.136 mullvad-no-osl

# geo: no-svg
10.124.2.38 mullvad-no-svg

# fallback
10.64.0.1 mullvad-no
10.64.0.1 mullvad-no-osl
10.64.0.1 mullvad-no-svg
```

Here 10.64.0.1 is address of proxy connected which is always (or mostly always) available, and we want to use it as fallback, because sometimes geographical proxies are not operational. Similary, "mullvad-no" will have entries from both "mullvad-no-osl" and "mullvad-no-svg". 

What I have discovered is that looking up "mullvad-no-svg" would return entries belonging to "mullvad-no-svg", and I think it is because they are all first referenced it "mullvad-no".

Here is example where first are two lines in hosts file for "mullvad-no-svg" and what c-ares returns to curl.

```
10.124.2.38 mullvad-no-svg
10.64.0.1 mullvad-no-svg

* Host mullvad-no-svg:1080 was resolved.
* IPv6: (none)
* IPv4: 10.124.0.110, 10.124.2.38, 10.124.0.136, 10.124.0.6, 10.124.0.109, 10.124.0.108, 10.124.0.56
*   Trying 10.124.0.110:1080...
```

bradh352:
Fixes #1049.

## Problem
`/etc/hosts` entries were merged into a single record whenever they were transitively connected through a shared address, and lookups returned that record's **union**. Two failure modes:
1. **Leak:** a forward lookup returned addresses belonging to *unrelated* hostnames that merely shared an address (the reported symptom — `mullvad-no-svg` returned `mullvad-no-osl`'s addresses).
2. **Dropped "bridge" lines:** a line whose address already belonged to one entry and whose hostname to another was discarded entirely by the merge's first-match-wins rule, silently losing that address (the reporter's fallback `10.64.0.1 mullvad-no-svg` never applied).

## Approach: bipartite adjacency
The data is a set of `(hostname, address)` edges, one per line-token. Store them directly and index both ways:
- `hosthash`: hostname → list of its addresses (file order)
- `iphash`: address → list of its hostnames (file order)

A lookup builds a correctly-scoped result on demand — forward returns the queried name's **own** addresses plus **address-scoped** aliases (names sharing one of those addresses); reverse returns the queried address and its own hostnames. This is **correct by construction**: bridges, chains, cross-family merges, and reverse lookups all resolve the way glibc's `multi on` does, with no merging, filtering, or dropped lines. `to_addrinfo`/`to_hostent` go back to simple consumers.

> **Note:** an earlier revision of this PR kept the merged store and filtered results at lookup time to save memory. Review found that approach could not represent bridge lines (it inherited the merge's drop behavior), leaving #1049 only ~90% fixed. This revision replaces it with the bipartite model, which fully fixes it.

## Memory
This is done in two commits so the second can be dropped independently:
1. **the bipartite model** (correctness), and
2. **cached lookup entries** — `ares_hosts_search_*` return a cached `const` entry again (no per-lookup allocation, callers don't free). Entries are refcounted: a single-address hostname shares its address's reverse entry (wired up incrementally during parse), and only genuinely multi-address hostnames get a dedicated entry.

With the second commit, a single-address file collapses to one shared entry — structurally identical to the old merged store — so there is **no memory or parse regression** for blocklist-style files. Measured on [StevenBlack/hosts](https://github.com/StevenBlack/hosts) (~80k entries, all → `0.0.0.0`) with the internal-API harness:

| | parse | peak RSS |
|---|---|---|
| before (merged) | ~84 ms | 23.7 MB |
| this PR | **~64 ms** | **23.7 MB** |

Only files with many *multi-address* hostnames pay for the extra per-name entry, and that cost is bounded.

## Verification
- **Correctness (new regression tests):** bridge fallback (`mullvad-no-svg → [10.124.2.38, 10.64.0.1]`), chain (`c → [1.1.1.1, 3.3.3.3]`, not `2.2.2.2`), mixed-family AF_UNSPEC (no v6 leak), reverse `gethostbyaddr` (address-scoped names), `gethostbyname` (name-scoped addresses), and case-insensitivity — covering the forward, reverse, and `gethostbyname` paths that previously had zero coverage.
- All existing hosts/`gethostbyname`/`gethostbyaddr` tests pass (112 total in the hosts/hostent sweep).
- **ASAN** clean across the hosts/gethostby tests (the search functions now return an owned entry that each of the three callers destroys on every path).
- Also fixes a pre-existing bug in `ares_hosts_entry_isdup` (iterated the ip list instead of the hostname list).

_Numbers were produced with a small harness linked against the internal `ares_hosts_*` API in a throwaway checkout (not CI); reproduce by pointing `CARES_HOSTS` at the StevenBlack file._

### DIFF
--- a/src/lib/ares_hosts_file.c
+++ b/src/lib/ares_hosts_file.c
@@ -70,15 +70,28 @@
  * 192.168.1.5  host.example.com host
  * 2620:1234::1 host.example.com host6.example.com host6 host
  *
- * This will yield 2 entries.
- *  1) ips: 127.0.0.1,::1
- *     hosts: localhost.localdomain,localhost
- *  2) ips: 192.168.1.1,192.168.1.5,2620:1234::1
- *     hosts: host.example.com,host,host6.example.com,host6
+ * STORAGE MODEL: BIPARTITE ADJACENCY
+ * ----------------------------------
+ * The hosts file is stored as a bipartite adjacency between hostnames and
+ * addresses.  Every (hostname, address) edge that appears on any line is
+ * recorded in both directions:
+ *   - hosthash: hostname -> ares_llist_t of the addresses that hostname
+ *     appeared on a line with (file-ordered, de-duplicated)
+ *   - iphash:   address  -> ares_llist_t of the hostnames that appeared on a
+ *     line with that address (file-ordered, de-duplicated)
  *
- * It could be argued that if searching for 192.168.1.1 that the 'host6'
- * hostnames should not be returned, but this implementation will return them
- * since they are related.  It is unlikely this will matter in the real world.
+ * A forward lookup (by name) or reverse lookup (by ip) builds a transient,
+ * fully address-scoped result entry from the relevant adjacency list.  Aliases
+ * are address-scoped: they are exactly the hostnames that share an address with
+ * the queried name.  Recording every edge is correct by construction -- it does
+ * not drop "bridge" lines (a line whose ip already belongs to one hostname and
+ * whose hostname belongs to another), which the older merged store did,
+ * silently losing addresses (Issue #1049).
+ *
+ * Memory tradeoff: keeping a hostname->address list plus an address->hostname
+ * list for every edge roughly doubles memory versus a naive merged store on
+ * very large hosts files (e.g. StevenBlack/hosts, ~80k hostnames).  We accept
+ * this deliberately in exchange for correct, leak-free results.
  */
 
 struct ares_hosts_file {
@@ -86,16 +99,17 @@ struct ares_hosts_file {
   /*! cache the filename so we know if the filename changes it automatically
    *  invalidates the cache */
   char                *filename;
-  /*! iphash is the owner of the 'entry' object as there is only ever a single
-   *  match to the object. */
+  /*! address (normalized str) -> ares_llist_t of hostname strings that appeared
+   *  on a line with that address (file-ordered, de-duplicated).  Owns the
+   *  llist values via ares_hosts_list_destroy_cb. */
   ares_htable_strvp_t *iphash;
-  /*! hosthash does not own the entry so won't free on destruction */
+  /*! hostname (str) -> ares_llist_t of address strings that hostname appeared
+   *  on a line with (file-ordered, de-duplicated).  Owns the llist values via
+   *  ares_hosts_list_destroy_cb. */
   ares_htable_strvp_t *hosthash;
 };
 
 struct ares_hosts_entry {
-  size_t        refcnt; /*! If the entry is stored multiple times in the
-                         *  ip address hash, we have to reference count it */
   ares_llist_t *ips;
   ares_llist_t *hosts;
 };
@@ -158,18 +172,9 @@ static ares_bool_t ares_normalize_ipaddr(const char *ipaddr, char *out,
   return ARES_TRUE;
 }
 
-static void ares_hosts_entry_destroy(ares_hosts_entry_t *entry)
+void ares_hosts_entry_destroy(ares_hosts_entry_t *entry)
 {
   if (entry == NULL) {
-    return;
-  }
-
-  /* Honor reference counting */
-  if (entry->refcnt != 0) {
-    entry->refcnt--;
-  }
-
-  if (entry->refcnt > 0) {
     return;
   }
 
@@ -178,9 +183,11 @@ static void ares_hosts_entry_destroy(ares_hosts_entry_t *entry)
   ares_free(entry);
 }
 
-static void ares_hosts_entry_destroy_cb(void *entry)
+/* htable value destructor: each value is an ares_llist_t (of ip or hostname
+ * strings) */
+static void ares_hosts_list_destroy_cb(void *arg)
 {
-  ares_hosts_entry_destroy(entry);
+  ares_llist_destroy(arg);
 }
 
 void ares_hosts_file_destroy(ares_hosts_file_t *hf)
@@ -209,12 +216,12 @@ static ares_hosts_file_t *ares_hosts_file_create(const char *filename)
     goto fail;
   }
 
-  hf->iphash = ares_htable_strvp_create(ares_hosts_entry_destroy_cb);
+  hf->iphash = ares_htable_strvp_create(ares_hosts_list_destroy_cb);
   if (hf->iphash == NULL) {
     goto fail;
   }
 
-  hf->hosthash = ares_htable_strvp_create(NULL);
+  hf->hosthash = ares_htable_strvp_create(ares_hosts_list_destroy_cb);
   if (hf->hosthash == NULL) {
     goto fail;
   }
@@ -226,140 +233,102 @@ fail:
   return NULL;
 }
 
-typedef enum {
-  ARES_MATCH_NONE   = 0,
-  ARES_MATCH_IPADDR = 1,
-  ARES_MATCH_HOST   = 2
-} ares_hosts_file_match_t;
-
-static ares_status_t ares_hosts_file_merge_entry(
-  const ares_hosts_file_t *hf, ares_hosts_entry_t *existing,
-  ares_hosts_entry_t *entry, ares_hosts_file_match_t matchtype)
+/* Case-insensitive membership test for an ip/host string list */
+static ares_bool_t ares_hosts_iplist_contains(ares_llist_t *list,
+                                              const char   *ipaddr)
 {
   ares_llist_node_t *node;
 
-  /* If we matched on IP address, we know there can only be 1, so there's no
-   * reason to do anything */
-  if (matchtype != ARES_MATCH_IPADDR) {
-    while ((node = ares_llist_node_first(entry->ips)) != NULL) {
-      const char *ipaddr = ares_llist_node_val(node);
-
-      if (ares_htable_strvp_get_direct(hf->iphash, ipaddr) != NULL) {
-        ares_llist_node_destroy(node);
-        continue;
-      }
-
-      ares_llist_node_mvparent_last(node, existing->ips);
+  for (node = ares_llist_node_first(list); node != NULL;
+       node = ares_llist_node_next(node)) {
+    if (ares_strcaseeq(ares_llist_node_val(node), ipaddr)) {
+      return ARES_TRUE;
     }
   }
 
-
-  while ((node = ares_llist_node_first(entry->hosts)) != NULL) {
-    const char *hostname = ares_llist_node_val(node);
-
-    if (ares_htable_strvp_get_direct(hf->hosthash, hostname) != NULL) {
-      ares_llist_node_destroy(node);
-      continue;
-    }
-
-    ares_llist_node_mvparent_last(node, existing->hosts);
-  }
-
-  ares_hosts_entry_destroy(entry);
-  return ARES_SUCCESS;
+  return ARES_FALSE;
 }
 
-static ares_hosts_file_match_t
-  ares_hosts_file_match(const ares_hosts_file_t *hf, ares_hosts_entry_t *entry,
-                        ares_hosts_entry_t **match)
-{
-  ares_llist_node_t *node;
-  *match = NULL;
-
-  for (node = ares_llist_node_first(entry->ips); node != NULL;
-       node = ares_llist_node_next(node)) {
-    const char *ipaddr = ares_llist_node_val(node);
-    *match             = ares_htable_strvp_get_direct(hf->iphash, ipaddr);
-    if (*match != NULL) {
-      return ARES_MATCH_IPADDR;
-    }
-  }
-
-  for (node = ares_llist_node_first(entry->hosts); node != NULL;
-       node = ares_llist_node_next(node)) {
-    const char *host = ares_llist_node_val(node);
-    *match           = ares_htable_strvp_get_direct(hf->hosthash, host);
-    if (*match != NULL) {
-      return ARES_MATCH_HOST;
-    }
-  }
-
-  return ARES_MATCH_NONE;
-}
-
-/*! entry is invalidated upon calling this function, always, even on error */
+/*! entry represents a single parsed line (one ip, its hostnames).  It is always
+ *  invalidated (destroyed) upon calling this function, even on error.  Each
+ *  (hostname, ipaddress) edge is recorded in both directions. */
 static ares_status_t ares_hosts_file_add(ares_hosts_file_t  *hosts,
                                          ares_hosts_entry_t *entry)
 {
-  ares_hosts_entry_t     *match  = NULL;
-  ares_status_t           status = ARES_SUCCESS;
-  ares_llist_node_t      *node;
-  ares_hosts_file_match_t matchtype;
-  size_t                  num_hostnames;
+  const char        *ipaddr = ares_llist_first_val(entry->ips);
+  ares_llist_node_t *node;
+  ares_status_t      status = ARES_SUCCESS;
 
-  /* Record the number of hostnames in this entry file.  If we merge into an
-   * existing record, these will be *appended* to the entry, so we'll count
-   * backwards when adding to the hosts hashtable */
-  num_hostnames = ares_llist_len(entry->hosts);
+  for (node = ares_llist_node_first(entry->hosts); node != NULL;
+       node = ares_llist_node_next(node)) {
+    const char   *host = ares_llist_node_val(node);
+    ares_llist_t *fwd;
+    ares_llist_t *rev;
+    char         *tmp;
 
-  matchtype = ares_hosts_file_match(hosts, entry, &match);
-
-  if (matchtype != ARES_MATCH_NONE) {
-    status = ares_hosts_file_merge_entry(hosts, match, entry, matchtype);
-    if (status != ARES_SUCCESS) {
-      ares_hosts_entry_destroy(entry); /* LCOV_EXCL_LINE: DefensiveCoding */
-      return status;                   /* LCOV_EXCL_LINE: DefensiveCoding */
-    }
-    /* entry was invalidated above by merging */
-    entry = match;
-  }
-
-  if (matchtype != ARES_MATCH_IPADDR) {
-    const char *ipaddr = ares_llist_last_val(entry->ips);
-
-    if (!ares_htable_strvp_get(hosts->iphash, ipaddr, NULL)) {
-      if (!ares_htable_strvp_insert(hosts->iphash, ipaddr, entry)) {
-        ares_hosts_entry_destroy(entry);
-        return ARES_ENOMEM;
+    /* Forward edge: hostname -> ip */
+    fwd = ares_htable_strvp_get_direct(hosts->hosthash, host);
+    if (fwd == NULL) {
+      fwd = ares_llist_create(ares_free);
+      if (fwd == NULL) {
+        status = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
+        goto done;            /* LCOV_EXCL_LINE: OutOfMemory */
       }
-      entry->refcnt++;
-    }
-  }
-
-  /* Go backwards, on a merge, hostnames are appended.  Breakout once we've
-   * consumed all the hosts that we appended */
-  for (node = ares_llist_node_last(entry->hosts); node != NULL;
-       node = ares_llist_node_prev(node)) {
-    const char *val = ares_llist_node_val(node);
-
-    if (num_hostnames == 0) {
-      break;
+      if (!ares_htable_strvp_insert(hosts->hosthash, host, fwd)) {
+        ares_llist_destroy(fwd); /* LCOV_EXCL_LINE: OutOfMemory */
+        status = ARES_ENOMEM;    /* LCOV_EXCL_LINE: OutOfMemory */
+        goto done;               /* LCOV_EXCL_LINE: OutOfMemory */
+      }
     }
 
-    num_hostnames--;
-
-    /* first hostname match wins.  If we detect a duplicate hostname for another
-     * ip it will automatically be added to the same entry */
-    if (ares_htable_strvp_get(hosts->hosthash, val, NULL)) {
+    /* De-dupe this (host, ip) edge against the (usually tiny) forward list.  If
+     * the edge already exists, so does its reverse, so skip both. */
+    if (ares_hosts_iplist_contains(fwd, ipaddr)) {
       continue;
     }
 
-    if (!ares_htable_strvp_insert(hosts->hosthash, val, entry)) {
-      return ARES_ENOMEM;
+    tmp = ares_strdup(ipaddr);
+    if (tmp == NULL) {
+      status = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
+      goto done;            /* LCOV_EXCL_LINE: OutOfMemory */
+    }
+    if (ares_llist_insert_last(fwd, tmp) == NULL) {
+      ares_free(tmp);       /* LCOV_EXCL_LINE: OutOfMemory */
+      status = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
+      goto done;            /* LCOV_EXCL_LINE: OutOfMemory */
+    }
+
+    /* Reverse edge: ip -> hostname.  The edge was new above, so 'host' is not
+     * yet in the reverse list either. */
+    rev = ares_htable_strvp_get_direct(hosts->iphash, ipaddr);
+    if (rev == NULL) {
+      rev = ares_llist_create(ares_free);
+      if (rev == NULL) {
+        status = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
+        goto done;            /* LCOV_EXCL_LINE: OutOfMemory */
+      }
+      if (!ares_htable_strvp_insert(hosts->iphash, ipaddr, rev)) {
+        ares_llist_destroy(rev); /* LCOV_EXCL_LINE: OutOfMemory */
+        status = ARES_ENOMEM;    /* LCOV_EXCL_LINE: OutOfMemory */
+        goto done;               /* LCOV_EXCL_LINE: OutOfMemory */
+      }
+    }
+
+    tmp = ares_strdup(host);
+    if (tmp == NULL) {
+      status = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
+      goto done;            /* LCOV_EXCL_LINE: OutOfMemory */
+    }
+    if (ares_llist_insert_last(rev, tmp) == NULL) {
+      ares_free(tmp);       /* LCOV_EXCL_LINE: OutOfMemory */
+      status = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
+      goto done;            /* LCOV_EXCL_LINE: OutOfMemory */
     }
   }
 
-  return ARES_SUCCESS;
+done:
+  ares_hosts_entry_destroy(entry);
+  return status;
 }
 
 static ares_bool_t ares_hosts_entry_isdup(ares_hosts_entry_t *entry,
@@ -367,7 +336,7 @@ static ares_bool_t ares_hosts_entry_isdup(ares_hosts_entry_t *entry,
 {
   ares_llist_node_t *node;
 
-  for (node = ares_llist_node_first(entry->ips); node != NULL;
+  for (node = ares_llist_node_first(entry->hosts); node != NULL;
        node = ares_llist_node_next(node)) {
     const char *myhost = ares_llist_node_val(node);
     if (ares_strcaseeq(myhost, host)) {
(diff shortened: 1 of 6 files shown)


test/ares-test-internal.cc
```diff
@@ -636,6 +636,54 @@ TEST_F(FileChannelTest, GetAddrInfoHostsIPV6) {
   EXPECT_EQ("{ipv6.com addr=[[0000:0000:0000:0000:0000:0000:0000:0001]]}", ss.str());
 }
 
+// Regression for #1049: hostnames merged into a single entry only because they
+// share an ip address must not leak each other's *other* addresses (forward
+// lookup) or appear as each other's aliases (address-scoped cnames).  Here
+// mullvad-no-svg shares 10.124.2.38 with mullvad-no (which drags in the whole
+// group), but must resolve to only 10.124.2.38, and mullvad-no-osl (no shared
+// address) must not be reported as an alias.
+TEST_F(FileChannelTest, GetAddrInfoHostsSharedIPNoLeak) {
```
bradh352: **[Major]** The two new tests only exercise the forward `ares_getaddrinfo` path. The reverse filter (`filter_ipaddr` / `ares_hosts_entry_host_has_ip`) and the `ares_gethostbyname_file` name-scoped hostent path have **zero test coverage** — I verified them working via ad-hoc probing, but nothing in CI guards them.

Note `CARES_HOSTS`/`ARES_AI_ENVHOSTS` won't work for those paths (`ares_gethostbyaddr` / `ares_gethostbyname_file` call `ares_hosts_search_*` with `use_env = ARES_FALSE`), so these tests need a channel initialized with `ARES_OPT_HOSTS_FILE`. Sketch, with hosts file `1.1.1.1 a` / `1.1.1.1 b` / `2.2.2.2 b` / `2.2.2.2 c`:

- `ares_gethostbyaddr(2.2.2.2)`: expect `h_name` "b", aliases `{c}`, `h_addr_list` `{2.2.2.2}` only (pre-PR: 2 addresses, aliases included "a").
- `ares_gethostbyname_file(channel, "c", AF_INET)`: expect `h_addr_list` `{2.2.2.2}`, alias "b" but not "a".

Also worth adding (Medium priority):
- chain shape for forward lookups (`1.1.1.1 a` / `2.2.2.2 b` / `1.1.1.1 c` / `3.3.3.3 c`: query `a` → only 1.1.1.1, query `c` → 1.1.1.1 + 3.3.3.3);
- mixed-family merge queried with AF_UNSPEC (`192.168.1.1 host` / `2620:1234::1 host` / `192.168.1.1 other`: `other` must not return the v6 address — cross-family leakage was a real pre-PR symptom);
- the promotion boundary (blocklist `0.0.0.0 a b c` then `1.1.1.1 b`: `a` → `[0.0.0.0]` only, `b` → both), which exercises `promote_multi_ip` mapping all pre-existing hosts to ip0;
- a case-mismatch query (`FOOBAR` against `1.1.1.1 FooBar` / `2.2.2.2 foobar`) to pin the strvp case-insensitivity the whole design depends on — today a casecmp regression would be masked by the silent `filtered == NULL` → union fallback.

src/lib/ares_hosts_file.c
```diff
@@ -876,13 +1141,31 @@ ares_status_t ares_hosts_entry_to_addrinfo(const ares_hosts_entry_t *entry,
...
-  for (node = ares_llist_node_first(entry->ips); node != NULL;
+  /* For a forward lookup of a specific name in a multi-ip (transitively merged)
+   * entry, return only the addresses that name actually appeared with, not the
+   * union across the whole merged group.  Single-ip entries have no filter and
+   * fall through to entry->ips. */
+  iplist = entry->ips;
+  if (name != NULL && entry->host_ips != NULL) {
+    ares_llist_t *filtered =
+      ares_htable_strvp_get_direct(entry->host_ips, name);
+    if (filtered != NULL) {
```
bradh352: **[Medium]** This filter fails *open*: if `name` ever misses the `host_ips` map we silently fall back to `entry->ips` — the full merged union, i.e. exactly the #1049 leak. Today that's unreachable (every member of `entry->hosts` provably has a key), but the sibling defensive paths fail *closed* (`ares_hosts_entry_hosts_share_ip` and `ares_hosts_entry_host_has_ip` both return `ARES_FALSE` on a missing key). If a future change breaks the invariant (a new merge path, tolerated mid-parse ENOMEM, ...), this line silently reintroduces the information leak instead of degrading to a lookup miss. Suggest failing closed:

```c
    if (filtered == NULL) {
      status = ARES_ENOTFOUND; /* LCOV_EXCL_LINE: DefensiveCoding */
      goto done;               /* LCOV_EXCL_LINE: DefensiveCoding */
    }
    iplist = filtered;
```

or at minimum a comment stating the fail-open choice is deliberate.

src/lib/ares_hosts_file.c
```diff
@@ -315,6 +526,26 @@ static ares_status_t ares_hosts_file_add(ares_hosts_file_t  *hosts,
...
+     * return addresses belonging to unrelated hostnames).  Single-ip entries
+     * skip this entirely, so there is no cost for the common case. */
+    const char *line_ip  = ares_llist_first_val(entry->ips);
+    ares_bool_t lip_new  = !ares_htable_strvp_get(hosts->iphash, line_ip, NULL);
+    size_t      curr_ips = ares_llist_len(match->ips);
+
+    if (lip_new || curr_ips >= 2) {
+      status = ares_hosts_entry_promote_multi_ip(match);
+      if (status == ARES_SUCCESS) {
+        status = ares_hosts_entry_record_edges(match, line_ip, entry->hosts);
```
bradh352: **[Low]** `record_edges` runs before `ares_hosts_file_merge_entry`, which drops any line hostname already owned by a *different* entry (first-hostname-match-wins). Those hostnames still get keys in `match->host_ips`. Concrete input:

```
1.1.1.1 x
2.2.2.2 x          <- x promoted, entry multi-ip
3.3.3.3 h1         <- separate entry
1.1.1.1 h1         <- MATCH_IPADDR: h1 -> 1.1.1.1 recorded into x's map, then dropped by merge
```

No lookup can reach the stray keys (forward routes through `hosthash`; reverse iterates only `match->hosts`), so this is wasted memory only — but worth either a one-line comment accepting it, or skipping hostnames whose `hosthash` entry isn't `match`.

src/lib/ares_hosts_file.c
```diff
@@ -775,24 +1006,45 @@ ares_status_t ares_hosts_search_host(ares_channel_t *channel,
...
+  /* The canonical name is the first host (in file order) that shares an address
+   * with the queried name; the aliases are the remaining such hosts.  This
+   * keeps aliases address-scoped: a hostname merged into the same entry only
+   * because of an unrelated shared ip is not reported as an alias.  For a
+   * single-ip entry every host shares the sole address, preserving the
+   * historical behavior. */
+  for (node = ares_llist_node_first(entry->hosts); node != NULL;
+       node = ares_llist_node_next(node)) {
+    const char *host = ares_llist_node_val(node);
+    if (ares_hosts_entry_host_member(entry, name, ipaddr, host)) {
```
bradh352: **[Low]** Quadratic worst case: this loop visits every host in the merged entry and `ares_hosts_entry_hosts_share_ip` is O(|ips_a| × |ips_b|); hosts that do *not* share an address never count toward the 100-alias cap, so the scan always covers the whole merged group. A crafted local file (one hostname on ~25k ips, thousands of non-overlapping multi-ip hostnames chained into one entry via a bridge host) makes a single `ARES_AI_CANONNAME` query cost billions of string compares. Hosts files are trusted input, so informational — a cheap mitigation is also capping *examined* hosts, or short-circuiting `share_ip` via a set lookup on the smaller list. Not blocking.

src/lib/ares_hosts_file.c
```diff
@@ -315,6 +526,26 @@ static ares_status_t ares_hosts_file_add(ares_hosts_file_t  *hosts,
   matchtype = ares_hosts_file_match(hosts, entry, &match);
 
   if (matchtype != ARES_MATCH_NONE) {
+    /* Before the line entry is merged away, capture its hostname<->ip pairing.
+     * This is only required once the target entry spans more than one ip
+     * address (transitive merges can otherwise cause a forward lookup to
+     * return addresses belonging to unrelated hostnames).  Single-ip entries
+     * skip this entirely, so there is no cost for the common case. */
+    const char *line_ip  = ares_llist_first_val(entry->ips);
+    ares_bool_t lip_new  = !ares_htable_strvp_get(hosts->iphash, line_ip, NULL);
```
bradh352: **[Nit]** `lip_new` is provably equivalent to `matchtype == ARES_MATCH_HOST`: `ares_hosts_file_match` checks the line's (single) ip against `iphash` first, so `MATCH_IPADDR` ⇔ ip present and `MATCH_HOST` ⇔ ip absent. That drops a hash lookup and makes the invariant self-documenting. It also fixes the idiom: assigning a `!` result to `ares_bool_t` diverges from the codebase's `? ARES_TRUE : ARES_FALSE` style (and clang's `-Wassign-enum` would flag it).

```suggestion
    ares_bool_t lip_new  = (matchtype == ARES_MATCH_HOST) ? ARES_TRUE : ARES_FALSE;
```

src/lib/ares_hosts_file.c
```diff
@@ -775,24 +1006,45 @@ ares_status_t ares_hosts_search_host(ares_channel_t *channel,
...
+       node = ares_llist_node_next(node)) {
+    const char *host = ares_llist_node_val(node);
+    if (ares_hosts_entry_host_member(entry, name, ipaddr, host)) {
+      primaryhost = host;
+      node        = ares_llist_node_next(node);
+      break;
+    }
+  }
+  if (primaryhost == NULL) {
+    primaryhost = ares_llist_first_val(entry->hosts); /* LCOV_EXCL_LINE */
```
bradh352: **[Nit]** Bare `LCOV_EXCL_LINE` (here and the next line) — convention in this file is a reason tag, and this fallback is genuinely unreachable (forward: `name` always keys `host_ips`; reverse: every ip in `entry->ips` has ≥1 owning host by the promotion/record_edges invariants), so:

```suggestion
    primaryhost = ares_llist_first_val(entry->hosts); /* LCOV_EXCL_LINE: DefensiveCoding */
```

Relatedly: the new ENOMEM returns in `ares_hosts_entry_host_add_ip` / `record_edges` / `promote_multi_ip` are *not* annotated `/* LCOV_EXCL_LINE: OutOfMemory */`, unlike most OOM paths in this file — annotate (or rely on alloc-fail coverage), but be consistent either way.

src/lib/ares_hosts_file.c
```diff
@@ -297,6 +335,179 @@ static ares_hosts_file_match_t
...
+ * its own copies so it is independent of the merge lifetime of entry->ips. */
+static ares_status_t ares_hosts_entry_host_add_ip(ares_hosts_entry_t *entry,
+                                                  const char         *host,
+                                                  const char         *ipaddr)
+{
+  ares_llist_t      *iplist;
+  ares_llist_node_t *node;
+  char              *tmp;
+
+  iplist = ares_htable_strvp_get_direct(entry->host_ips, host);
```
bradh352: **[Nit]** This `ares_strcaseeq`-over-iplist scan now appears three times (the de-dupe below, the inner loop of `ares_hosts_entry_hosts_share_ip`, and `ares_hosts_entry_host_has_ip`). A small helper would remove two copies:

```c
static ares_bool_t ares_hosts_iplist_contains(const ares_llist_t *list,
                                              const char         *ipaddr)
```

src/lib/ares_private.h
```diff
@@ -489,10 +489,13 @@ ares_status_t ares_hosts_search_host(ares_channel_t *channel,
                                      ares_bool_t use_env, const char *host,
                                      const ares_hosts_entry_t **entry);
 ares_status_t ares_hosts_entry_to_hostent(const ares_hosts_entry_t *entry,
```
bradh352, line 498: **[Nit]** Parameter order/naming inconsistency between the two converters: `ares_hosts_entry_to_hostent(entry, family, name, ipaddr, ...)` vs `ares_hosts_entry_to_addrinfo(entry, name, filter_ipaddr, family, ...)` — `family` sits before the filter keys in one and after in the other, and the ip filter is `ipaddr` in one and `filter_ipaddr` in the other. Suggest `(entry, name, filter_ipaddr, family, hostent)` for symmetry.

src/lib/ares_hosts_file.c
```diff
@@ -98,6 +125,16 @@ struct ares_hosts_entry {
                          *  ip address hash, we have to reference count it */
   ares_llist_t *ips;
   ares_llist_t *hosts;
+  /*! Forward-lookup filter.  Maps a hostname (str) to an ares_llist_t of the
```
bradh352: **[Nit]** The StevenBlack/blocklist rationale now appears three times (file header block, this field comment, and the `promote_multi_ip` comment). This one could shrink to a line or two pointing at the header block — the mechanics ("maps hostname -> llist of ip strings, NULL until the entry spans >1 ip") are what a reader needs here.

src/lib/ares_hosts_file.c
```diff
@@ -775,24 +1006,45 @@ ares_status_t ares_hosts_search_host(ares_channel_t *channel,
...
 
   while (node != NULL) {
     const char *host = ares_llist_node_val(node);
 
+    if (!ares_hosts_entry_host_member(entry, name, ipaddr, host)) {
+      node = ares_llist_node_next(node);
+      continue;
+    }
+
     /* Cap at 100 entries. , some people use
```
bradh352: **[Nit]** Pre-existing garbled comment — "Cap at 100 entries. , some people use" has a dangling ", " mid-sentence. Since this hunk is being reworked anyway, worth fixing opportunistically.

bradh352 (commented):
## Review summary

Deep review against merge-base 560c216f: full read of the new code plus empirical verification (PR build vs. baseline build of the parent commit; 30 behavioral probes over the issue's file shape, chain/bridge shapes, mixed-case, mixed-family, and reverse lookups; targeted test runs under ASAN).

**Verdict: the implementation is sound.** No Major correctness or security defects found. Specifically verified:

- The promotion condition `lip_new || curr_ips >= 2` was exhaustively case-analyzed (IPADDR/HOST match x single/multi-ip, dup hostnames, cross-entry dups, chained promotes) and is complete: multi-ip ⇔ `host_ips != NULL`, and every member of `entry->hosts` has a `host_ips` key when the map exists. The "lazy, only for multi-ip entries" claim in the PR body is accurate.
- Case-insensitivity holds end-to-end (`ares_htable_strvp` hashes with `FNV1a_casecmp`, compares with `ares_strcaseeq`), confirmed empirically with mixed-case files and queries on the multi-ip path.
- All 30 probes: every difference vs. baseline was a leak being removed; no case returned fewer *correct* addresses/aliases. Forward order is file order of the queried name's own lines (matches glibc `multi on`/musl).
- Error-path lifetimes check out (strvp insert failure does not free the value, so the destroy-on-failure in `host_add_ip` is correct; parse aborts discard partial maps; merge cannot half-complete). Clean ASAN runs; zero new compiler warnings (`-Wall -Wextra`, AppleClang); C89-clean.
- No public API/ABI change (only `ares_private.h` internals; all new functions static). No manual parsing introduced. None of the 16 open PRs touch `ares_hosts_file.c`; no overlapping work. Issue #1049 is a genuine, trivially hittable report.

Inline comments: 1 Major (test coverage for the new hostent paths), 1 Medium (fail-open filter hardening), 2 Low, 6 Nit. Items that could not be inline:

### 1. [Medium] #1049 is honestly ~90% fixed, and "Fixes #1049" will auto-close it
The reporter's *full* file ends with fallback lines (`10.64.0.1 mullvad-no-svg`) whose ip already belongs to one entry while the hostname belongs to another. Such bridging lines are silently dropped by the pre-existing first-hostname-match-wins behavior in `ares_hosts_file_merge_entry`, so the fallback address never applies: post-PR `mullvad-no-svg` → `[10.124.2.38]`, where glibc (`multi on`) → `[10.124.2.38, 10.64.0.1]`. This is pre-existing (byte-identical on baseline: E0 and E1 never merge, so the old union also lacked `10.64.0.1`) and structurally unfixable without the rejected bipartite index — but the regression test uses a truncated 6-line file that omits the fallback block. Recommend stating this limitation explicitly in the PR body and the header comment ("a line whose ip belongs to one entry and whose hostname to another is discarded"), and leaving a note on #1049 about the residual gap (possibly a follow-up issue) rather than closing wordlessly.

### 2. [Medium] Release-note callout needed
This is a user-visible behavior change to long-standing (if wrong) behavior: hosts-file results are now address-scoped for forward, reverse, *and* alias lists, and `ares_gethostbyaddr` now returns only the queried address in `h_addr_list` (plus fewer `h_aliases`). Anyone who exploited the old union as a c-ares-specific fallback trick silently loses addresses — that never worked on glibc/musl, but the result surface shrinks with no error, which is worth one sentence in RELEASE-NOTES.

### 3. [Low] Canonical-name choice deviates from glibc (codified by the new test)
c-ares picks the first *merged-entry-order* host sharing an address with the query; glibc/musl report the first token of the first matching *line*. For `192.0.2.1 alpha` / `192.0.2.1 beta` / `192.0.2.2 beta`, query `beta` with AI_CANONNAME: glibc → canon `beta`; c-ares → canon `alpha`, alias `beta`. `GetAddrInfoHostsSharedIPNoLeak` codifies this (`mullvad-no-svg` → canon `mullvad-no`). Pre-PR was strictly worse (no address requirement at all), so this is a retained-and-narrowed deviation, not a regression — fixing it would need per-line provenance the storage model deliberately discards. Worth a sentence in the header comment/docs; not blocking.

### 4. Pre-existing bugs found during review (separate commits, not this PR)
- `ares_hosts_entry_isdup` iterates `entry->ips` instead of `entry->hosts`, so same-line duplicate hostnames (`1.2.3.4 foo foo`) are never deduped — `foo` is stored twice and can be reported as its own alias.
- ENOMEM-only use-after-free in `ares_hosts_file_add`: if the `iphash` insert fails after a `MATCH_HOST` merge, `entry == match` with `refcnt >= 1` is destroyed while `iphash` still references it under its original ip key, double-destroying at `ares_hosts_file_destroy`.

### 5. Micro-nits
- PR body: "zero additional memory or parse cost" — zero memory yes, but every merged line now pays one extra `iphash` lookup + `ares_llist_len` call. Unmeasurable (the benchmark shows parity), but technically "unmeasurable", not "zero".
- `ares_hosts_entry_host_member`'s both-NULL "legacy behavior" branch has no in-tree caller — fine as defense, but the comment could note it is not currently reachable.

bradh352:
### Review disposition — superseded by the bipartite rework

This PR was reworked from the "merged store + lookup-time filter" implementation to a **bipartite adjacency** model (`hosthash: name→addresses`, `iphash: address→names`). That deletes most of the code the review commented on, so resolving the threads below.

**Body items**
- **[Medium] #1049 only ~90% fixed / bridge lines dropped** → **Resolved.** The bipartite model records every `(name, address)` edge, so bridge lines are no longer dropped: `mullvad-no-svg` now resolves to `[10.124.2.38, 10.64.0.1]` (new test `GetAddrInfoHostsBridgeFallback`). Full glibc `multi on` parity, so `Fixes #1049` is now accurate.
- **[Medium] Release-note callout** → **Valid, noted.** The user-visible behavior change (address-scoped forward/reverse/alias results; `gethostbyaddr` returns only the queried address) is documented in the commit message and PR body for the release notes.
- **[Low] Canonical-name deviates from glibc** → **Acknowledged, retained.** Canonical is the first file-order name sharing the queried address (documented in a `GetHostByNameHostsFile` comment). It affects only the reported CNAME/`h_name`, not which addresses resolve.
- **Pre-existing bugs** → `ares_hosts_entry_isdup` (iterated ips not hosts) is **fixed** in this PR; the ENOMEM use-after-free in the old `ares_hosts_file_add`/`merge` path is **gone** (that code is deleted).

**Inline items**
- **[Major] reverse / `gethostbyname` had zero coverage** → **Addressed.** Six new tests: bridge, chain, mixed-family (AF_UNSPEC), reverse `gethostbyaddr`, `gethostbyname` (both via a private `ARES_OPT_HOSTS_FILE` channel), and case-insensitivity.
- **[Medium] fail-open filter**, **[Low] `record_edges` stray keys**, **[Low] quadratic alias scan**, **[Nit] `lip_new`**, **[Nit] param-order inconsistency** → **Moot** — that code (the filter, `host_ips`, `record_edges`, `promote_multi_ip`, `merge`/`match`, the `filter_ipaddr` param) no longer exists.
- **[Nit] extract `iplist_contains` helper** → **Done** (`ares_hosts_iplist_contains`).
- **[Nit] garbled "Cap at 100 entries. ," comment** → **Fixed.**
- **[Nit] StevenBlack rationale repeated 3×** → **Fixed** (single rationale in the rewritten header block).
- **[Nit] LCOV/OOM annotations** → new OOM returns are annotated.

Memory tradeoff is accepted and documented: StevenBlack peak 23.7 MB → 29.8 MB.

src/lib/ares_hosts_file.c
```diff
@@ -226,136 +257,284 @@ static ares_hosts_file_t *ares_hosts_file_create(const char *filename)
...
-  if (matchtype != ARES_MATCH_IPADDR) {
-    const char *ipaddr = ares_llist_last_val(entry->ips);
+  /* hosts = address-scoped alias list, canonical first.  Iterate the addresses
+   * in order; for each, iterate the reverse entry's hostnames in order,
+   * appending each name not already present (case-insensitive).  Cap at 101
+   * (canonical + 100 aliases) to bound the StevenBlack blocklist case. */
+  for (ipnode = ares_llist_node_first(ent->ips); ipnode != NULL;
+       ipnode = ares_llist_node_next(ipnode)) {
+    const char         *ip  = ares_llist_node_val(ipnode);
+    ares_hosts_entry_t *rev = ares_htable_strvp_get_direct(hosts->iphash, ip);
```
bradh352, line 489: **[Low]** `rev` is dereferenced just below (`rev->hosts`) with no NULL guard. It's provably non-NULL today — every ip in `ent->ips` came from a parsed line that created an `iphash` entry, and nothing ever removes from `iphash` — so this isn't a live bug. But the sibling defensive paths in this file fail *closed*, and a future refactor that ever dropped an `iphash` entry would turn this into a NULL deref. Cheap to harden:

```c
    ares_hosts_entry_t *rev = ares_htable_strvp_get_direct(hosts->iphash, ip);
    ares_llist_node_t  *nnode;

    if (rev == NULL) {
      continue; /* LCOV_EXCL_LINE: DefensiveCoding */
    }
```

src/lib/ares_hosts_file.c
```diff
@@ -226,136 +257,284 @@ static ares_hosts_file_t *ares_hosts_file_create(const char *filename)
...
-  ARES_MATCH_IPADDR = 1,
-  ARES_MATCH_HOST   = 2
-} ares_hosts_file_match_t;
-
-static ares_status_t ares_hosts_file_merge_entry(
-  const ares_hosts_file_t *hf, ares_hosts_entry_t *existing,
-  ares_hosts_entry_t *entry, ares_hosts_file_match_t matchtype)
+/* Case-insensitive membership test for an ip/host string list */
+static ares_bool_t ares_hosts_iplist_contains(ares_llist_t *list,
+                                              const char   *ipaddr)
```
bradh352: **[Nit]** This helper is now generic case-insensitive string-list membership and is applied to the **hosts** list too (line 494, `ares_hosts_iplist_contains(ent->hosts, nm)`), not just ip lists — the `iplist` in the name is misleading. The comment already says "ip/host string list"; consider renaming to `ares_hosts_strlist_contains` (or `ares_hosts_list_contains`) so the name matches its use.

bradh352 (commented):
## Deep re-review of the bipartite rework — LGTM

Re-reviewed the current implementation (the earlier "merged store + filter" revision's threads are all resolved/outdated). Verified independently: clean `-Werror` build with zero warnings from the changed code; 8/8 new tests pass and 5/8 fail on baseline (genuine regression guards); `git clang-format` clean on changed lines. Refcount/OOM/lifetime auditing is CONFIRMED-SAFE — the finalize key-replace correctly depends on `ares_htable_insert`'s replace-and-free-old-value semantics, and refcounts balance in the single-address, multi-address-finalized, and every OOM partial-state case. No API/ABI change (`ares_private.h` + public headers untouched, signatures match). No dead code left from the old implementation. #1049's core defect (forward address leakage + dropped bridge lines) is fully fixed.

Three minor inline notes (1 Low defensive-hardening, 2 Nits). Non-inline items:

- **[Nit] PR-body memory table vs. disposition comment.** The description's table shows peak RSS 23.7 → 23.7 MB, but the earlier disposition comment says 23.7 → 29.8 MB. They describe different revisions (pre- vs post- the caching commit), but a reader seeing both is left confused — worth reconciling the record.
- **[Info] glibc alias/canonical deviation (retained by design).** Resolved addresses match glibc `multi on`. The alias set and canonical name are *address-scoped* (broader than glibc, which derives them only from lines where the queried name literally appears) — e.g. querying `mullvad-no-svg` reports canonical `mullvad-no`. This is documented in the header block + `GetHostByNameHostsFile` and codified by tests; it's a naming cosmetic, not an address leak, and fixing it would require the per-line provenance the storage model intentionally discards. Noting for the record only.
- **[Info] Overlap with #1251.** #1251 (fixes #976, localhost hosts entries) also edits `ares_getaddrinfo.c` `file_lookup()` and appends to `ares-test-internal.cc` — a textual merge conflict is likely though the logic is independent; whichever lands first, the other needs a trivial rebase.
- **[Info] Release note.** This is a user-visible behavior change (forward/reverse/alias results now address-scoped; `gethostbyaddr` returns only the queried address). Per repo policy `RELEASE-NOTES.md` isn't touched in-PR, but it's worth a line at release time.

src/lib/ares_hosts_file.c
```diff
@@ -226,136 +257,284 @@ static ares_hosts_file_t *ares_hosts_file_create(const char *filename)
...
+   * in order; for each, iterate the reverse entry's hostnames in order,
+   * appending each name not already present (case-insensitive).  Cap at 101
+   * (canonical + 100 aliases) to bound the StevenBlack blocklist case. */
+  for (ipnode = ares_llist_node_first(ent->ips); ipnode != NULL;
+       ipnode = ares_llist_node_next(ipnode)) {
+    const char         *ip  = ares_llist_node_val(ipnode);
+    ares_hosts_entry_t *rev = ares_htable_strvp_get_direct(hosts->iphash, ip);
+    ares_llist_node_t  *nnode;
+
+    if (ares_llist_len(ent->hosts) >= 101) {
```
bradh352: **[Nit]** The "100 aliases" policy is spelled with two different literals in two places — `>= 101` here (canonical + 100) and `cnt > 100` in `ares_hosts_ai_append_cnames` (~line 1000). A shared `#define ARES_HOSTS_MAX_ALIASES 100` would keep them in lockstep and self-document. Minor: both cap-break lines are tagged `LCOV_EXCL_LINE: DefensiveCoding`, but the cap is genuinely reachable with a crafted local file (a multi-address name whose addresses collectively carry >100 names), so `FallbackCode` is the more accurate tag — or add one small test that trips it.

src/lib/ares_hosts_file.c
bradh352: Addressed in 11b1ebf7 — added the `if (rev == NULL) { continue; }` guard before the deref.

src/lib/ares_hosts_file.c
bradh352: Addressed in 11b1ebf7 — renamed to `ares_hosts_strlist_contains`.

src/lib/ares_hosts_file.c
bradh352: Addressed in 11b1ebf7 — extracted `#define ARES_HOSTS_MAX_ALIASES 100` used in both call sites, and retagged the reachable cap breaks as `FallbackCode`.

PR #1251, opened 2026-07-09
getaddrinfo: ignore non-loopback localhost hosts entries

### LINKED ISSUE #976: Don't return non-loopback addresses when resolving `localhost`.

Alb3e3:
## Summary

- Synthesize localhost loopback results before consulting hosts-file entries in the getaddrinfo hosts-file lookup path.
- Add a regression covering a `CARES_HOSTS` file that maps `localhost` to a documentation-only non-loopback IPv4 address.
- Leave `.onion` rejection and ordinary non-localhost hosts-file lookups unchanged.

Fixes #976.

## Test Plan

- Red check before the fix: `./build-make/bin/arestest -4 --gtest_filter=FileChannelTest.GetAddrInfoHostsLocalhostIgnoresNonLoopback` failed with `{addr=[192.0.2.10]}` instead of `{addr=[127.0.0.1]}`.
- `cmake -S . -B build-make -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=DEBUG -DCARES_BUILD_TESTS=ON -DGTEST_ROOT=/tmp/c-ares-gtest-install`
- `cmake --build build-make --target arestest -j2`
- `./build-make/bin/arestest -4 --gtest_filter=FileChannelTest.GetAddrInfoHostsLocalhostIgnoresNonLoopback`
- `./build-make/bin/arestest -4 --gtest_filter='FileChannelTest.GetAddrInfoHosts*'`
- `./build-make/bin/arestest -4 --gtest_filter='*Localhost*:*LocalHost*'`
- `./build-make/bin/arestest -4 --gtest_filter='-*LiveSearchTXT*:*LiveSearchANY*'`

### DIFF
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -401,6 +401,16 @@ static ares_status_t file_lookup(struct host_query *hquery)
     return ARES_ENOTFOUND;
   }
 
+  /* RFC6761 section 6.3 #3 states that "Name resolution APIs and libraries
+   * SHOULD recognize localhost names as special and SHOULD always return the
+   * IP loopback address for address queries".
+   * We will also ignore ALL errors when trying to resolve localhost, such
+   * as permissions errors reading /etc/hosts or a malformed /etc/hosts. */
+  if (ares_is_localhost(hquery->name)) {
+    return ares_addrinfo_localhost(hquery->name, hquery->port, &hquery->hints,
+                                   hquery->ai);
+  }
+
   status = ares_hosts_search_host(
     hquery->channel,
     (hquery->hints.ai_flags & ARES_AI_ENVHOSTS) ? ARES_TRUE : ARES_FALSE,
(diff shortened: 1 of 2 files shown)


Alb3e3:
No rebase needed here: the branch is already directly on current main locally. I reran ./build/bin/arestest -4 --gtest_filter="FileChannelTest.GetAddrInfoHostsLocalhostIgnoresNonLoopback" before flipping this out of draft.

Alb3e3:
Gentle ping on this one. It has been open since 9 July without a review yet. It is on
current main, mergeable, and all 28 checks are green.

Recap of the issue it fixes (#976): the `ares_getaddrinfo()` hosts-file path consults the
hosts file before synthesizing loopback results, so a hosts file that maps `localhost` to a
non-loopback address (for example the documentation range address `192.0.2.10`) makes
`localhost` resolve to that address. The change synthesizes the loopback result first for
`localhost` lookups, and leaves `.onion` rejection and ordinary non-localhost hosts-file
lookups untouched.

The regression test is non-vacuous: without the change it fails with `{addr=[192.0.2.10]}`
instead of `{addr=[127.0.0.1]}`.

No rush if the queue is long. I mainly wanted to check it had not been missed.

PR #1254, opened 2026-07-12
Fix double-free in ares_requeue_queries() on connection close

g0w6y:
## Summary

`ares_close_sockets.c`'s `ares_requeue_queries()` called `ares_requeue_query()` with `requeue=NULL`. On retry exhaustion, this hit `end_query()`'s legacy direct-callback path, invoking the query's callback before detaching it from `channel->all_queries`/`queries_by_qid`. A reentrant `ares_cancel()` from that callback — an ordinary pattern, not misuse — would find the same still-linked query, invoke its callback again, and free it, causing the original `end_query()` call to free it a second time on return.

This is the same root cause as CVE-2026-33630 / GHSA-6wfj-rwm7-3542. That fix (#1237) hardened `process_timeouts()` and `read_answers()` via `ares_flush_requeue()`, but never touched `ares_close_sockets.c`, so the identical bug remained reachable through any connection close/error while a query was on its final retry — no malformed traffic required.

## What changed

`ares_close_connection()` and the internal `ares_requeue_queries()` now take an `ares_array_t **requeue` parameter and thread it down to `ares_requeue_query()`, mirroring the existing pattern used by `read_answers()` / `process_timeouts()` / `ares_send_query()`.

`handle_conn_error()` — the sole caller of `ares_close_connection()` that can reach a connection with live queries attached — now also takes `requeue`. When called with a non-NULL `requeue` (currently only from `read_answers()`, which already owns one), it appends into that scope. When called with `NULL` (all its other 6 call sites, none of which previously had an active requeue scope), it creates a local one and flushes it before returning, so every existing caller keeps working unchanged.

The other two callers of `ares_close_connection()` (`ares_close_sockets()` and `ares_check_cleanup_conns()`) now explicitly pass `NULL`. Both only ever close connections with an empty `queries_to_conn` list in every current call path (`ares_check_cleanup_conns()` skips connections with any attached queries via its own guard; `ares_destroy()` drains all queries before tearing down servers), so this is unchanged behavior made explicit, not a new gap.

## Testing

- Added `MockChannelTest.CancelInCallbackNoDoubleFreeOnConnError`, mirroring the existing `CancelInCallbackNoDoubleFree` regression test for #1237 but reached via `Disconnect()` (connection close) instead of a normal answer, across all 4 address-family/transport variants.
- Verified this test reproduces the double-free on the pre-fix code: `___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED` through `ares_requeue_queries → ares_close_connection → handle_conn_error → end_query → ares_free`, matching the reported stack exactly.
- Full existing suite (1102 tests) plus the 4 new variants (1106 total) pass clean with the fix applied.
- Also independently verified with two standalone reproducers (C+ASan, Python+ctypes) against a live channel: pre-fix, the query callback fires twice for one query and the process aborts on double-free; post-fix, it fires exactly once and exits cleanly.

Reported privately per SECURITY.md; opening this publicly with the security team's go-ahead.

### DIFF
--- a/src/lib/ares_close_sockets.c
+++ b/src/lib/ares_close_sockets.c
@@ -29,7 +29,8 @@
 #include <assert.h>
 
 static void ares_requeue_queries(ares_conn_t  *conn,
-                                 ares_status_t requeue_status)
+                                 ares_status_t requeue_status,
+                                 ares_array_t **requeue)
 {
   ares_query_t  *query;
   ares_timeval_t now;
@@ -37,11 +38,12 @@ static void ares_requeue_queries(ares_conn_t  *conn,
   ares_tvnow(&now);
 
   while ((query = ares_llist_first_val(conn->queries_to_conn)) != NULL) {
-    ares_requeue_query(query, &now, requeue_status, ARES_TRUE, NULL, NULL);
+    ares_requeue_query(query, &now, requeue_status, ARES_TRUE, NULL, requeue);
   }
 }
 
-void ares_close_connection(ares_conn_t *conn, ares_status_t requeue_status)
+void ares_close_connection(ares_conn_t *conn, ares_status_t requeue_status,
+                           ares_array_t **requeue)
 {
   ares_server_t  *server  = conn->server;
   ares_channel_t *channel = server->channel;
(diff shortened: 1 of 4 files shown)


g0w6y:
@bradh352 any updates ?

PR #1256, opened 2026-07-13
Regression: Query removal after random qid reuse

eriklax:
This PR fixes a regression in main and 1.34.8 that causes a serious issue. If a reentrant query is sent on the same channel with the same qid (yes selected by random, yes till will happen at some point, yes it done so), the query is removed without being executed, resulting in a stale execution flow. The behaviour and regression can be seen if

Make a test-program that from its result callback send another query

```
#include <ares.h>
#include <string.h>
#include <stdio.h>

ares_channel_t *channel;
ares_dns_record_t *query;

static void result_cb(void *arg, ares_status_t status, size_t timeouts,
                      const ares_dns_record_t *answer)
{
  printf("woop!\n");
  ares_send_dnsrec(channel, query, result_cb, 0x0, 0x0);
}

int main(int argc, char **argv)
{
  struct ares_options options;
  memset(&options, 0, sizeof(options));
  options.qcache_max_ttl = 0;
  options.evsys = ARES_EVSYS_DEFAULT;

  ares_library_init(ARES_LIB_INIT_ALL);
  ares_dns_record_create(&query, 0, ARES_FLAG_RD,
                         ARES_OPCODE_QUERY, ARES_RCODE_NOERROR);
  ares_dns_record_query_add(query, "example.com",
                            ARES_REC_TYPE_A, ARES_CLASS_IN);
  ares_init_options(&channel, &options, ARES_OPT_QUERY_CACHE | ARES_OPT_EVENT_THREAD);
  ares_send_dnsrec(channel, query, result_cb, 0x0, 0x0);
  while (ares_queue_wait_empty(channel, 1000) == ARES_ETIMEOUT)
    ;
  ares_destroy(channel);
  ares_dns_record_destroy(query);
  ares_library_cleanup();
  return 0;
}
```

Make the ```generate_unique_qid```/```ares_generate_new_id``` always generate the same number (eg id = 42).

```
unsigned short ares_generate_new_id(ares_rand_state *state)
{
  return 42;
}
```

In 1.34.6 generate_unique_qid would fail in this case (since the qid would exist) and it would loop forever, in main and 1.34.7/.8 generate_unique_qid happily returns 42 and ares_detach_query removes the query twice and the new query callback is never executed. With this patch, "woop" is printed again and again as one would expect.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -1598,7 +1598,12 @@ static void ares_detach_query(ares_query_t *query)
 {
   /* Remove the query from all the lists in which it is linked */
   ares_query_remove_from_conn(query);
-  ares_htable_szvp_remove(query->channel->queries_by_qid, query->qid);
+  /* A callback may queue a new query that reuses this ID. Only remove the
+   * entry if it still points to this query. */
+  if (ares_htable_szvp_get_direct(query->channel->queries_by_qid, query->qid) ==
+      query) {
+    ares_htable_szvp_remove(query->channel->queries_by_qid, query->qid);
+  }
   ares_llist_node_destroy(query->node_all_queries);
   query->node_all_queries = NULL;
 }


olegchorny:
We hit this in production and reached the same line independently before finding this PR. Posting a reproduction at **natural qid width** — no forced constant — plus the field data, in the hope it helps this get reviewed.

## Reproduction

Public-API-only probe, no patches to c-ares, natural 16-bit qid space. Alpine 3.21 / musl, `ndots:5` with 4 search domains, `AF_UNSPEC`, answered by an in-process mock server returning NXDOMAIN for the non-terminal search names — 10.0 DNS queries per logical lookup. 8 trials × 20,000 lookups per arm.

An "orphan" is counted only when the callback has not fired **and** `ares_timeout()` returns NULL **and** `ares_fds_nfds() == 0` — i.e. a lookup is outstanding with no socket and no timer, so nothing that exists can ever complete it.

| arm | trials with ≥1 orphan | lookups | hazard |
|---|---|---|---|
| 1.34.6 | **0 / 8** | 160,000 | — |
| 1.34.7 | 6 / 8 | 74,548 | 80.5 ppm |
| 1.34.8 | **8 / 8** | 51,544 | 155 ppm |
| 1.34.7 + the guard in this PR | **0 / 8** | 160,000 | — |

Pooled defective arms: **14 orphans / 126,092 lookups = 111 ppm** (95% Poisson CI 61–186). Pooled clean arms: **0 / 320,000**. At 111 ppm that is P ≈ 4×10⁻¹⁶.

Predicted from first principles: 1/65535 = 15.26 ppm per qid draw inside the window, and an all-NXDOMAIN 5-name walk opens ~8 such draws per lookup → ~122 ppm. Measured 111 ppm.

**The diff in this PR takes it to zero.**

## The window, as we read it

Both removals are by key and the second is unguarded:

- `ares_process.c:658` — `ares_detach_query(query)` added ahead of the callback, freeing qid X
- `:659` — the callback runs and may start a new query, which `generate_unique_qid()` may legitimately mint **at X**, because X is no longer registered
- `:661` — `ares_free_query(query)` → `ares_detach_query()` a second time → `ares_htable_szvp_remove(queries_by_qid, X)` evicts the **new** query

The evicted query is then unreachable: `process_answer` matches strictly by qid and discards its responses (`:926-932`, *"We may have stopped listening for this query, that's ok"*), `ares_query_remove_from_conn()` has already taken its timer and socket, and at retry exhaustion the endqueue drain's `if (query != NULL)` is false, so its callback never runs.

Notably **no application reentrancy is required**. c-ares supplies it itself: with `ndots > 1`, each search-suffix NXDOMAIN chains through `ares_getaddrinfo.c` `next_lookup()` → `ares_query_nolock()`, synchronously inside the callback at `:659`. Any container with a search list gets this on every external lookup.

## Field data

Two unrelated services, separate repositories, both Alpine, both rebuilt onto a base image whose unpinned `apk upgrade` moved **c-ares 1.34.6-r0 → 1.34.8-r0**. Alpine never packaged 1.34.7, and `ares_process.c` is untouched between 1.34.7 and 1.34.8, so the delta lands squarely on this change. Each service began failing within minutes of its own deploy; `libcurl` (8.21.0), `musl` and OpenSSL are byte-identical across the boundary, delivered by the same layer digest.

Live since 2026-07-30 at roughly **1,165 stalled resolutions/day**, ~55–65 ppm per outbound call. Old-image pods co-tenant on the *same* hosts resolving the *same* names show zero failures.

Two negative observations that we think are the signature of an orphaned query rather than a slow one:

- **Zero SERVFAIL** throughout.
- The **bounded** failure mode never moved — `ARES_ETIMEOUT` surfacing as curl error 6 ran 4–39/day before and 14–75/day after, flat — while the unbounded mode went 0 → ~1,165/day. Stalls end only at the *caller's* budget, piling up at exactly libcurl's timeouts including 945 events at a full 300 s (`DEFAULT_CONNECT_TIMEOUT`). A query still on the timeout list would have landed in the bounded population.

Endpoints reached through pooled SDK clients, which perform very few fresh lookups, show zero failures across millions of calls — consistent with a per-resolution hazard.

## Caveats

The A/B above was built on **aarch64**; production is x86_64. The mechanism is architecture-independent and `ares_process.c` is identical, but we have not run the probe against the shipped x86_64 binary. The production attribution is inference from the version lock and the symptom match, not a captured orphan in production.

Happy to share the probe or run further variations if useful.

grigory51:
Please accept this edit, we are suffering: requests to aiohttp are hanging.

ballard26:
To join the chorus here we've also hit the issue described in this PR in our production environment at Redpanda. As a result we've had to revert to c-ares 1.34.6 until a fix lands. 

For us the lost callbacks surfaced as DNS futures that never resolve in our event loop causing permanent hangs while trying to create connections. We also have an independent reproducer that we'd be happy to share if it'd be useful to getting a fix in.

eriklax:
@bradh352 any updates on when the next (patch) release is planned, this regression is causing a lot of issues

MarkMcCaskey:
We're also routinely hitting this in staging. It happens very frequently in pgbouncer and it causes pgbouncer to fail all new connections indefinitely once it's hit.

travisdowns:
This PR looks good overall, but currently the bug it fixes is hiding a UAF just a bit further down that will be exposed if this merges as is (admittedly fairly obscure):

`ares_send_query()` has a second lookup by qid that has the same identity problem, and this PR turns it from "orphaned query" into a heap use-after-free write:

```c
/* src/lib/ares_process.c, ares_send_query() */
if (status == ARES_SUCCESS &&
    ares_htable_szvp_get_direct(channel->queries_by_qid, qid) == NULL) {
  status = ARES_ETIMEOUT;
}
```

That check exists so `ares_send_nolock()` does not write `*qid` after the callback freed the memory it points to (the comment says as much). It is keyed on the id, not the query. If the callback sends a new query that draws the same id and stays outstanding, the lookup finds the *new* query, `ares_send_query()` returns `ARES_SUCCESS` for a query that was already freed, and `ares_send_nolock()` writes `*qid`. `ares_getaddrinfo()` reaches this by itself: `next_dns_lookup()` passes `&hquery->qid_a`, and if the terminal failure is delivered during that send, `end_hquery()` frees `hquery` while `ares_send_nolock()` is still unwinding underneath it.

Today this cannot happen: the double removal evicts the new query, the lookup returns NULL, `ETIMEOUT` is returned and nothing is written (the new query is orphaned instead, which is the bug this PR fixes). With this PR the write happens. Conditions are narrow (a retry deferred and then every connection attempt failing synchronously inside the initial send, a callback that frees the `*qid` owner and synchronously sends a new query, and the 1/65536 id collision), but it is the same order of probability as the bug being fixed here, and it is a write rather than a hang.

Suggested fix, same change: compare identity instead of presence.

```c
if (status == ARES_SUCCESS &&
    ares_htable_szvp_get_direct(channel->queries_by_qid, qid) != query) {
  status = ARES_ETIMEOUT;
}
```

`query` may already be freed there, but only its value is compared, never dereferenced, and the replacement is allocated while the old query is still live so the addresses cannot coincide. Alternatives: have `ares_flush_requeue()` report that the query ended, or assign `*qid` in `ares_send_nolock()` before calling `ares_send_query()`.

Two smaller notes, not blockers for this PR:

- The lookup at the top of the `ares_flush_requeue()` loop is also by qid, so a stale entry can resolve to a reused id if an earlier callback in the same drain called `ares_cancel()`. Pre-existing and much less likely; storing the query pointer in `ares_requeue_t` and validating it would close it.
- #1254, #1260 and #1283 add detach-before-callback to the synchronous `end_query()` path. Once any of those lands, that path detaches twice as well and needs this guard to avoid the same orphaning, so this PR is a prerequisite for them.

## Reproducer

Forces ids to collide so the 1/65536 case happens every time:

```diff
--- a/src/lib/util/ares_rand.c
+++ b/src/lib/util/ares_rand.c
@@ unsigned short ares_generate_new_id(ares_rand_state *state)
 {
+  (void)state; return 42;
```

```sh
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCARES_STATIC=ON -DCARES_SHARED=OFF \
      -DCARES_BUILD_TOOLS=OFF -DCARES_BUILD_TESTS=OFF \
      -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g"
cmake --build build
gcc -g -fsanitize=address -I include -I build -DCARES_STATICLIB -o repro qid_reuse_repro.c build/lib/libcares.a -lpthread
ASAN_OPTIONS=detect_leaks=0 ./repro orphan   # the bug this PR fixes
ASAN_OPTIONS=detect_leaks=0 ./repro uaf      # the write described above
```

No network is used. Socket hooks black-hole sends so queries end via the timeout path, or fail `socket()` on demand.

On current `main` (`5b181482`):

```
$ ./repro orphan
initial send -> Successful completion
cb #1 status=Timeout while contacting DNS servers; re-sent -> Successful completion
callbacks=1 active_queries=1
cb #2 status=Channel is being destroyed; re-sent -> Successful completion
repro: src/lib/ares_destroy.c:91: ares_destroy: Assertion `ares_llist_len(channel->all_queries) == 0' failed.

$ ./repro uaf
getaddrinfo cb status=Could not contact DNS servers
callbacks=1 active_queries=1          <- second query orphaned, no callback
second query ended: Channel is being destroyed
```

On `main` + this PR:

```
$ ./repro orphan
cb #1 ... cb #5 status=Timeout while contacting DNS servers; done
callbacks=5 active_queries=0          <- fixed

$ ./repro uaf
getaddrinfo cb status=Could not contact DNS servers
==ERROR: AddressSanitizer: heap-use-after-free ... WRITE of size 2
    #0 ares_send_nolock src/lib/ares_send.c:222        <- *qid = id
    #1 ares_query_nolock src/lib/ares_query.c:108
    #2 next_dns_lookup src/lib/ares_getaddrinfo.c:784
    #3 next_lookup src/lib/ares_getaddrinfo.c:450
freed by thread T0 here:
    #3 hquery_free src/lib/ares_getaddrinfo.c:336
    #4 end_hquery src/lib/ares_getaddrinfo.c:364
    #5 host_callback src/lib/ares_getaddrinfo.c:578
    #6 ares_query_dnsrec_cb src/lib/ares_query.c:46
    #7 ares_flush_requeue src/lib/ares_process.c:656
    #8 ares_send_query src/lib/ares_process.c:1384
    #9 ares_send_nolock src/lib/ares_send.c:220
```

<details>
<summary>qid_reuse_repro.c</summary>

```c
/* Reproducer: query id reuse after ares_detach_query() (c-ares >= 1.34.7).
 *
 * Build against a c-ares whose ares_generate_new_id() always returns 42, so a
 * query sent from inside a callback reuses the id of the query that is
 * finishing.  With random ids this happens with probability 1/65536 per send.
 *
 *   ./repro orphan   callback re-sends a query; count how many callbacks fire
 *   ./repro uaf      ares_getaddrinfo() whose first send fails synchronously;
 *                    the user callback sends a new query that stays outstanding
 *
 * No network is used: the socket hooks below either black-hole sends (so
 * queries end via the timeout path) or fail socket() on demand.
 */
#include <ares.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/select.h>
#include <sys/socket.h>
#include <sys/time.h>
#include <sys/uio.h>
#include <unistd.h>

static ares_channel_t    *channel;
static ares_dns_record_t *query;
static int                ncb;
static int                fail_sockets;

static ares_socket_t my_socket(int d, int t, int p, void *u)
{
  (void)u;
  if (fail_sockets) {
    errno = EACCES;
    return ARES_SOCKET_BAD;
  }
  return socket(d, t, p);
}
static int my_close(ares_socket_t s, void *u)
{
  (void)u;
  return close(s);
}
static int my_connect(ares_socket_t s, const struct sockaddr *a,
                      ares_socklen_t l, void *u)
{
  (void)u;
  return connect(s, a, l);
}
static ares_ssize_t my_recvfrom(ares_socket_t s, void *b, size_t len, int flags,
                                struct sockaddr *a, ares_socklen_t *al, void *u)
{
  (void)u;
  return recvfrom(s, b, len, flags, a, al);
}
/* Pretend the send succeeded but never put anything on the wire. */
static ares_ssize_t my_sendv(ares_socket_t s, const struct iovec *v, int n,
                             void *u)
{
  ares_ssize_t total = 0;
  int          i;
  (void)s;
  (void)u;
  for (i = 0; i < n; i++) {
    total += (ares_ssize_t)v[i].iov_len;
  }
  return total;
}
static const struct ares_socket_functions funcs = { my_socket, my_close,
                                                    my_connect, my_recvfrom,
                                                    my_sendv };

static void pump(int seconds)
{
  struct timeval deadline, now;
  gettimeofday(&deadline, NULL);
  deadline.tv_sec += seconds;
  for (;;) {
    fd_set         rfds, wfds;
    int            nfds;
    struct timeval tv;
    gettimeofday(&now, NULL);
    if (timercmp(&now, &deadline, >=)) {
      break;
    }
    FD_ZERO(&rfds);
    FD_ZERO(&wfds);
    nfds = ares_fds(channel, &rfds, &wfds);
    if (nfds == 0) {
      break; /* no sockets left: nothing can make progress */
    }
    tv.tv_sec  = 0;
    tv.tv_usec = 100000;
    select(nfds, &rfds, &wfds, NULL, ares_timeout(channel, &tv, &tv));
    ares_process(channel, &rfds, &wfds);
  }
}

static void resend_cb(void *arg, ares_status_t status, size_t timeouts,
                      const ares_dns_record_t *answer)
{
  (void)arg;
  (void)timeouts;
  (void)answer;
  ncb++;
  if (ncb < 5) {
    ares_status_t st = ares_send_dnsrec(channel, query, resend_cb, NULL, NULL);
    printf("cb #%d status=%s; re-sent -> %s\n", ncb, ares_strerror(status),
           ares_strerror(st));
  } else {
    printf("cb #%d status=%s; done\n", ncb, ares_strerror(status));
  }
}

static void noop_cb(void *arg, ares_status_t status, size_t timeouts,
                    const ares_dns_record_t *answer)
{
  (void)arg;
  (void)timeouts;
  (void)answer;
  printf("second query ended: %s\n", ares_strerror(status));
}

static void gai_cb(void *arg, int status, int timeouts,
                   struct ares_addrinfo *res)
{
  (void)arg;
  (void)timeouts;
  ncb++;
  printf("getaddrinfo cb status=%s\n", ares_strerror(status));
  ares_freeaddrinfo(res);
  /* Sockets work from here on, so the query sent below stays outstanding. */
  fail_sockets = 0;
  ares_send_dnsrec(channel, query, noop_cb, NULL, NULL);
  /* On return end_hquery() frees the host_query that owns the qid_a slot the
   * still-unwinding ares_send_nolock() is about to write to. */
}

int main(int argc, char **argv)
{
  struct ares_options options;
  const char         *mode = argc > 1 ? argv[1] : "orphan";

  setvbuf(stdout, NULL, _IONBF, 0);
  memset(&options, 0, sizeof(options));
  options.qcache_max_ttl = 0;
  options.tries          = 2;
  options.timeout        = 150;

  ares_library_init(ARES_LIB_INIT_ALL);
  ares_dns_record_create(&query, 0, ARES_FLAG_RD, ARES_OPCODE_QUERY,
                         ARES_RCODE_NOERROR);
  ares_dns_record_query_add(query, "nonexistent.invalid", ARES_REC_TYPE_A,
                            ARES_CLASS_IN);
  ares_init_options(&channel, &options,
                    ARES_OPT_QUERY_CACHE | ARES_OPT_TRIES | ARES_OPT_TIMEOUTMS);
  ares_set_servers_csv(channel, "127.0.0.1:1");
  ares_set_socket_functions(channel, &funcs, NULL);

  if (strcmp(mode, "uaf") == 0) {
    struct ares_addrinfo_hints hints;
    memset(&hints, 0, sizeof(hints));
    hints.ai_family = AF_INET;
    fail_sockets    = 1;
    ares_getaddrinfo(channel, "nonexistent.invalid", NULL, &hints, gai_cb,
                     NULL);
    pump(3);
  } else {
    ares_status_t st = ares_send_dnsrec(channel, query, resend_cb, NULL, NULL);
    printf("initial send -> %s\n", ares_strerror(st));
    pump(3);
  }
  printf("callbacks=%d active_queries=%zu\n", ncb,
         ares_queue_active_queries(channel));

  ares_destroy(channel);
  ares_dns_record_destroy(query);
  ares_library_cleanup();
  return 0;
}
```

</details>

travisdowns:
I reviewed this carefully since we hit the orphaned-query regression in production too (see @ballard26 above). The guard in `ares_detach_query()` is correct: of the three steps in that function only the hashtable removal is non-idempotent, and the pointer comparison is safe because the old query is not freed until after the callback, so the replacement can never share its address. Full test suite passes under ASan/UBSan with this applied to current `main`.

One thing should go in alongside it though. `ares_send_query()` has a second lookup by qid that has the same identity problem, and this PR turns it from "orphaned query" into a heap use-after-free write:

```c
/* src/lib/ares_process.c, ares_send_query() */
if (status == ARES_SUCCESS &&
    ares_htable_szvp_get_direct(channel->queries_by_qid, qid) == NULL) {
  status = ARES_ETIMEOUT;
}
```

That check exists so `ares_send_nolock()` does not write `*qid` after the callback freed the memory it points to (the comment says as much). It is keyed on the id, not the query. If the callback sends a new query that draws the same id and stays outstanding, the lookup finds the *new* query, `ares_send_query()` returns `ARES_SUCCESS` for a query that was already freed, and `ares_send_nolock()` writes `*qid`. `ares_getaddrinfo()` reaches this by itself: `next_dns_lookup()` passes `&hquery->qid_a`, and if the terminal failure is delivered during that send, `end_hquery()` frees `hquery` while `ares_send_nolock()` is still unwinding underneath it.

Today this cannot happen: the double removal evicts the new query, the lookup returns NULL, `ETIMEOUT` is returned and nothing is written (the new query is orphaned instead, which is the bug this PR fixes). With this PR the write happens. Conditions are narrow (a retry deferred and then every connection attempt failing synchronously inside the initial send, a callback that frees the `*qid` owner and synchronously sends a new query, and the 1/65536 id collision), but it is the same order of probability as the bug being fixed here, and it is a write rather than a hang.

Suggested fix, same change: compare identity instead of presence.

```c
if (status == ARES_SUCCESS &&
    ares_htable_szvp_get_direct(channel->queries_by_qid, qid) != query) {
  status = ARES_ETIMEOUT;
}
```

`query` may already be freed there, but only its value is compared, never dereferenced, and the replacement is allocated while the old query is still live so the addresses cannot coincide. Alternatives: have `ares_flush_requeue()` report that the query ended, or assign `*qid` in `ares_send_nolock()` before calling `ares_send_query()`.

Two smaller notes, not blockers for this PR:

- The lookup at the top of the `ares_flush_requeue()` loop is also by qid, so a stale entry can resolve to a reused id if an earlier callback in the same drain called `ares_cancel()`. Pre-existing and much less likely; storing the query pointer in `ares_requeue_t` and validating it would close it.
- #1254, #1260 and #1283 add detach-before-callback to the synchronous `end_query()` path. Once any of those lands, that path detaches twice as well and needs this guard to avoid the same orphaning, so this PR is a prerequisite for them.

## Reproducer

Forces ids to collide so the 1/65536 case happens every time:

```diff
--- a/src/lib/util/ares_rand.c
+++ b/src/lib/util/ares_rand.c
@@ unsigned short ares_generate_new_id(ares_rand_state *state)
 {
+  (void)state; return 42;
```

```sh
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCARES_STATIC=ON -DCARES_SHARED=OFF \
      -DCARES_BUILD_TOOLS=OFF -DCARES_BUILD_TESTS=OFF \
      -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g"
cmake --build build
gcc -g -fsanitize=address -I include -I build -DCARES_STATICLIB -o repro qid_reuse_repro.c build/lib/libcares.a -lpthread
ASAN_OPTIONS=detect_leaks=0 ./repro orphan   # the bug this PR fixes
ASAN_OPTIONS=detect_leaks=0 ./repro uaf      # the write described above
```

No network is used. Socket hooks black-hole sends so queries end via the timeout path, or fail `socket()` on demand.

On current `main` (`5b181482`):

```
$ ./repro orphan
initial send -> Successful completion
cb #1 status=Timeout while contacting DNS servers; re-sent -> Successful completion
callbacks=1 active_queries=1
cb #2 status=Channel is being destroyed; re-sent -> Successful completion
repro: src/lib/ares_destroy.c:91: ares_destroy: Assertion `ares_llist_len(channel->all_queries) == 0' failed.

$ ./repro uaf
getaddrinfo cb status=Could not contact DNS servers
callbacks=1 active_queries=1          <- second query orphaned, no callback
second query ended: Channel is being destroyed
```

On `main` + this PR:

```
$ ./repro orphan
cb #1 ... cb #5 status=Timeout while contacting DNS servers; done
callbacks=5 active_queries=0          <- fixed

$ ./repro uaf
getaddrinfo cb status=Could not contact DNS servers
==ERROR: AddressSanitizer: heap-use-after-free ... WRITE of size 2
    #0 ares_send_nolock src/lib/ares_send.c:222        <- *qid = id
    #1 ares_query_nolock src/lib/ares_query.c:108
    #2 next_dns_lookup src/lib/ares_getaddrinfo.c:784
    #3 next_lookup src/lib/ares_getaddrinfo.c:450
freed by thread T0 here:
    #3 hquery_free src/lib/ares_getaddrinfo.c:336
    #4 end_hquery src/lib/ares_getaddrinfo.c:364
    #5 host_callback src/lib/ares_getaddrinfo.c:578
    #6 ares_query_dnsrec_cb src/lib/ares_query.c:46
    #7 ares_flush_requeue src/lib/ares_process.c:656
    #8 ares_send_query src/lib/ares_process.c:1384
    #9 ares_send_nolock src/lib/ares_send.c:220
```

<details>
<summary>qid_reuse_repro.c</summary>

```c
/* Reproducer: query id reuse after ares_detach_query() (c-ares >= 1.34.7).
 *
 * Build against a c-ares whose ares_generate_new_id() always returns 42, so a
 * query sent from inside a callback reuses the id of the query that is
 * finishing.  With random ids this happens with probability 1/65536 per send.
 *
 *   ./repro orphan   callback re-sends a query; count how many callbacks fire
 *   ./repro uaf      ares_getaddrinfo() whose first send fails synchronously;
 *                    the user callback sends a new query that stays outstanding
 *
 * No network is used: the socket hooks below either black-hole sends (so
 * queries end via the timeout path) or fail socket() on demand.
 */
#include <ares.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/select.h>
#include <sys/socket.h>
#include <sys/time.h>
#include <sys/uio.h>
#include <unistd.h>

static ares_channel_t    *channel;
static ares_dns_record_t *query;
static int                ncb;
static int                fail_sockets;

static ares_socket_t my_socket(int d, int t, int p, void *u)
{
  (void)u;
  if (fail_sockets) {
    errno = EACCES;
    return ARES_SOCKET_BAD;
  }
  return socket(d, t, p);
}
static int my_close(ares_socket_t s, void *u)
{
  (void)u;
  return close(s);
}
static int my_connect(ares_socket_t s, const struct sockaddr *a,
                      ares_socklen_t l, void *u)
{
  (void)u;
  return connect(s, a, l);
}
static ares_ssize_t my_recvfrom(ares_socket_t s, void *b, size_t len, int flags,
                                struct sockaddr *a, ares_socklen_t *al, void *u)
{
  (void)u;
  return recvfrom(s, b, len, flags, a, al);
}
/* Pretend the send succeeded but never put anything on the wire. */
static ares_ssize_t my_sendv(ares_socket_t s, const struct iovec *v, int n,
                             void *u)
{
  ares_ssize_t total = 0;
  int          i;
  (void)s;
  (void)u;
  for (i = 0; i < n; i++) {
    total += (ares_ssize_t)v[i].iov_len;
  }
  return total;
}
static const struct ares_socket_functions funcs = { my_socket, my_close,
                                                    my_connect, my_recvfrom,
                                                    my_sendv };

static void pump(int seconds)
{
  struct timeval deadline, now;
  gettimeofday(&deadline, NULL);
  deadline.tv_sec += seconds;
  for (;;) {
    fd_set         rfds, wfds;
    int            nfds;
    struct timeval tv;
    gettimeofday(&now, NULL);
    if (timercmp(&now, &deadline, >=)) {
      break;
    }
    FD_ZERO(&rfds);
    FD_ZERO(&wfds);
    nfds = ares_fds(channel, &rfds, &wfds);
    if (nfds == 0) {
      break; /* no sockets left: nothing can make progress */
    }
    tv.tv_sec  = 0;
    tv.tv_usec = 100000;
    select(nfds, &rfds, &wfds, NULL, ares_timeout(channel, &tv, &tv));
    ares_process(channel, &rfds, &wfds);
  }
}

static void resend_cb(void *arg, ares_status_t status, size_t timeouts,
                      const ares_dns_record_t *answer)
{
  (void)arg;
  (void)timeouts;
  (void)answer;
  ncb++;
  if (ncb < 5) {
    ares_status_t st = ares_send_dnsrec(channel, query, resend_cb, NULL, NULL);
    printf("cb #%d status=%s; re-sent -> %s\n", ncb, ares_strerror(status),
           ares_strerror(st));
  } else {
    printf("cb #%d status=%s; done\n", ncb, ares_strerror(status));
  }
}

static void noop_cb(void *arg, ares_status_t status, size_t timeouts,
                    const ares_dns_record_t *answer)
{
  (void)arg;
  (void)timeouts;
  (void)answer;
  printf("second query ended: %s\n", ares_strerror(status));
}

static void gai_cb(void *arg, int status, int timeouts,
                   struct ares_addrinfo *res)
{
  (void)arg;
  (void)timeouts;
  ncb++;
  printf("getaddrinfo cb status=%s\n", ares_strerror(status));
  ares_freeaddrinfo(res);
  /* Sockets work from here on, so the query sent below stays outstanding. */
  fail_sockets = 0;
  ares_send_dnsrec(channel, query, noop_cb, NULL, NULL);
  /* On return end_hquery() frees the host_query that owns the qid_a slot the
   * still-unwinding ares_send_nolock() is about to write to. */
}

int main(int argc, char **argv)
{
  struct ares_options options;
  const char         *mode = argc > 1 ? argv[1] : "orphan";

  setvbuf(stdout, NULL, _IONBF, 0);
  memset(&options, 0, sizeof(options));
  options.qcache_max_ttl = 0;
  options.tries          = 2;
  options.timeout        = 150;

  ares_library_init(ARES_LIB_INIT_ALL);
  ares_dns_record_create(&query, 0, ARES_FLAG_RD, ARES_OPCODE_QUERY,
                         ARES_RCODE_NOERROR);
  ares_dns_record_query_add(query, "nonexistent.invalid", ARES_REC_TYPE_A,
                            ARES_CLASS_IN);
  ares_init_options(&channel, &options,
                    ARES_OPT_QUERY_CACHE | ARES_OPT_TRIES | ARES_OPT_TIMEOUTMS);
  ares_set_servers_csv(channel, "127.0.0.1:1");
  ares_set_socket_functions(channel, &funcs, NULL);

  if (strcmp(mode, "uaf") == 0) {
    struct ares_addrinfo_hints hints;
    memset(&hints, 0, sizeof(hints));
    hints.ai_family = AF_INET;
    fail_sockets    = 1;
    ares_getaddrinfo(channel, "nonexistent.invalid", NULL, &hints, gai_cb,
                     NULL);
    pump(3);
  } else {
    ares_status_t st = ares_send_dnsrec(channel, query, resend_cb, NULL, NULL);
    printf("initial send -> %s\n", ares_strerror(st));
    pump(3);
  }
  printf("callbacks=%d active_queries=%zu\n", ncb,
         ares_queue_active_queries(channel));

  ares_destroy(channel);
  ares_dns_record_destroy(query);
  ares_library_cleanup();
  return 0;
}
```

</details>

PR #1257, opened 2026-07-16
reject rdata larger than the 16-bit rdlength field on write

jmestwa-coder:
ares_dns_write_rr() narrows the assembled rdata length into the 16-bit RDLENGTH field without checking it fits:
- rdata is appended to the buffer first, then RDLENGTH is written as (unsigned short)(rdlength & 0xFFFF)
- the record setters accept lengths well past 65535 (#1168 only guards the SIZE_MAX allocation), so a record can hold oversized rdata
- ares_dns_write() then returns success but emits a truncated RDLENGTH that mis-frames every following record

Reject the record with ARES_EBADQUERY instead, matching the oversized-message check already in ares_dns_write_buf_tcp. Added DNSRecordRejectsOversizedRdata.


jmestwa-coder:
any update?

PR #1258, opened 2026-07-17
size name buffers for the escaped presentation form

jmestwa-coder:
A DNS name is at most 255 octets on the wire, but its presentation form escapes each non-printable/reserved octet as `\DDD` (4 chars), so a valid name's escaped string can reach ~4x that. Three sites sized against the 255-octet assumption reject such a name:
- `ares_dns_name_write`: `name_copy[512]` was silently truncated by `ares_strcpy`, so the name split/validated below was a different (or malformed) one
- `ares_nameoffset_create`: rejected an escaped name > 255 with `EFORMERR` when recording it as a compression target, which is what actually fails a full `ares_dns_write`
- `ares_apply_dns0x20`: `dns0x20name[256]` rejected the query outright

Added `ARES_MAX_NAME_ESCAPED_LEN` (4*255), sized the buffers/limit and the 0x20 randdata to it, and reject an over-long name instead of truncating.

Repro: a 3x63 octet name of `0xFF` (192 wire octets, 758-char presentation) returned `ARES_EBADNAME` from `ares_dns_write` before; the added `DNSNameWriteEscapedPresentation` test round-trips it after.

### DIFF
--- a/src/lib/ares_send.c
+++ b/src/lib/ares_send.c
@@ -49,8 +49,11 @@ static ares_status_t ares_apply_dns0x20(ares_channel_t    *channel,
 {
   ares_status_t status = ARES_SUCCESS;
   const char   *name   = NULL;
-  char          dns0x20name[256];
-  unsigned char randdata[256 / 8];
+  /* A name is at most 255 octets on the wire, but its presentation form
+   * escapes each non-printable/reserved octet as "\DDD" (4 chars), so a valid
+   * name can require up to ~4 bytes per octet here. */
+  char          dns0x20name[1024];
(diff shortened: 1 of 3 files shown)


jmestwa-coder:
rebased onto main to re-run CI; the NetBSD failure was the pkgsrc CDN returning Service Unavailable during pkg_add, not related to this change

jmestwa-coder:
rebased again to re-run CI; this NetBSD failure was the runner VM timing out on boot (kex_exchange_identification reset for 4+ hours), also unrelated to the change

PR #1259, opened 2026-07-18
ares_destroy: support being called from within a query callback

bradh352:
## Summary

Rescoped from a documentation-only change to actually **supporting** `ares_destroy()` being called from within a query callback.

Previously this was undefined behavior:
- In the non-event-thread case, frames above the callback on the stack still reference the channel and its queries, so tearing the channel down immediately is a use-after-free.
- In the event-thread case, `ares_destroy()` tries to join the very thread the callback is running on — a self-join **deadlock**.

## Approach

Detect the reentrant call by tracking user-callback recursion depth on the channel (`callback_depth`), maintained under the channel lock around every query-callback dispatch. When `ares_destroy()` is invoked while a callback is on the stack, teardown can't run synchronously, so it is **deferred** (`destroy_pending`) and completed once the callback stack unwinds, at the outermost callback-dispatch frame:

- **Non-event-thread:** `ares_process()` / `ares_process_fds()` / `ares_cancel()` / `ares_process_pending_write()` complete the deferred destroy after processing returns (normal join-based teardown).
- **Event thread:** the run loop observes `destroy_pending` after processing, breaks out, and tears the channel down on its own thread — **detaching** rather than joining itself to avoid the self-join deadlock (new `ares_thread_detach()` primitive).

A claim flag guards against two completers racing.

Because a channel whose callback is still on the stack cannot be freed synchronously, reentrant destroy is **asynchronous**: `ares_destroy()` returns while the callback finishes, and the actual teardown completes afterward. The channel must not be used again after such a call. This is documented in the `ares_destroy(3)` manpage as supported **as of 1.35.0**; prior versions explicitly did not support it.

## Tests

Adds regression tests exercising destroy-from-callback in both modes:
- `MockUDPChannelTest.DestroyInCallback` (non-event-thread)
- `MockUDPEventThreadTest.DestroyInCallback` (event thread, all evsys backends)

Both pass cleanly under ASAN/UBSAN. The event-thread test would deadlock/time out if the self-join bug were reintroduced.

### DIFF
--- a/docs/ares_destroy.3
+++ b/docs/ares_destroy.3
@@ -21,6 +21,14 @@ channel, passing a status of \fIARES_EDESTRUCTION\fP. These calls give the
 callbacks a chance to clean up any state which might have been stored in their
 arguments. A callback must not add new requests to a channel being destroyed.
 
+As of c-ares 1.35.0, \fBares_destroy(3)\fP may be called from within a query
+callback.  In that case the channel cannot be torn down immediately, as the
+callback stack still references it, so the teardown is deferred until the
+callback returns and completes asynchronously.  The channel must not be used
+again after such a call.  Prior versions explicitly did not support this and
+doing so resulted in undefined behavior (including a deadlock when using an
+event thread).
+
 There is no ability to make this function thread-safe.  No additional calls
 using this channel may be made once this function is called.
 .SH SEE ALSO
--- a/src/lib/ares_destroy.c
+++ b/src/lib/ares_destroy.c
@@ -31,11 +31,44 @@
 
 void ares_destroy(ares_channel_t *channel)
 {
+  ares_bool_t deferred = ARES_FALSE;
+
+  if (channel == NULL) {
+    return;
+  }
+
+  /* If ares_destroy() is called from within a query callback, we cannot tear
+   * the channel down right now: frames above us on the stack still reference
+   * the channel and its queries, and with an event thread we would be trying
+   * to join our own thread.  Defer the teardown; the outermost callback-
+   * dispatch frame (ares_process_fds(), ares_cancel(), or the event thread
+   * run loop) will complete it once the callback stack unwinds. */
+  ares_channel_lock(channel);
+  if (channel->callback_depth > 0) {
+    channel->destroy_pending = ARES_TRUE;
+    deferred                 = ARES_TRUE;
+  }
+  ares_channel_unlock(channel);
+
+  if (deferred) {
+    /* Nudge the event thread so it observes the pending destroy promptly even
+     * if the deferral originated on another thread. */
+    if (channel->optmask & ARES_OPT_EVENT_THREAD) {
+      ares_event_thread_wake_channel(channel);
+    }
+    return;
+  }
+
+  ares_destroy_int(channel, ARES_FALSE);
+}
+
+void ares_destroy_int(ares_channel_t *channel, ares_bool_t from_ethread)
+{
   size_t             i;
   ares_llist_node_t *node = NULL;
 
   if (channel == NULL) {
-    return;
+    return; /* LCOV_EXCL_LINE: DefensiveCoding */
   }
 
   /* Mark as being shutdown */
@@ -76,7 +109,7 @@ void ares_destroy(ares_channel_t *channel)
     ares_query_t      *query = ares_llist_node_claim(node);
 
     query->node_all_queries = NULL;
-    query->callback(query->arg, ARES_EDESTRUCTION, 0, NULL);
+    ares_invoke_query_callback(query, ARES_EDESTRUCTION, 0, NULL);
     ares_free_query(query);
 
     node = next;
@@ -104,7 +137,7 @@ void ares_destroy(ares_channel_t *channel)
 
   /* Shut down the event thread */
   if (channel->optmask & ARES_OPT_EVENT_THREAD) {
-    ares_event_thread_destroy(channel);
+    ares_event_thread_destroy(channel, from_ethread);
   }
 
   if (channel->domains) {
--- a/src/lib/ares_getaddrinfo.c
+++ b/src/lib/ares_getaddrinfo.c
@@ -762,8 +762,13 @@ void ares_getaddrinfo(ares_channel_t *channel, const char *name,
     return;
   }
   ares_channel_lock(channel);
+  /* Callbacks below may run synchronously; defer a reentrant destroy. */
+  ares_channel_callback_enter(channel);
   ares_getaddrinfo_int(channel, name, service, hints, callback, arg);
+  ares_channel_callback_leave(channel);
   ares_channel_unlock(channel);
+  /* Complete a destroy deferred from one of those callbacks. */
+  ares_destroy_if_deferred(channel);
 }
 
 static ares_bool_t next_dns_lookup(struct host_query *hquery)
--- a/src/lib/ares_private.h
+++ b/src/lib/ares_private.h
@@ -307,6 +307,18 @@ struct ares_channeldata {
    * system config changes might get triggered and we need a flag to make
    * sure we don't take action. */
   ares_bool_t                         sys_up;
+
+  /* Depth of user query callbacks currently executing on the stack.  Modified
+   * only while holding the channel lock.  Used to detect reentrant API calls
+   * (e.g. ares_destroy()) issued from within a callback. */
+  size_t                              callback_depth;
+
+  /* TRUE if ares_destroy() was invoked from within a callback and therefore
+   * had to be deferred.  The outermost callback-dispatch frame (ares_process_
+   * fds(), ares_cancel(), or the event thread loop) completes the teardown
+   * once callback_depth returns to zero.  Modified only while holding the
+   * channel lock. */
+  ares_bool_t                         destroy_pending;
 };
 
 /* Does the domain end in ".onion" or ".onion."? Case-insensitive. */
@@ -351,6 +363,24 @@ void ares_dnsrec_convert_cb(void *arg, ares_status_t status, size_t timeouts,
                             const ares_dns_record_t *dnsrec);
 
 void ares_free_query(ares_query_t *query);
+
+/*! Invoke a query's user callback, tracking callback recursion depth on the
+ *  channel so reentrant API calls (notably ares_destroy()) can be detected.
+ *  Must be called while holding the channel lock. */
+void ares_invoke_query_callback(ares_query_t *query, ares_status_t status,
+                                size_t                   timeouts,
+                                const ares_dns_record_t *dnsrec);
+
+/*! Perform the actual channel teardown.  from_ethread must be ARES_TRUE only
+ *  when the event thread is completing a deferred destroy on its own thread,
+ *  so the event thread is detached rather than join()'d. */
+void ares_destroy_int(ares_channel_t *channel, ares_bool_t from_ethread);
+
+/*! If an ares_destroy() was deferred (invoked from within a callback) and the
+ *  callback stack has now fully unwound, complete the teardown.  Must be called
+ *  with no locks held from a non-event-thread callback-dispatch frame.  Returns
+ *  ARES_TRUE if the channel was destroyed (and is now invalid). */
+ares_bool_t ares_destroy_if_deferred(ares_channel_t *channel);
 
 unsigned short ares_generate_new_id(ares_rand_state *state);
 ares_status_t ares_expand_name_validated(const unsigned char *encoded,
@@ -638,8 +668,17 @@ void ares_channel_unlock(const ares_channel_t *channel);
 struct ares_event_thread;
 typedef struct ares_event_thread ares_event_thread_t;
 
-void ares_event_thread_destroy(ares_channel_t *channel);
+/*! Destroy the event thread associated with the channel.  from_self must be
+ *  set to ARES_TRUE only when invoked from within the event thread itself
+ *  (i.e. a deferred ares_destroy() being completed by the event thread), in
+ *  which case the thread is detached rather than join()'d to avoid a
+ *  self-join deadlock. */
+void ares_event_thread_destroy(ares_channel_t *channel, ares_bool_t from_self);
 ares_status_t ares_event_thread_init(ares_channel_t *channel);
+
+/*! Wake the channel's event thread, if any, so it promptly re-evaluates its
+ *  state (used to make it observe a deferred destroy request). */
+void ares_event_thread_wake_channel(ares_channel_t *channel);
 
 
 #ifdef _WIN32
--- a/src/lib/ares_private.h
+++ b/src/lib/ares_private.h
@@ -367,6 +367,13 @@ void ares_free_query(ares_query_t *query);
 /*! Invoke a query's user callback, tracking callback recursion depth on the
  *  channel so reentrant API calls (notably ares_destroy()) can be detected.
  *  Must be called while holding the channel lock. */
+/*! Mark the start and end of a region in which the library may synchronously
+ *  invoke an application callback while frames above still need the channel.
+ *  A reentrant ares_destroy() inside such a region is deferred rather than
+ *  tearing the channel down underneath those frames.  Nesting is fine. */
+void ares_channel_callback_enter(ares_channel_t *channel);
+void ares_channel_callback_leave(ares_channel_t *channel);
+
 void ares_invoke_query_callback(ares_query_t *query, ares_status_t status,
                                 size_t                   timeouts,
                                 const ares_dns_record_t *dnsrec);
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -262,6 +262,14 @@ ares_status_t ares_process_fds(ares_channel_t         *channel,
   ares_channel_lock(channel);
   status = ares_process_fds_nolock(channel, events, nevents, flags);
   ares_channel_unlock(channel);
+
+  /* If a callback invoked during processing called ares_destroy(), it was
+   * deferred; complete it now that the callback stack has unwound.  The event
+   * thread completes its own deferred destroy in its run loop (it must detach
+   * rather than join itself), so skip that here. */
+  if (!(channel->optmask & ARES_OPT_EVENT_THREAD)) {
+    ares_destroy_if_deferred(channel);
+  }
   return status;
 }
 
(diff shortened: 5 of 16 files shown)


src/lib/ares_getaddrinfo.c
```diff
@@ -762,8 +762,13 @@ void ares_getaddrinfo(ares_channel_t *channel, const char *name,
     return;
   }
   ares_channel_lock(channel);
+  /* Callbacks below may run synchronously; defer a reentrant destroy. */
+  ares_channel_callback_enter(channel);
   ares_getaddrinfo_int(channel, name, service, hints, callback, arg);
+  ares_channel_callback_leave(channel);
   ares_channel_unlock(channel);
+  /* Complete a destroy deferred from one of those callbacks. */
+  ares_destroy_if_deferred(channel);
```
bradh352, line 771: **Major** -- with an event thread, this races the event thread and can touch a freed channel.

An `ares_destroy()` deferred from a callback dispatched inside this locked region also wakes the event thread (`src/lib/ares_destroy.c:56`). Once `ares_channel_unlock()` returns, the event thread -- which may already be blocked in `ares_channel_lock()` inside `ares_process_fds()` -- takes the lock, claims `destroy_pending`, and runs `ares_destroy_int()`, which frees the channel. This frame then calls `ares_destroy_if_deferred(channel)` on freed memory.

Confirmed under AddressSanitizer with an `ARES_OPT_EVENT_THREAD` channel and `ares_getaddrinfo(channel, "127.0.0.1", ...)` -- a literal address is answered from inside the call, so the callback runs on the caller's thread -- destroying the channel from that callback:

```
ERROR: AddressSanitizer: heap-use-after-free
READ of size 8 ... thread T0
    #0 ares_channel_lock         util/ares_threads.c:802
    #1 ares_destroy_if_deferred  ares_process.c:1662
    #2 ares_getaddrinfo          ares_getaddrinfo.c:772
freed by thread T1 here:
    #3 ares_destroy_int          ares_destroy.c:167
    #4 ares_event_thread         ares_event_thread.c:399
```

I widened the window with a 100 ms sleep between the unlock and the call so it reproduces every run; the window exists without it, it is just narrow.

The same shape applies to all seven entry points the second commit added -- `ares_getaddrinfo()`, `ares_gethostbyaddr()`, `ares_getnameinfo()`, `ares_query_dnsrec()`, `ares_search()`, `ares_search_dnsrec()`, `ares_send_dnsrec()`. None carries the `!(channel->optmask & ARES_OPT_EVENT_THREAD)` guard that `ares_process_fds()`, `ares_process()`, `ares_process_pending_write()` and `ares_cancel()` use.

Those four are not safe either, only narrower: they read `channel->optmask` *after* unlocking, which is the same read of freed memory. `ares_cancel()` on an event-thread channel is an ordinary thing to do, and its `ARES_ECANCELLED` callback can destroy the channel.

The fix is to decide and claim while the lock is still held. A helper avoids repeating it at eleven sites:

```c
/*! Release the channel lock, completing a destroy deferred from within this
 *  frame's callbacks if this frame owns the completion.  With an event thread
 *  the event thread completes it instead (it must detach rather than join
 *  itself), and ares_destroy() has already woken it. */
void ares_channel_unlock_destroy_if_deferred(ares_channel_t *channel)
{
  ares_bool_t do_destroy =
    (ares_bool_t)(!(channel->optmask & ARES_OPT_EVENT_THREAD) &&
                  channel->destroy_pending && channel->callback_depth == 0);

  if (do_destroy) {
    channel->destroy_pending = ARES_FALSE;
  }
  ares_channel_unlock(channel);

  if (do_destroy) {
    ares_destroy_int(channel, ARES_FALSE);
  }
}
```

It subsumes `ares_destroy_if_deferred()` at every call site. Then here:

```suggestion
  /* Releases the lock, and completes a destroy deferred from one of those
   * callbacks. */
  ares_channel_unlock_destroy_if_deferred(channel);
```

With that change the reproducer above stayed clean across repeated runs with the same widened window.

bradh352 (commented):
Reviewed at `c1c5de8` against merge base `589b5887`. Built and tested in a throwaway worktree.

The deferral mechanism itself is sound, and `callback_depth` is a good way to detect the reentrant case. Two problems block it as it stands, both verified with runnable reproducers rather than by reading:

1. On an event-thread channel, the frame that defers and the event thread both race to complete the teardown, and the loser touches a freed channel (AddressSanitizer trace in the inline comment on `src/lib/ares_getaddrinfo.c`).
2. The server-list update path dispatches query callbacks synchronously and has no completion site, so an `ares_destroy()` from one of those callbacks is silently dropped and the whole channel leaks (`src/lib/ares_process.c` comment).

Beyond that, the man page now promises a supported operation whose completion the application has no way to wait for -- worth nailing down in the same PR, since it cannot be fixed later without new API.

**What I built and ran** (macOS, arm64, clang):

- `DEBUG` + tests: clean; `*Cancel*:*Destroy*:*EventThread*` -- 285/285 pass, including the five new tests.
- `-DCARES_WERROR=ON`: clean (the only warning is the pre-existing `tempnam` deprecation in `test/ares-test.cc:1327`).
- `-DCARES_THREADS=OFF`: clean.
- Library under AddressSanitizer/UndefinedBehaviorSanitizer for the reproducers below.
- `git clang-format --diff` against the merge base: `.c`/`.h` are clean. `test/ares-test-mock-et.cc` and `test/ares-test-mock.cc` differ, but the `clang-format` CI job filters to `'*.c' '*.h'`, so it will not flag them.
- C89 declaration order checked by reading; not built with OpenWatcom or MSVC, and the Win32 `ares_thread_detach()` path was not executed.

**Overlap with open work** (stated as fact, not as a merge suggestion):

- `#1260`, "detach query before synchronous callback in end_query", changes `end_query()` in `src/lib/ares_process.c` -- the same function this PR changes to route the callback through `ares_invoke_query_callback()`.
- `#1154`, "ares_cancel: Avoid Allocation", changes `src/lib/ares_cancel.c`, `src/lib/ares_destroy.c` and `src/lib/ares_private.h` -- the same three files.

src/lib/ares_process.c
```diff
@@ -1603,6 +1624,58 @@ static void ares_detach_query(ares_query_t *query)
...
+  /* Track that we are executing a user callback so that a reentrant
+   * ares_destroy() from within it can be detected and deferred rather than
+   * tearing the channel down underneath the frames still using it (or, with an
+   * event thread, deadlocking on a self-join). */
+  ares_channel_callback_enter(channel);
+  query->callback(query->arg, status, timeouts, dnsrec);
+  ares_channel_callback_leave(channel);
+}
+
+ares_bool_t ares_destroy_if_deferred(ares_channel_t *channel)
```
bradh352, line 1658: **Major** -- server-list changes dispatch query callbacks synchronously and have no completion site, so `ares_destroy()` from those callbacks is silently dropped and the channel leaks.

`ares_servers_update()` removes servers under the channel lock; `ares_destroy_server()` -> `ares_close_sockets()` -> `ares_requeue_queries()` -> `ares_requeue_query()` ends any query that has run out of servers to try, and `end_query()` now runs the callback through `ares_invoke_query_callback()`. So `callback_depth` is non-zero and `ares_destroy()` correctly defers -- but the entry points that got there never call anything that completes the deferral:

- `ares_set_servers()`, `ares_set_servers_ports()` -- `src/lib/ares_update_servers.c:1255`, `:1279`
- `ares_set_servers_csv()`, `ares_set_servers_ports_csv()` -- `src/lib/ares_update_servers.c:1301`, `:1312`
- `ares_reinit()` -> `ares_init_by_sysconfig()` -> `ares_sysconfig_apply()` -- `src/lib/ares_sysconfig.c:690`

Without an event thread nothing else ever looks at `destroy_pending`, so the channel is never torn down. Reproducer -- no event thread, one pending query, then blank the server list:

```c
ares_set_servers_csv(ch, "192.0.2.1");
ares_query_dnsrec(ch, "example.com", ARES_CLASS_IN, ARES_REC_TYPE_A,
                  qcb, NULL, NULL);
ares_set_servers_csv(ch, "");   /* ends the query from inside this call */
```

with `qcb` calling `ares_destroy(ch)`:

```
callback status=12, calling ares_destroy()
ares_destroy() returned
after set_servers_csv: callback fired=1
active queries=0        <- channel still alive; the whole channel leaked
```

`ares_queue_active_queries()` on the supposedly destroyed channel returns cleanly under AddressSanitizer, which shows the channel is still allocated.

With an event thread it happens to work, because `ares_destroy()` wakes the event thread and its loop picks the deferral up -- but that is incidental rather than by design.

Since `docs/ares_destroy.3` now states this is supported, these entry points need the same `ares_channel_callback_enter()`/`leave()` bracket and completion as the seven added in the second commit.

docs/ares_destroy.3
```diff
@@ -21,6 +21,14 @@ channel, passing a status of \fIARES_EDESTRUCTION\fP. These calls give the
...
 arguments. A callback must not add new requests to a channel being destroyed.
 
+As of c-ares 1.35.0, \fBares_destroy(3)\fP may be called from within a query
+callback.  In that case the channel cannot be torn down immediately, as the
+callback stack still references it, so the teardown is deferred until the
+callback returns and completes asynchronously.  The channel must not be used
+again after such a call.  Prior versions explicitly did not support this and
+doing so resulted in undefined behavior (including a deadlock when using an
+event thread).
+
```
bradh352, line 31: **Medium** -- "the teardown is deferred until the callback returns and completes asynchronously" needs to say what the application may do next, because "asynchronously" here means "on a thread you cannot wait for".

When the destroy is deferred on an event-thread channel, the event thread detaches itself via `ares_thread_detach()` and finishes the teardown after `ares_destroy()` has already returned. There is no `ares_queue_wait_empty()` equivalent for it, and `ares_destroy()` returns `void`, so the caller has no way to learn when the channel is actually gone. That leaves these racing with library code still executing:

- `ares_library_cleanup()`
- unloading the library -- `dlclose()` / `FreeLibrary()`
- returning from `main()` / `exit()`

The new event-thread test papers over exactly this with `std::this_thread::sleep_for(250ms)` at `test/ares-test-mock-et.cc:1831`.

Please state it: the teardown may complete on another thread after `ares_destroy()` returns, there is no way to wait for it, and the application must not call `ares_library_cleanup()`, unload the library, or exit until it is certain the callback that issued the destroy has returned and the event thread has finished.

If a way to wait is wanted instead, that is a bigger design question worth settling before the behavior ships -- it cannot be added afterwards without another public function.

test/ares-test-mock-et.cc
```diff
@@ -1752,6 +1753,88 @@ static std::string PrintEvsysFamily(const testing::TestParamInfo<std::tuple<ares
...
+                     std::chrono::high_resolution_clock::now() - tv_begin)
+                     .count();
+    ASSERT_LT(elapsed, 10) << "callback never fired (possible self-join "
+                              "deadlock in ares_destroy())";
+  }
+
+  // The event thread completes teardown asynchronously after the callback
+  // returns; give it a moment to detach and free the channel.  Under ASAN /
+  // valgrind this validates there is no leak or use-after-free.
+  std::this_thread::sleep_for(std::chrono::milliseconds(250));
```
bradh352, line 1831: **Medium** -- a fixed 250 ms sleep is the only thing separating the test from the detached event thread's teardown. Under valgrind, and on a loaded CI runner, 250 ms will not always be enough -- and when it is not, the failure is a use-after-free inside the detached thread after this test has moved on, so it surfaces as a confusing crash in an unrelated later test rather than as this test failing.

There is no way to wait properly today (see the `docs/ares_destroy.3` comment), which is itself the argument for giving the application one.

Separately, several of the new completion sites have no test:

- `ares_cancel()` -- `src/lib/ares_cancel.c:91` -- destroy from an `ARES_ECANCELLED` callback
- `ares_process_pending_write()` -- `src/lib/ares_process.c:466`
- destroy from an `ARES_EDESTRUCTION` callback, i.e. `ares_destroy()` re-entered from inside `ares_destroy_int()`'s own callback loop
- destroy from a synchronous callback on an **event-thread** channel

`MockUDPChannelTest.DestroyInSynchronousCallback` covers the non-event-thread half of that last one. The event-thread half is the case that is broken, and adding it would have caught it.

src/lib/ares_private.h
```diff
@@ -352,6 +364,31 @@ void ares_dnsrec_convert_cb(void *arg, ares_status_t status, size_t timeouts,
...
+/*! Perform the actual channel teardown.  from_ethread must be ARES_TRUE only
+ *  when the event thread is completing a deferred destroy on its own thread,
+ *  so the event thread is detached rather than join()'d. */
+void ares_destroy_int(ares_channel_t *channel, ares_bool_t from_ethread);
+
+/*! If an ares_destroy() was deferred (invoked from within a callback) and the
+ *  callback stack has now fully unwound, complete the teardown.  Must be called
+ *  with no locks held from a non-event-thread callback-dispatch frame.  Returns
+ *  ARES_TRUE if the channel was destroyed (and is now invalid). */
+ares_bool_t ares_destroy_if_deferred(ares_channel_t *channel);
```
bradh352, line 390: **Low** -- two things about `ares_destroy_if_deferred()`.

The comment says it "Must be called with no locks held from a non-event-thread callback-dispatch frame", but seven of its ten call sites call it unconditionally on channels that may have an event thread -- see the comment on `src/lib/ares_getaddrinfo.c`. However that is resolved, the comment and the code should agree.

"Returns `ARES_TRUE` if the channel was destroyed" -- no caller uses the return value. Either make it `void` or drop the sentence.

src/lib/util/ares_threads.h
```diff
@@ -59,4 +59,10 @@ ares_status_t ares_thread_create(ares_thread_t    **thread,
                                  ares_thread_func_t func, void *arg);
 ares_status_t ares_thread_join(ares_thread_t *thread, void **rv);
 
+/*! Detach a thread so its resources are released automatically on exit without
+ *  another thread join()'ing it.  Used when a thread must tear itself down and
+ *  therefore cannot be joined (e.g. the event thread destroying its own
+ *  channel from within a callback).  After this call the handle is invalid. */
```
bradh352, line 65: **Low** -- the doc omits the constraint the Win32 implementation depends on. There, `ares_thread_detach()` does not free the `ares_thread_t`: it sets `thread->detached` and leaves the free to `ares_thread_func()`, which runs on the detached thread itself (`src/lib/util/ares_threads.c:367-370`). That is only correct when a thread detaches itself. A call from another thread would race that flag against the thread's exit and can leak the struct.

The `.c` says "Only ever called by the thread on itself" -- put it in the header too, since that is what a future caller reads.

```suggestion
/*! Detach a thread so its resources are released automatically on exit without
 *  another thread join()'ing it.  Used when a thread must tear itself down and
 *  therefore cannot be joined (e.g. the event thread destroying its own
 *  channel from within a callback).  After this call the handle is invalid.
 *
 *  MUST only ever be called by the thread on itself.  The Win32 implementation
 *  relies on this: it leaves freeing the handle to the thread wrapper, which
 *  still writes the return value after the thread function returns. */
```

src/lib/ares_private.h
```diff
@@ -352,6 +364,31 @@ void ares_dnsrec_convert_cb(void *arg, ares_status_t status, size_t timeouts,
...
+/*! Mark the start and end of a region in which the library may synchronously
+ *  invoke an application callback while frames above still need the channel.
+ *  A reentrant ares_destroy() inside such a region is deferred rather than
+ *  tearing the channel down underneath those frames.  Nesting is fine. */
+void ares_channel_callback_enter(ares_channel_t *channel);
+void ares_channel_callback_leave(ares_channel_t *channel);
+
+void ares_invoke_query_callback(ares_query_t *query, ares_status_t status,
+                                size_t                   timeouts,
+                                const ares_dns_record_t *dnsrec);
```
bradh352, line 379: **Nit** -- two `/*!` blocks are stacked here. The first documents `ares_invoke_query_callback()` but sits above `ares_channel_callback_enter()`/`leave()`, and `ares_invoke_query_callback()` below ends up with no comment at all. AGENTS.md asks for a doxygen comment on every function in a header.

```suggestion
/*! Mark the start and end of a region in which the library may synchronously
 *  invoke an application callback while frames above still need the channel.
 *  A reentrant ares_destroy() inside such a region is deferred rather than
 *  tearing the channel down underneath those frames.  Nesting is fine.
 *  Must be called while holding the channel lock. */
void ares_channel_callback_enter(ares_channel_t *channel);
void ares_channel_callback_leave(ares_channel_t *channel);

/*! Invoke a query's user callback, tracking callback recursion depth on the
 *  channel so reentrant API calls (notably ares_destroy()) can be detected.
 *  Must be called while holding the channel lock. */
void ares_invoke_query_callback(ares_query_t *query, ares_status_t status,
                                size_t                   timeouts,
                                const ares_dns_record_t *dnsrec);
```

src/lib/ares_destroy.c
```diff
@@ -30,12 +30,45 @@
...
+  ares_channel_lock(channel);
+  if (channel->callback_depth > 0) {
+    channel->destroy_pending = ARES_TRUE;
+    deferred                 = ARES_TRUE;
+  }
+  ares_channel_unlock(channel);
+
+  if (deferred) {
+    /* Nudge the event thread so it observes the pending destroy promptly even
+     * if the deferral originated on another thread. */
```
bradh352, line 55: **Nit** -- "even if the deferral originated on another thread" cannot happen. Query callbacks always run with the channel lock held (`ares_invoke_query_callback()` says so, and the seven entry points bracket their `enter`/`leave` inside the lock), so `callback_depth > 0` implies the deferring thread holds the lock. Another thread calling `ares_destroy()` blocks in `ares_channel_lock()` a few lines above until the depth is back to zero, and so never observes the deferral. I checked: a second thread calling `ares_destroy()` while an event-thread callback sleeps for two seconds blocks until that callback returns, then destroys synchronously.

The wake is still needed -- for a deferral raised on a caller thread inside one of the synchronous dispatch entry points on an event-thread channel -- so only the stated reason is wrong.

```suggestion
    /* The deferral may have been raised on a caller thread inside a
     * synchronous dispatch entry point (e.g. ares_getaddrinfo() answering a
     * literal address), so nudge the event thread to observe it promptly. */
```

test/ares-test-mock.cc
```diff
@@ -1946,6 +1946,115 @@ TEST_P(MockUDPChannelTest, CancelInCallbackNoDoubleFree) {
...
+  DestroyInCbData data;
+  data.channel = channel_;
+  data.done    = false;
+  ares_query_dnsrec(channel_, "www.google.com", ARES_CLASS_IN, ARES_REC_TYPE_A,
+                    DestroyChannelCallback, &data, NULL);
+
+  // Drive processing manually.  We can't use Process() here because the channel
+  // is destroyed from within the callback and Process() would touch the freed
+  // channel on its next loop iteration.  Stop the moment the callback fires.
+  while (!data.done) {
```
bradh352, line 2024: **Nit** -- unlike the event-thread version, this loop has no elapsed-time guard. If the deferral ever stops working, the test hangs until the CI job times out rather than failing with a message. Worth mirroring the `ASSERT_LT(elapsed, 10)` from `test/ares-test-mock-et.cc:1824`.

PR #1260, opened 2026-07-19
detach query before synchronous callback in end_query

jmestwa-coder:
end_query()'s synchronous branch runs the query callback before detaching the query, so a reentrant ares_cancel() from that callback double-frees it:
- reached when ares_close_connection() requeues a query with a NULL requeue array: a malformed response makes process_answer() return EBADRESP, handle_conn_error() tears the connection down, and a retry-exhausted query on it is ended right there
- the callback's ares_cancel() finds the still-linked query (still in all_queries/queries_by_qid) and frees it, then the ares_free_query() below frees it again
- the CVE-2026-33630 fix added detach-before-callback only to the deferred ares_flush_requeue() path, not this synchronous one

Detach before the callback here too, mirroring ares_flush_requeue(). The added test CancelInCallbackConnTeardownNoDoubleFree double-frees on the current code under ASAN and passes with the fix.

### DIFF
--- a/src/lib/ares_process.c
+++ b/src/lib/ares_process.c
@@ -1621,6 +1621,13 @@ static void end_query(ares_channel_t *channel, ares_server_t *server,
     return;
   }
 
+  /* Detach from all lookup lists before the callback, otherwise a reentrant
+   * ares_cancel() from within it would find this query still linked, free it,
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
any update?

PR #1261, opened 2026-07-21
ares_cookie: keep dns cookies usable when source ip is unknown

jmestwa-coder:
ares_addr_equal() treats two AF_UNSPEC addresses as not equal, which quietly disables DNS cookies when the connection source IP can't be determined:
- the classic ares_set_socket_functions() API installs socket callbacks with no getsockname, so ares_conn_set_self_ip() leaves conn->self_ip as AF_UNSPEC (the comment there says cookies should still work with an empty self_ip)
- ares_cookie_apply()'s source-ip-change check then fires on every query, wiping the learned server cookie and rotating the client cookie
- the server cookie is never echoed back, so the RFC 7873/9018 anti-spoofing round-trip is off and the client cookie changes every request

Compare two AF_UNSPEC addresses as equal. DNSCookieUnknownSelfIP added, which drives the unknown source-ip path and checks the server cookie is echoed and the client cookie stays stable across two queries.

### DIFF
--- a/src/lib/ares_cookie.c
+++ b/src/lib/ares_cookie.c
@@ -284,6 +284,12 @@ static ares_bool_t ares_addr_equal(const struct ares_addr *addr1,
         return ARES_TRUE;
       }
       break;
+    case AF_UNSPEC:
+      /* Both source IPs are unknown, e.g. when there's no getsockname socket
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
gentle ping

PR #1263, opened 2026-07-28
Harden DNS response acceptance, caching, and RDATA TLV bounds

M0nd0R:
## Summary
- Enable DNS 0x20 by default alongside EDNS for extra off-path spoofing entropy (explicit opt-out still available via `ares_init_options`).
- Require `QR=1` and a matching opcode before accepting an answer (drops query echoes that only match QID+question).
- Do not cache empty NOERROR responses that lack a usable SOA TTL (avoids clamping forged empty answers to `qcache_max_ttl`).
- Reject OPT/SVCB/HTTPS TLV lengths that exceed remaining `rdlength` before allocating/copying.

## Test plan
- [ ] Existing c-ares unit / fuzz tests
- [ ] Confirm default channel has `ARES_FLAG_DNS0x20`
- [ ] QR=0 / opcode-mismatched packets are ignored
- [ ] Empty NOERROR without SOA is not inserted into qcache
- [ ] Oversized OPT/SVCB/HTTPS TLV returns `ARES_EBADRESP`


M0nd0R:
Closing.

PR #1269, opened 2026-08-06
Do not assume gethostname() NUL terminates its buffer

evilgensec:
POSIX leaves it unspecified whether gethostname() NUL terminates the name when it does not fit the supplied buffer, and whether it reports an error in that case. Both call sites then passed the buffer to strchr(), so on a platform that truncates without terminating the scan can run past the end of the buffer.

ares_init.c passed an uninitialised 256 byte buffer to gethostname(). Zero it with ares_malloc_zero(), pass the full length, and force the last byte to a terminator after the call, so a terminator is present whatever the platform does. On such a platform a truncated name could in principle carry a stray '.' into channel->domains[0], though I have not reproduced that on a mainstream libc.

ares_getnameinfo.c is the weaker of the two. buf[255] was uninitialised and the gethostname() return value was discarded, so an ordinary failure such as ENAMETOOLONG or a sandbox denial left the array uninitialised before strchr() walked it. Zero initialise it, reset it on a failed call to match ares_init.c, and force the last byte to a terminator as well.

No behaviour change for host names that fit the buffer.


evilgensec:
Thanks, both points were fair. Addressed in the amended commit:

**Portability cast.** `ares_getnameinfo.c` now passes `(GETHOSTNAME_TYPE_ARG2)(sizeof(buf) - 1)`, matching the existing call in `ares_init.c`. `ares_setup.h` defines that type as `int` on some targets, so the bare `size_t` was wrong there.

**clang-format.** The CI failure was a comment I added in `ares_init.c` running past the column limit. Both changed files are now clean under `clang-format --style=file`.

**On a regression test.** I could not write one that actually reaches the unsafe condition, and I would rather say so than add a test that passes for the wrong reason.

Triggering it needs either a host name at least as long as the buffer, 256 bytes in `ares_init.c` and 255 in `ares_getnameinfo.c`, or a `gethostname()` that fails so the discarded return value leaves the array uninitialised. Linux caps the host name at `HOST_NAME_MAX`, which is 64 with glibc, so a contained test cannot set a name long enough to truncate, and the harness has no way to force the call to fail. My own reproduction interposed `gethostname()` at link time, which is not a pattern used anywhere in `test/`.

That constraint is also the honest scope of the bug. It is latent on Linux and matters on platforms whose host name limit meets or exceeds the buffer, plus the failure path in `ares_getnameinfo.c`, which is platform independent because the return value was ignored outright.

What I can add if you would find it useful is a `CONTAINED_TEST_F` that pins domain derivation at the longest host name the kernel accepts. That does not exercise truncation, but it does guard the `len - 1` change against an off-by-one. Happy to push it if you want it, and equally happy to leave it out.

Verified locally before and after: heap out-of-bounds read at `ares_init.c:209` and stack out-of-bounds read at `ares_getnameinfo.c:229` under ASAN, both gone with the patch, and the search-domain disclosure demo goes from a leaked value to `ndomains=0`. Normal host names behave the same as before.

Unrelated to this PR, and I am not proposing anything for it here, but I hit two misaligned-load UBSan reports in `ares_sysconfig_mac.c` at lines 276 and 316 while testing on macOS. Line 275 already carries a `/* UBSAN alignment workaround */` comment, so it looks like the same class was handled at one site and not the other. Let me know if you want that as a separate issue.

evilgensec:
@bradh352 the red `Ubuntu (latest)` / `All Checks` jobs are not this patch, and re-running will not clear them.

The test suite builds and runs to completion first — 6,679 passing assertions in run 31886341109. The job then dies in the post-test Valgrind step:

```
valgrind:  Fatal error at startup: a function redirection
valgrind:  which is mandatory for this platform-tool combination
valgrind:  cannot be set up.  Details of the redirection are:
valgrind:  A must-be-redirected function
valgrind:  whose name matches the pattern:      strlen
valgrind:  in an object with soname matching:   ld-linux-x86-64.so.2
```

That is Valgrind refusing to start because it cannot place a mandatory `strlen` redirection in the dynamic loader, which happens when the `libc6-dbg` symbols do not match the runtime glibc. The workflow installs packages via `cache-apt-pkgs-action`, so a cached package set that predates a runner glibc bump would produce exactly this — I can see the action in the log but not the restored version, so treat the cause as a hypothesis and the symptom as verified.

Two things worth noting: it is deterministic rather than flaky, so a re-trigger reproduces it identically (I closed and reopened earlier today and it did, though that same re-trigger did clear genuinely hung NetBSD and Solaris VM jobs, which now pass in 3m36s and 5m23s). And nothing in this diff touches Valgrind, the build, or anything the loader would care about — it is two `gethostname()` call sites.

Happy to rebase if that helps, but I do not think it will change this particular job. Is there anything outstanding on the patch itself?

src/lib/ares_getnameinfo.c
```diff
@@ -223,9 +223,9 @@ static void nameinfo_callback(void *arg, int status, int timeouts,
      */
 #ifdef HAVE_GETHOSTNAME
     if (niquery->flags & ARES_NI_NOFQDN) {
-      char        buf[255];
+      char        buf[255] = "";
       const char *domain;
-      gethostname(buf, 255);
+      gethostname(buf, (GETHOSTNAME_TYPE_ARG2)(sizeof(buf) - 1));
```
bradh352: The pull request description names the discarded return value as the actual defect at this site, but the patch only zero-initialises the buffer. Zeroing does stop the out-of-bounds read, and that part is right. It does not stop the code from *using* whatever a failed call left behind.

POSIX leaves the buffer contents unspecified when `gethostname()` fails, and glibc copies as many bytes as fit before returning -1 with `ENAMETOOLONG`. In that case `strchr()` finds a `.` inside a truncated host name, and `ares_striendstr()` strips a suffix that is not this host's domain from the name handed back to the application, so `ARES_NI_NOFQDN` returns a wrongly shortened name rather than an unmodified one.

`ares_init.c:206` already handles exactly this by resetting the buffer when the call fails. Do the same here, so both sites treat a failed call the same way.

```suggestion
      char        buf[255] = "";
      const char *domain;
      /* POSIX leaves it unspecified whether gethostname() NUL terminates the
       * name when it does not fit, and on failure the buffer contents are
       * unspecified, so keep a terminator and discard anything a failed call
       * left behind */
      if (gethostname(buf, (GETHOSTNAME_TYPE_ARG2)(sizeof(buf) - 1)) != 0) {
        buf[0] = '\0';
      }
```

src/lib/ares_init.c
```diff
@@ -194,13 +194,16 @@ static ares_status_t init_by_defaults(ares_channel_t *channel)
...
     if (!hostname) {
       rc = ARES_ENOMEM; /* LCOV_EXCL_LINE: OutOfMemory */
       goto error;       /* LCOV_EXCL_LINE: OutOfMemory */
     }
 
-    if (gethostname(hostname, (GETHOSTNAME_TYPE_ARG2)len) != 0) {
+    /* POSIX leaves it unspecified whether gethostname() NUL terminates when the
+     * host name doesn't fit, so keep the last byte as a guaranteed terminator
+     */
+    if (gethostname(hostname, (GETHOSTNAME_TYPE_ARG2)(len - 1)) != 0) {
```
bradh352: Reserving a byte by passing `len - 1` costs one character of host name capacity, and there is a form that guarantees the terminator without that cost.

On macOS `sysconf(_SC_HOST_NAME_MAX)` is 255, not counting the terminator, and Darwin's `gethostname()` truncates silently and returns success rather than an error (verified locally). So a 255-character host name used to fit this 256-byte buffer exactly and produced the right search domain; with `len - 1` it is now truncated by one character, and `channel->domains[0]` becomes a corrupted domain that gets appended to every unqualified query. Narrow, but it is a case that worked before and does not now.

Passing the full `len` and forcing the last byte afterwards is safe on every platform. It is a no-op when the name fit, and it supplies the terminator when a platform truncates without one:

```c
    if (gethostname(hostname, (GETHOSTNAME_TYPE_ARG2)len) != 0) {
      /* Lets not treat a gethostname failure as critical, since we
       * are ok if gethostname doesn't even exist */
      *hostname = '\0';
    }

    /* POSIX leaves it unspecified whether gethostname() NUL terminates the name
     * when it does not fit, so force a terminator */
    hostname[len - 1] = '\0';
```

Not written as an applyable suggestion because the added line falls just outside the diff hunk. The same applies to `ares_getnameinfo.c`, where `sizeof(buf)` plus `buf[sizeof(buf) - 1] = '\0'` keeps the full 254-character capacity.

bradh352 (commented):
The change is correct, minimal, and worth taking. Clean build under `-DCARES_WERROR=ON`; `arestest -4 --gtest_filter='*Init*:*NameInfo*:*Search*:-*Live*'` passes under AddressSanitizer and UndefinedBehaviorSanitizer with no new reports; `git clang-format` is clean on the changed lines. `gethostname()` really does have only these two call sites in the tree, so nothing is missed.

Two points inline, plus a few things for the record.

**On how the first site is described.** The pull request body says bytes from outside the buffer end up in DNS query names on the wire. Reaching that in `ares_init.c` needs a host name of at least 256 characters, and no platform in CI permits one: Linux caps it at 64 (`HOST_NAME_MAX`), and macOS at 255 (`sysconf(_SC_HOST_NAME_MAX)`), where `gethostname()` also truncates and NUL-terminates inside the buffer. The trigger is the local system host name, not anything an attacker sends. So this site is defensive hardening rather than a live bug, and since pull requests here are squash-merged, that framing is what ends up in the permanent commit message. Worth toning down there. The second site, `ares_getnameinfo.c`, is the genuine one: the return value was ignored outright, so any failure left an uninitialised stack array for `strchr()` to walk, which is undefined behaviour regardless of host name length.

**On the missing regression test.** I agree with the decision not to add one, with a correction to the record. The truncation path cannot be reached from the test suite: the only in-tree way to control the host name is `CONTAINED_TEST_F`, which calls `sethostname()` inside a Linux user namespace (`test/ares-test-ns.cc:106`), and Linux will not accept a name longer than 64 characters. But the normal derivation path is already covered. `ContainerEmptyInit` in `test/ares-test-init.cc` sets the container host name to `host.domain.org` with no config files and asserts `channel->domains[0] == "domain.org"`, so the `ares_init.c` change is already guarded against breaking the ordinary case. The offered fallback test pinned at the longest name Linux accepts would not add much on top of that, since 64 characters is nowhere near the buffer boundary and so would not catch an off-by-one either.

**On other open work touching these files.** Stated as fact, not as a request to wait for anything. #1263, "Harden DNS response acceptance, caching, and RDATA TLV bounds", edits `init_by_defaults()` in `src/lib/ares_init.c`, a different part of the same function, changing the default channel flags. #1259, "ares_destroy: support being called from within a query callback", edits `src/lib/ares_getnameinfo.c`, specifically the `ares_getnameinfo()` entry point above the `nameinfo_callback()` function changed here. Neither touches the same lines.

**On the red CI job.** For what it is worth, #1278 and #1279 both ran the `Ubuntu (latest)` job more recently and both pass, which suggests whatever broke the runner environment has since cleared. A rebase or fresh run may well go green after all.

**Commit message.** `CONTRIBUTING.md` asks for a `subsystem: summary` subject and a trailing `Signed-off-by:` line; this commit has neither, and every one of the last 25 commits on `main` carries a sign-off. Since pull requests here are squash-merged, the title and body are what lands. Something like `ares_init/ares_getnameinfo: do not assume gethostname() NUL terminates its buffer` would fit.

evilgensec:
Thanks for the review, and for the clear guidance. Updated accordingly:

- Both sites now pass the full length to `gethostname()` and force the last byte to a terminator after the call, so a maximum-length host name is no longer truncated by one.
- `ares_getnameinfo.c` now resets the buffer on a failed call, matching the existing treatment in `ares_init.c`.
- The commit subject uses the `ares_init/ares_getnameinfo:` prefix, a `Signed-off-by` line is added, and the on-the-wire wording in the body is toned down to reflect that I have not reproduced it on a mainstream libc.
- Rebased on `main` so the checks re-run.

Happy to fold in anything further you would like. Thanks again.

PR #1289, opened 2026-09-03
ares_sysconfig: reject overflowing timeout values

XananasX7:
## Summary

Prevent resolver timeout values from wrapping during seconds-to-milliseconds conversion.

The `timeout:N` and `retrans:N` resolver options are parsed as unsigned seconds and converted to milliseconds. Values above `UINT_MAX / 1000` previously wrapped to a much smaller timeout. This change rejects those overflowing values while preserving the existing tolerant handling of malformed resolver options. It also applies the same guard to the libc resolver configuration path.

## Testing

- `git diff --check`
- Strict standalone C89 boundary test for the conversion limits: passed
- Full c-ares build/test: unavailable in the audit container because generated build headers and CMake/Autotools are not installed

## Scope

This is a local resolver-configuration correctness/hardening fix, not a claim of a remotely exploitable vulnerability.

### DIFF
--- a/src/lib/ares_sysconfig.c
+++ b/src/lib/ares_sysconfig.c
@@ -27,6 +27,8 @@
 
 #include "ares_private.h"
 
+#include <limits.h>
+
 #ifdef HAVE_SYS_PARAM_H
 #  include <sys/param.h>
 #endif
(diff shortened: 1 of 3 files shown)


dimbleby:
#1278

PR #1296, opened 2026-09-18
ci: publish release SBOMs

mcc0nnell:
## Summary

- generate a single SPDX JSON SBOM directly from the exact source archive produced by `make dist`
- publish that SPDX document alongside the source tarball and existing SLSA provenance
- verify before upload that the SPDX contains the same SHA-256 used as the provenance subject
- rely on the repository's existing GitHub Actions Dependabot configuration to keep the pinned SBOM action current
- fix `upload-provenance` so its use of `needs.build.outputs.version` has `build` in its declared dependency context

No library, API, ABI, or build-system behavior changes.

## Why

The release workflow already creates the source archive, computes its digest, and generates provenance for those exact bytes.

This change adds one machine-readable inventory at that same boundary. `anchore/sbom-action` scans the release tarball itself using Syft's `file:` source, so the generated SPDX naturally records the archive's SHA-256. There is no format conversion and no SBOM JSON rewriting.

The claim is intentionally narrow: this describes the released source archive as seen by Syft; it is not a complete transitive native dependency inventory.

## Dependency maintenance

The SBOM generator is pinned as a GitHub Action:

```
anchore/sbom-action@3ad7283483fc7af8ff2b4ea19663c2d5ca935e26 # v0.24.2
```

That action pins its default Syft version internally (v1.51.1 at this commit), and this repository's existing `github-actions` Dependabot configuration covers the action pin.

## Validation

Replayed against the published `v1.34.8` source archive:

- archive: `c-ares-1.34.8.tar.gz`
- SHA-256: `c222b6d681096f9444d2c4863d2c1174019e27cacca0a4a5c114d36dd7d7bf78`
- Syft v1.51.1 `file:` scan records that exact SHA-256 in the generated SPDX
- `actionlint 1.7.12`: PASS
- `git diff --check`: PASS


dfandrich:
Rather than supplying two files with (essentially) identical information that have to be tweaked separately, I'd rather just supply one and let users who need the other convert it using one of the many conversion utilities out there, especially since the JSON manipulation seems like it could be brittle. SPDX would be my choice since it is more open.

Also, the syft and sbom-utility dependencies should be set up to be updated and pinned by dependabot (or renovate if dependabot can't do it) so security updates happen automatically.

mcc0nnell:
Agreed. I simplified this along the lines you suggested.

It now publishes only SPDX. I removed the CycloneDX output, sbom-utility, the separate Syft download/version input, and all of the SBOM JSON rewriting.

The pinned anchore/sbom-action scans the actual release tarball via Syft file: input, and the generated SPDX already carries that tarball SHA-256. The only jq left is a validation check that the SBOM digest matches the same archive digest used for provenance.

The action pins its default Syft version internally, and the repo already has Dependabot covering pinned GitHub Actions, so that dependency now follows the existing update path too.

Pushed as 0d4f5bf9. Thanks — this is a lot smaller and easier to reason about now.

PR #1298, opened 2026-09-18
ares_mem: overflow-check the remaining multiplied allocation sizes

Stephen-Psaradellis:
## Summary

`AGENTS.md` says multiplied allocation sizes must be overflow-checked. #1147 added `ares_malloc_zero_array()` and `ares_realloc_zero_array()`, and #1264 converts 21 sites that zero their memory. **This PR converts the other 14**, so that once both are in, no allocation in `src/` multiplies its size unchecked. The two PRs touch different lines. This one is based on `main` and does not depend on #1264.

What this PR does:

- **Adds `ares_malloc_array()` and `ares_realloc_array()`**, the non-zeroing versions of the existing `*_zero_array` pair. #1264 left its 10 `ares_malloc()`/`ares_realloc()` sites alone because no checked equivalent existed, and mentioned these helpers as a possible follow-up. This PR adds them. If the #1264 author would rather carry this part, I'm happy to close this in favour of theirs. Doxygen comments now cover all four array helpers in `ares_mem.h`. That header is internal and not installed, so there is no public API or ABI change.
- **Converts 14 sites.** 10 are the ones #1264 listed as left alone. 3 are zeroing sites #1264 does not include. `ares_sortaddrinfo.c:411` is in neither of #1264's lists.
- **Fixes two unchecked allocation results.** In both, an allocation failure wrote through NULL instead of returning an error (see "Two real defects" below).
- **Adds tests**, including one that fails without this change. They use a small new test-fixture helper that #1264 could also use; its description asks what approach to take for testing these sites.

## Sites

Line numbers are on `main` at 5b18148. "Count source" says where the element count comes from. "Can it overflow?" is my assessment. Most of these are defence-in-depth: the count describes memory that already exists, so a real program cannot make the product wrap.

| # | Site | Before | After | Count source | Can it overflow? |
|---|---|---|---|---|---|
| 1 | `ares_options.c:153` (`ares_save_options`, domains) | `ares_malloc(n * sz)` | `ares_malloc_array` | `channel->ndomains`, the length of an array the channel already holds | No. Defence-in-depth. |
| 2 | `ares_options.c:179` (`ares_save_options`, sortlist) | `ares_malloc(n * sz)` | `ares_malloc_array` | `channel->nsort`, same as above | No. Defence-in-depth. |
| 3 | `ares_options.c:388` (`ares_init_by_options`, domains) | `ares_malloc_zero(n * sz)` | `ares_malloc_zero_array` | `options->ndomains`, an `int` from the application | Only on a 32-bit `size_t`, and only if the caller passes a count larger than its own array. See the note below. |
| 4 | `ares_options.c:417` (`ares_init_by_options`, sortlist) | `ares_malloc(n * sz)` | `ares_malloc_array` | `options->nsort`, an `int` from the application | Same as #3. Covered by `InitSortlistCountOverflow`. |
| 5 | `ares_sortaddrinfo.c:411` | `ares_malloc(n * sz)` | `ares_malloc_array` | number of `ares_addrinfo_node`s in the result list. Influenced by DNS responses and the hosts file. | No. Every counted node is already a larger heap object than its array slot. Defence-in-depth. |
| 6 | `ares_sysconfig.c:605` | `ares_malloc(sz * n)` | `ares_malloc_array` | `sysconfig->nsortlist` (resolv.conf / env / `ares_set_sortlist`), length of an existing array | No. Defence-in-depth. The `memcpy()` below it uses the same product, which cannot wrap once the allocation succeeded. |
| 7 | `ares_sysconfig_files.c:101` (`sortlist_append`) | `ares_realloc(p, (n + 1) * sz)` | `ares_realloc_array` | grows by one per parsed sortlist entry (local config) | No. Defence-in-depth. |
| 8 | `ares_sysconfig_win.c:347` | `ares_malloc(sz * n)` | `ares_malloc_array` | constant 4 | No. |
| 9 | `ares_sysconfig_win.c:405` | `ares_realloc(p, sz * n)` | `ares_realloc_array` | +4 per DNS server found while enumerating Windows adapters | No. Defence-in-depth. |
| 10 | `ares_sysconfig_win.c:453` | `ares_realloc(p, sz * n)` | `ares_realloc_array` | as #9 | No. Defence-in-depth. |
| 11 | `ares_sysconfig_mac.c:207` | `ares_realloc_zero(p, sz * n, sz * m)` | `ares_realloc_zero_array` | existing domains plus `resolver->n_search` (macOS system config) | No. Defence-in-depth. |
| 12 | `event/ares_event_kqueue.c:126` (`ares_evsys_kqueue_enqueue`) | `ares_realloc_zero(p, sz * n, sz * m)` | `ares_realloc_zero_array` | change-list capacity, which doubles on demand | No overflow. **The unchecked result was a real bug**, see below. |
| 13 | `ares_android.c:363` | `ares_malloc(sz * n)` | `ares_malloc_zero_array` | Android DNS server count, capped at `MAX_DNS_PROPERTIES` (8) | No overflow. **The unchecked result was a real bug**, see below. |
| 14 | `legacy/ares_parse_ptr_reply.c:73` | `ares_malloc(2 * sz)` + `memset` | `ares_malloc_zero_array(2, sz)` | constant 2 | No. Converted so the rule has no exceptions. It also drops a redundant `memset()`. |

The 21 sites in #1264 are unchanged here: `ares_addrinfo2hostent.c` x2, `ares_options.c:59`, `ares_process.c`, `ares_search.c` x3, `ares_sysconfig.c:458`, `ares_htable.c` x4, `ares_htable_asvp.c`, `ares_htable_dict.c`, `ares_slist.c` x4, `ares_event_kqueue.c:96`, `ares_event_poll.c`, `ares_strsplit.c`.

**On #3 and #4.** These are the only sites where the count comes straight from the caller. #1264 already flagged #4. `ares_options.ndomains` and `.nsort` are `int`s. On a 32-bit `size_t`, `nsort * sizeof(struct apattern)` wraps at about 179 million entries, which still fits in an `int`. Without the check, `ares_init_by_options()` gets a short buffer back and then copies `nsort` entries into it. But a correct caller cannot get there: its own array of that many entries would not fit in a 32-bit address space. So this is **local API-misuse hardening**, not a remotely reachable fix. None of the 14 counts can overflow on a 64-bit `size_t`.

### Two real defects (allocation failure, not overflow)

These two sites were broken whenever the allocation failed. That includes out-of-memory and an allocator installed with `ares_library_init_mem()` that returns NULL. c-ares guarantees `ARES_ENOMEM` rather than a crash in those cases, and these two sites broke that guarantee.

- **`ares_evsys_kqueue_enqueue()`** stored the result of growing the change list straight into `kq->changelist`, then ran `EV_SET(&kq->changelist[idx], ...)`. On failure that writes through NULL plus an offset. It also doubled `nchanges_alloc` before the allocation succeeded. Now the new list goes into a temporary, and on failure the old list and capacity are kept and this one change is dropped. The function returns `void`, so it cannot report the error; making it do so would change the event-system interface, which I did not want to do here. Affects the BSD/macOS event thread only.
- **`ares_get_android_server_list()`** used the server-list array and each 64-byte entry without a NULL check: `dns_list[i] = ares_malloc(len); dns_list[i][0] = 0;`. Now both are checked. On failure the partial list is freed and no servers are returned; the caller already treats that as "no servers". The list is zero-allocated so that freeing a partial list is safe, and each entry is allocated before the JNI lookup so the failure path does not leak a JNI local reference. Affects Android only.

**Heads-up on #1291.** It adds a parallel `eventlist` realloc inside the same `if` block in `ares_evsys_kqueue_enqueue()`, and that realloc has the same unchecked result. Whichever of the two lands second needs a small rebase. If #1291 lands first, I'll apply the same treatment to its `eventlist`.

## Tests

In `test/ares-test-internal.cc`:

- `LibraryTest.ArrayMisuse` (extended): `ares_malloc_array()` and `ares_realloc_array()` return NULL on overflow in either argument, matching the existing `*_zero_array` checks.
- `LibraryTest.ArrayAlloc` (new): normal use. Also checks that a rejected `ares_realloc_array()` leaves the original array intact and still owned by the caller (the contract documented in `ares_mem.h`), and that a NULL pointer allocates fresh.
- `LibraryTest.ArrayAllocOverflowNeverReachesAllocator` (new): picks a count whose product with `sizeof(char *)` wraps to exactly `2 * sizeof(char *)` on any `size_t` width, arms `SetAllocSizeFail()` for that wrapped size, then calls all four helpers. The armed failure must still be pending afterwards, which proves the allocator was never asked for the wrapped size.
- `LibraryTest.InitSortlistCountOverflow` (new, **compiled only when `SIZE_MAX <= 0xFFFFFFFF`**): the same technique through the public API. It calls `ares_init_options()` with `ARES_OPT_SORTLIST` and an `nsort` that wraps to a distinctive size, and expects `ARES_ENOMEM` with the armed failure still pending. **It fails without this change.** Because the armed failure makes the wrapped allocation return NULL, the unpatched code fails the test cleanly instead of overrunning the buffer.

In `test/ares-test.{h,cc}`: adds `LibraryTest::AllocSizeFailPending(size_t)`, which reports whether a failure armed by `SetAllocSizeFail()` is still pending. That lets a test check "the allocator was never asked for size N" without the unpatched code overrunning anything.

### How it was tested

Windows 11, MSVC 19.44 (VS 2022 Build Tools), CMake 3.31 and Ninja from the Build Tools, GoogleTest 1.15.2 built from source. Base is `main` at 5b181482e5b6d1dc04d1e899dd2c3a9f5695ce62.

```sh
cmake -DCMAKE_BUILD_TYPE=DEBUG -DCARES_BUILD_TESTS=ON -G Ninja -B build -DCMAKE_PREFIX_PATH=<gtest>
ninja -C build
./build/bin/arestest -4 --gtest_filter='-*Live*'
```

I built x64 and 32-bit x86 (`vcvarsamd64_x86`, and `arestest.exe` is a PE32 i386 image). "Before" means unpatched `main` plus only the new test and the fixture accessor, so it compiles unchanged.

| Build | Tree | Result (`-4 --gtest_filter='-*Live*'`) |
|---|---|---|
| x86 | before | 464 run, 463 pass, **1 fail: `LibraryTest.InitSortlistCountOverflow`** |
| x86 | after | **466 / 466 pass** |
| x64 | after | **465 / 465 pass** (`InitSortlistCountOverflow` is compiled out on a 64-bit `size_t`) |

Direct evidence on x86, from the test fixture's verbose allocator log (`arestest -v`):

```
before:  Failing malloc(24008) request
         ...ares-test-internal.cc: error: Value of: AllocSizeFailPending(wrapped)  Actual: false  Expected: true
         [  FAILED  ] LibraryTest.InitSortlistCountOverflow
after:   [       OK ] LibraryTest.InitSortlistCountOverflow
```

In other words, unpatched `ares_init_options()` asked the allocator for 24,008 bytes to hold 178,957,971 24-byte `struct apattern`s. Without the armed failure, the copy loop that follows would have written far past that buffer. With this change it returns `ARES_ENOMEM` without allocating.

**Interaction with #1264.** I applied this patch and #1264's diff to clean exports of `main` in both orders with `patch -p1`. Both orders apply with no rejects and produce identical trees, and in the combined tree no allocation in `src/` multiplies its size unchecked. The combined tree also builds on x64 MSVC and passes the same suite, 465 / 465.

**Build warnings.** MSVC x64 and x86: no new warnings. The only `src/lib` warning on x86 is C4826 in `ares_rand.c`, which appears identically in the unpatched build. MinGW clang 14.0.6 (x86_64-w64-windows-gnu) with `-DCARES_WERROR=ON`: all 100 targets build. This configuration probed `-Wdeclaration-after-statement`, `-Wconversion`, `-Wsign-conversion`, `-Wvla`, `-Wshadow`, `-Wpedantic` and `-Wcast-qual` as supported, and the library is built with `-Werror`.

**Formatting.** `git clang-format` (clang-format 20.1.8) made no changes. Running the CI check locally (`git diff -U0 main -- '*.c' '*.h' ':!src/lib/thirdparty' | clang-format-diff -p1`) produces no output. The `.cc` test code follows the surrounding file's style, as the CI check covers only `.c`/`.h`.

### What I did not run

- **3 of the 14 sites never compiled locally.** Their files are `#if`'d out on Windows: `ares_android.c:363` (Android), `ares_sysconfig_mac.c:207` (Apple) and `ares_event_kqueue.c:126` (BSD/macOS). I relied on careful review and on CI for them. The kqueue and Android changes are the ones that restructure logic, so they deserve the closest look.
- **No autotools build.** No autoconf/automake was available locally. No files were added or removed, so `Makefile.inc` and `Makefile.am` are unchanged.
- **No Linux, macOS or BSD runs, and no ASAN, UBSAN or valgrind.**
- **No coverage report.** The new reachable code, the four helpers, is covered by the tests above. The new failure paths are allocation-failure-only and annotated `LCOV_EXCL_LINE`/`START`/`STOP: OutOfMemory`.
- **No live-network tests** (`-*Live*`).

**Disclosure.** An AI system running an autonomous software project found these sites, wrote the change and the tests, and wrote this description; I'm the account holder submitting it and I have reviewed the patch.


Stephen-Psaradellis:
Closing this as a duplicate: I opened the same change as #1297 an hour earlier and did not spot it before pushing this branch. Same commit, same 14 files. Sorry for the noise — please review #1297 instead.

